PR #2, opened 2019-12-19
Functions for bacterial images saved as .jpk files

gdbenn:
This branch only works for .jpk files, but is easy to change to .spm files etc. 

When imaging bacteria, large scans are performed over whole cells and small scans on their surface. Also, the height and phase data of tapping images show valuable information. All these images must be flattened differently. 

I have also tried to automatically get nice vertical scale values, this is not always successful but is pretty ok (I think).


bobturneruk:
@gdbenn says - this should not be merged as it is doesn't contain changes relevant to topostats.

PR #7, opened 2020-09-23
Add license

bobturneruk:
I'd like some RSE team feedback which I'll solicit via slack before putting this to @alicepyne .


bobturneruk:
Seems like this is OK!

PR #9, opened 2020-10-13
Add contributing instructions

bobturneruk:
I've written this as a template for a few different projects - it can be tailored as needed.


alicepyne (commented):
Looks great to me!

PR #15, opened 2020-11-16
Remove IDE stuff pycharm vscode

bobturneruk:
As far as I know the files in the `.idea/` folder are local IDE settings e.g. pycharm settings. These should not be part of the repository - there's a risk that something you don't want to share could be cached in here. It'd be nice to know if I'm right and if someone could see if checking out this branch and then running some test code still works!


derollins (approved):
It works fine for me in PyCharm and .idea/workspace.xml is generated when I run the code.

PR #40, opened 2021-02-16
Fix errors caused by line breaking

Jean-Du:
During the previous update, some lines were broken into two. I just realised that this would cause errors, so I fixed it by adding brackets.


JoeBeton (approved):
My bad! Changes look good and code runs.

PR #42, opened 2021-03-03
Jean du read the docs

Jean-Du:
An attempt at making a read the docs. At the moment the documentation site doesn't quite work yet.


bobturneruk:
I can build locally on Windows 10 using:

```
cd docs
make html
````

Readthedocs runs on linux, so I'm going to check what happens on linux next.

bobturneruk:
So it doesn't build properly on Linux using the `Makefile`. So there's a problem in the `Makefile`.

Issue #18, opened 2020-11-30
Add a command line interface


bobturneruk:
argparse (https://docs.python.org/3/library/argparse.html) is a od way of turning command line arguments into python variables.

bobturneruk:
https://click.palletsprojects.com/en/8.0.x/ may be better

JoeBeton:
Can this issue be closed since introduction of cfg file?

Jean-Du:
I think it can be closed too

PR #44, opened 2021-03-25
Docker

bobturneruk:
I'd welcome feedback on this PR which adds a Docker container in which TopoStats can be run, and instructions for doing this. I'm interested to know if this works and if it's a massive hassle. I'm keen to get input from people with different levels of familiarity with the software - I think this may be most useful for people not doing development, but just trying to run it.

Please do not merge this pull request - I need to change the tag in the readme to point to a different Docker image version before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,14 +37,9 @@ Firstly, you will need to clone the TopoStats repository if you have not done so
 git clone https://github.com/AFM-SPM/TopoStats.git
 ```
 
+If this does not work, then you may have to install git. Alternatively, you can just download and extract the zip file from GitHub directly.
+
 Docker conatiners provide a way to run software in a controlled environment, without having to install lots of packages on your computer. To experiment with TopoStats using Docker you will first need to install [Docker](https://docs.docker.com/get-docker/). Then, using the command line, "pull" the container image to your computer:
-
-```
-docker pull afmspm/topostats:wip
-```
-
-Docker containers provide a way to run software in a controlled environment, without having to install lots of packages on your computer. To experiment with TopoStats using Docker you will first need to install Docker.
-After installing Docker, using the command line, "pull" the container image to your computer:
 
 ```
 docker pull afmspm/topostats:wip
@@ -52,7 +47,9 @@ docker pull afmspm/topostats:wip
 
 This allows docker to know how and what to run as a virtual machine. Docker then needs to be able to access TopoStats on your operating system. This can be done by mapping the TopoStats folder on your operating system to a folder in the virtual machine. The following command both maps the folders and starts the virtual machine:
 
+```
 docker run -it -v <path/to/your/topostats>:/home/TopoStats afmspm/topostats:wip
+```
 
 This should have changed the location stated in the terminal from 'C:\ 'to something like 'root@', meaning that any commands will be running in the Docker container.
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Docker conatiners provide a way to run software in a controlled environment, wit
 docker pull afmspm/topostats:wip
 ```
 
-This allows docker to know how and what to run as a virtual machine. Docker then needs to be able to access TopoStats on your operating system. This can be done by mapping the TopoStats folder on your operating system to a folder in the virtual machine. The following command both maps the folders and starts the virtual machine:
+This allows docker to know how and what to run as a virtual machine. Docker then needs to be able to access TopoStats on your operating system. This can be done by mapping the TopoStats folder on your operating system to a folder in the virtual machine. **The files that you wish to analyse will need to be in your TopoStats folder.** The following command both maps the folders and starts the virtual machine:
 
 ```
 docker run -it -v <path/to/your/topostats>:/home/TopoStats afmspm/topostats:wip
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,6 +29,29 @@ Statistical analysis and plotting is performed using the 'statsplotting' script.
...
+Docker conatiners provide a way to run software in a controlled environment, without having to install lots of packages on your computer. To experiment with TopoStats using Docker you will first need to install [Docker](https://docs.docker.com/get-docker/). Then, using the command line, "pull" the container image to your computer:
+
+```
+docker pull afmspm/topostats:wip
+```
+
+Then run the container, mapping your current folder (the root of this repository) to a folder in the container (e.g.):
+
+```
+docker run -it -v C:\Users\bobturner\Documents\TopoStats:/home/TopoStats afmspm/topostats:wip
```
willfurnass, line 43: Be warned that this gives the container write access as root to whatever directory/directories you expose into the container.  One could either make the 'bind-mount' read-only or modify the container image so that by default it doesn't run as root.

README.md
```diff
@@ -29,6 +29,29 @@ Statistical analysis and plotting is performed using the 'statsplotting' script.
...
+```
+
+Your next commands will be executed within the container:
+
+```
+cd home/TopoStats
+xvfb-run python pygwytracing.py
```
willfurnass, line 50: https://stackoverflow.com/a/25280523 might be useful if you want to access the GUI from the Docker host.

willfurnass: Alternatively, if you're running a 'X' display system server on the Docker host you could add `-e DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix:rw` to the Docker command-line so programs in the container can display UIs using that X display server.  It's less portable/versatile than the above (VNC-based) approach though.

JoeBeton:
Hi all, sorry for getting to this so late. I think using Docker for package management is a great idea and should simplify gwyddion installation and locating the pygwy files after installing. However, I had some problems using Docker initially which we might want to address in the README file. 

Basically, (from what I can tell) on Linux Docker installs all image files into /var/lib/docker by default. This was an issue for me as my filesystem root directory is pretty tight on space and the TopoStats image is surprisingly large (>1 GB) and so downloading/installing the image file (using `docker pull afmspm/topostats:wip`) completely filled up my filesystem directory. 

I solved this by changing the download location for docker image files (using this method: https://stackoverflow.com/a/50217666/15382229). This might all have happened from something stupid I did during the install but could also be a problem for other Linux users so might be worth addressing in the README?

willfurnass:
@JoeBeton If you're struggling for space with Docker writing parts of images to /var/lib/docker then another approach is to use Podman, which is a drop-in replacement for Docker but can be configured to write parts of images to your home directory instead. https://developers.redhat.com/blog/2020/09/25/rootless-containers-with-podman-the-basics/

SylviaWhittle:
I would like to suggest that the README instructions be added to, to clarify parts of the installation process. Here are a few vague suggestions based on my experience following the guide:

- Make it clear that pulling the docker image is to enable docker to run the linux VM, which is not the same as cloning the repository.
- Instruct the user to clone the repository to their computer as well as pulling the docker container image. 
- Add a little bit about how there will be a virtual environment running which will need access to the cloned repository folder on the user's OS to be able to run TopoStats.
- Make it a little more clear that the mapping of the current folder to the folder in the container is so that the docker container can 'see' the TopoStats repository on the user's OS.

An example:

Firstly, you will need to clone the TopoStats repository if you have not done so. It can be cloned by opening a terminal window and navigating to where you want to save TopoStats, then running the command:

`git clone https://github.com/AFM-SPM/TopoStats.git`

If this does not work, then you may have to install git. Alternatively, you can just download and extract the zip file from GitHub directly.

Docker containers provide a way to run software in a controlled environment, without having to install lots of packages on your computer. To experiment with TopoStats using Docker you will first need to install Docker. 
After installing Docker, using the command line, "pull" the container image to your computer:

`docker pull afmspm/topostats:wip`

This allows docker to know how and what to run as a virtual machine. Docker then needs to be able to access TopoStats on your operating system. This can be done by mapping the TopoStats folder on your operating system to a folder in the virtual machine. The following command both maps the folders and starts the virtual machine:

`docker run -it -v C:\Users\bobturner\Documents\TopoStats:/home/TopoStats afmspm/topostats:wip`

Replacing 'bobturner' with your username. The file structure may be different depending on where you cloned TopoStats to.

This should have changed the location stated in the terminal from 'C:\ 'to something like 'root@', meaning that any commands will be running in the Docker container. 

Navigate to the mapped folder in the virtual machine using:

`cd home/TopoStats`

Then finally run TopoStats, using the command:

`xvfb-run python pygwytracing.py`

This should run the pygwytracing.py script for TopoStats, targetting the repository folder that was mapped on your operating system, looking for spm files.

The terminal will output information about the status of the TopoStats script. Any results will be placed in a folder in the TopoStats folder, called 'processed'.

README.md
bobturneruk: Sure - however, I feel like this could be the subject of a future pull request.

bobturneruk:
@MaxGamill-Sheffield - does this latest change help, for now?

TopoStats needs a command line interface (https://github.com/AFM-SPM/TopoStats/issues/18). It will then be possible to pass command line arguments (e.g. paths) to docker. But this is more work than we can do quickly enough for this PR.

PR #46, opened 2021-08-09
Protein

Jean-Du:
New plotting script Plotting.py that allows for more customizable parameters such as axes limits and groupability; introducing customizable sample type in pwygwytracing.py so the user can choose between DNA, protein and MAC without having to manually adjust hardcoded variables


derollins (approved):
Parameter switching works although I've still has to change parameters manually for protein and DNA on PLL that appear higher. 
New plotting script is really easy to use. 
All seem to work.

PR #49, opened 2021-11-02
Added configuration file support

SylviaWhittle:
I have changed how TopoStats handles input parameters. Previously, one would have to change the code in pygwytracing.py to edit the parameters such as `sample_type`, but with this change, all the parameters are saved in a configuration file, config.ini. This is able to easily be opened by a simple text editor and is able to be understood without seeming intimidating to the user. It also allows the user to have pre-set configurations for the different types of samples, such that when changing from dna to protein, one does not lose their parameters for dna and doesn't have to comment out large blocks of code like before. 

This config file also means that when a user updates TopoStats, their configuration is not overwritten since the repository does not contain the config file, rather, the user has it on their machine. If a config file is not found when TopoStats is ran, a config.ini file is created with some default parameters that are set in the main function of pygwytracing.py.


bobturneruk:
Thanks for doing this. It's on my list to have a look.

bobturneruk:
Thanks again @SylviaWhittle - some initial thoughts on this generally excellent work:

- Consider adding `config.ini` to .gitignore, to avoid people accidentally committing their custom config files.
- Consider using a `default_config.ini` to store the defaults, rather than multiple `add_section`, `set` statements etc. in imperative code. Something like this will be easier to maintain:
```
if(not os.path.isfile("Config.ini")) :
    # Copy default_config.ini to config.ini
```
- Consider add instructions on how to use the new system
- Consider adding a note to this PR on how reviewers should test your code to make sure it's doing what you intended

Happy to chat about any of this, just stick a meeting in my calendar. Once again, thanks - this is great progress.

PS Ultimately, we will need to move away from print statements towards logging / output files, but maybe not now.

SylviaWhittle:
For reviewers: 

The intended functionality is as follows:

- If there is no config file, ```default_config.ini``` should be copied to ```config.ini```
- The code should successfully obtain configuration parameters for the data analysis from ```config.ini```, and run the analysis using them. 
- ```config.ini``` should be able to be edited easily with a basic text editor to change the parameters that will be loaded. 
- ```config.ini``` should be ignored when committing to the repo. 

To test this, please try running ```pygwytracing.py``` and check that ```config.ini``` is successfully copied and loaded, then try editing the file in a text editor and confirm that this works by running the script again.

bobturneruk:
@SylviaWhittle - this looks good. Thanks for addressing my comments. Please mark this as "Ready for review", if it is. I think it needs feedback from some others in the team to check it still works and is usable.

derollins:
Tested code and it worked fine although had to `pip install configparser` this could be added to instructions. 

Other hard coded features that I change regularly and would be helpful to add to the config file, mainly to do with the image outputs:

- Turn `saveTraceFigures` on and off (slows down and sometimes causes memory errors, not often needed)
- Change `savefiles` scale option  `s["/module/pixmap/xytype"] = 1` 1 for ruler and 2 for inset scale bar (0 for non)
- Turn z bar, on change line 55 to `s["/module/pixmap/ztype"] = 1` 
- Change `savefile` output colour scale line 557 `palette = data.set_string_by_name("/%s/base/palette" % k, "Nanoscope.txt")
` Nanoscope.txt to something else e.g. Gwyddion.net

derollins (approved):
Works for me using the standard (non-docker) install on windows although I had to install configparser before it would work, this could be added to installation instrucitons.

pygwytracing.py
```diff
@@ -701,70 +703,62 @@ def searchgrainstats(df, dfargtosearch, searchvalue1, searchvalue2):
...
+    print("Reading config file")
+    config = configparser.ConfigParser()
+    config.read("config.ini")
+    print("Config loaded: ")
+    print(" ")
+
+    # Load and print the current parameters
+
+    # Main section
+    path = config.get("MainSection", "path")
```
MaxGamill-Sheffield, line 725: This relative file path stops the "Tracestats.json" file from being created (on docker at least)
```suggestion
    path = os.path.abspath(config.get("MainSection", "path"))
```
Screenshot 2021-11-22 at 14 30 45

MaxGamill-Sheffield (changes requested):
Might also want to pull the change I made to dnatracing.py or make the change yourself in that file as this contains a bug fix seen when running TopoStats for multiple files.

pygwytracing.py
MaxGamill-Sheffield: Also wanted to say this is super helpful and looks like a great addition to TopoStats, awesome job!

SylviaWhittle: I have added your change to the config file branch and resolved the merge conflict.

SylviaWhittle: Hopefully I have done it correctly! Also thank you for reviewing it!

default_config.ini
```diff
@@ -41,3 +41,13 @@ mindeviation = 0.5
 gaussian = 0.25e-09
 thresholdingcriteria = 2.1
 
+[ImageOutput]
+; Turn saving figures on and off. 1 : on, 0 : off
+saveTraceFigures_option = 1
+; Save file scale option: 0 : none, 1 : ruler, 2 : inset scale bar
+savefilesScale_option = 1
+; Save file name for output colour scale eg: Nanoscope.txt
+savefilename = Nanoscope.txt
```
MaxGamill-Sheffield, line 50: I'm not sure this an output rather than an input that is missing - I think the Nanoscope.txt file is a colour pallet used in Gwyddion for shading the AFM images. @derollins could you confirm and suggest if this needs any clearing up?

MaxGamill-Sheffield (approved):
Just confirmation of what that Nanoscope.txt file does from @derollins for clarity but this should still work.

default_config.ini
derollins: Yes, that is a colour scale option for Gwyddion.  `savefilename` might be a confusing name, maybe `savefile_zscalecolour'

derollins (changes requested):
Name isn't updated in `default_cofig.ini` so won't run

derollins:
Would be helpful to add a note in the config.ini file that the `sample_type` is case sensitive

derollins (approved):
Works great, very useful

bobturneruk:
Hi @SylviaWhittle and @derollins - more great work, I see - I suggest that anything "helpful" (i.e. not an error) is made into a new issue so it can be handled in a later pull request and this one can be merged as soon as possible, giving everyone the benefit of the changes made so far.

PR #52, opened 2021-11-15
Relative File Path & Wrong Filename Bug Fixes

MaxGamill-Sheffield:
The relative file path made it so certain files were not created - resolved by amending to fullpath.
Incorrect reference to self made it so all image names and image paths in the tracestats.json files were the same regardless of different files being used.

### DIFF
--- a/dnatracing.py
+++ b/dnatracing.py
@@ -678,7 +678,7 @@ class traceStats(object):
 
         data_dict = {}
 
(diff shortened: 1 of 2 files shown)


bobturneruk (approved):
This looks good to me. In the absence of automated tests, it should also be practically checked by a topostats **user**.

derollins (approved):
Still works when I run this update

PR #58, opened 2021-12-17
NDP

Jean-Du:
1) The measurement of end to end distance
2) The measurement of curvature (uses splined traces for circular molecules, and fitted traces for linear ones
3) A quick fix for the masking issue: previously grains that are lower than the mean height by the specified threshold were also masked, and this has been fixed by adding a high threshold (5 sigma) for the lower end; there might be a better solution to this
4) Changing output txt files to csv files
5) Adding the parameters for NDP into default_config.ini
6) Adding plotdist2var and computeStats (used for the NDP paper) in Plotting.py,


derollins (changes requested):
Remove NDP variables from default config. Allow the protein variables to be edited by user. Don't want the config file to get too big with every individual sample.

derollins (approved):

PR #59, opened 2021-12-17
Add smoke test

bobturneruk:
i.e. we switch it on and see if starts emitting smoke. By which I mean: install and run the software in a standard basic way and see if there's an error.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -9,13 +9,9 @@ jobs:
   test:
     name: Smoke Test
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bobturneruk:
After merge, a successful run of this will be a requirement for future pull requests. A successful run will mean that the instructions to run topostats using docker have been automatically carried successfully out by GitHub. Expaning `> Run TopoStats` here shows output https://github.com/AFM-SPM/TopoStats/runs/4561862393?check_suite_focus=true

SylviaWhittle (approved):
Looks good to me. I have no hands-on experience with GitHub automated tests, however I have watched a few tutorials on it.

PR #60, opened 2022-01-28
Disable curvature measurements to prevent it from causing errors

Jean-Du:
The curvature measurement could cause errors for linear molecules as it's using the fitted traces. As this is still work in progress, the curvature measurement functions are disabled so that these errors won't interfere with the rest of the program.


bobturneruk:
Hi @Jean-Du - good stuff. Best to use the reviewers tab:

image

to request reviews. Little cog thing.

bobturneruk (approved):
I can't really test this out myself, but seems fine in principle.

Issue #10, opened 2020-11-02
Add installation instructions to readme


bobturneruk:
On Windows, these might form the basis, but need to be checked.

# Install

* Install 32 bit Anaconda (https://www.anaconda.com/distribution/#windows) - if you have 64 bit installed as well, make sure you’re using the 32 bit version
* Install 32 bit Gwyddion for Windows (http://gwyddion.net/download.php#stable-windows)
* Download the additional prerequisites (https://sourceforge.net/projects/gwyddion/files/pygtk-win32/)
* Use Anaconda Navigator to import an environment from gwyconda.yml (in the Google Drive)
Identify the path to the copy of Python in the new conda environment. A way to do this:
    * Open Anaconda Powershell prompt
    * Type `conda activate gwyconda` (this puts you in the correct environment)
    * Type `Get-Command python`
* Install (for all users) from the Google Drive Folder making sure to select the folder containing the version of python identified in the previous step
    * pycairo-1.8.10.win32-py2.7.msi
    * pygobject-2.28.3.win32-py2.7.msi
    * Pygtk-2.24.0.win32-py2.7.msi

# Test

* Open Anaconda Powershell prompt
* Type `conda activate gwyconda` (this puts you in the correct environment)
* Type `cd 'C:\Program Files (x86)\Gwyddion\bin'` or `cd <your_gwyddion_path>\bin`
* Type `python` (you should see something like:
```
Python 2.7.16 |Anaconda, Inc.| (default, Mar 14 2019, 15:49:40) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
```
* Type `import gwy`

If you get no errors, you should be good to go…

bobturneruk:
Also add instructions to clone the repository.

derollins:
> Also add instructions to clone the repository.

Here how to clone the repository in command line, haven't worked out how to do it in PyCharm yet

# Clone the Repository 

- Ensure that you have git installed (download git here: https://git-scm.com/downloads)
- In command line or powershell navigate to the folder where you want to install Topostats,
- Type `git clone https://github.com/afmstats/TopoStats.git`. This will make a copy of the code in a folder called TopoStats.
- Make a branch to work on by typing `git checkout -b <new branch name>`

bobturneruk:
Good work @derollins! In pycharm you need to close your current project then start here:
[...]

PR #62, opened 2022-02-17
Make topostats pip installable

bobturneruk:
This pull requests makes topostats pip installable via `pip install -e .` using the package name `topostats` (all lower case). It does not enable `pip install topostats` (i.e. from PyPi). This will require another step to push our package to the index, and we should probably wait for this until we've sorted some of the dependency issues.

The changes are:
- Put the core topostats code into a folder called `topostats` and the utility functions in `utils`.
- Add an `__init__.py` to the `topostats` folder.
- Add a `setup.py` to the repository root.
- Make the `main` function of `pywgytracing.py` into a function called `trace`.
- Rename `pywgytracing.py` `tracing.py` in the module.
- Remove several global variables from the new `tracing.py`.

This should only minimally affect installation (if you're in the cloned repo. folder you don't actually need to pip install), and should not affect how the user operates the software. **I'm keen to know if this is the case!**

This pull request enables:
- Publishing to PyPi at some future point.
- Setting up tests using `pytest` in a conventional way.


bobturneruk:
The changes to what was `pygwytracing.py` can be more easily seen here:

https://github.com/AFM-SPM/TopoStats/compare/3fe5b00..53e7af1

JoeBeton:
This is nice and would be great to have topostats available on PyPi. 

Something for the future: we can edit the setup.py so that pwgwytracing can be executed from anywhere after installing (I think using entry points keyword in setup.py)

bobturneruk:
Hi @JoeBeton - thanks! The idea was to leave `pygwytracing.py` out of the package, such that it serves as an example of how to call `topostats.tracing.trace()`. I guess if we want people to be able to execute topostats via a single command after pip installing, that's something to explore. There's probably a few ways of doing this. This may be an option https://packaging.python.org/en/latest/guides/installing-stand-alone-command-line-tools/

bobturneruk:
Also (mainly a note to self this) https://docs.python.org/3/library/__main__.html#main-py-in-python-packages

bobturneruk:
Would probably want to take config file as a command line arg.

MaxGamill-Sheffield:
No changes in the produced TopoStats.csv files from before and after this change.

bobturneruk:
That's great. (We will automate that check in a future update.) Did you 😨 try to reinstall?

MaxGamill-Sheffield:
> That's great. (We will automate that check in a future update.) Did you 😨  try to reinstall?

I did reinstall using docker but I did it the normal way but using `git clone ... -b pip`. Is this what you meant or would you like me to try and install using pip? - if so I may need instructions for this

bobturneruk:
This needs to be checked against https://github.com/AFM-SPM/TopoStats/issues/10 before merging.

Jean-Du (approved):
The pip install seems to be working for me and I can run pwygwytracing.py normally. Would it be worth changing this into main.py and renamaing the topostats/tracing.py as topostats/pwygwytracing.py, for consistency reasons?

bobturneruk:
Thanks @Jean-Du - that's a good point. I didn't want to change how users run TopoStats at all, but now we're targeting the pre-release `dev` branch, we could make that change and keep things consistent for the developers.

bobturneruk:
@Jean-Du - how does this look, please? You can click the last commit to see the changes.

Jean-Du (changes requested):
It's giving me the error message 
`IOError: [Errno 2] No such file or directory: 'default_config.ini'
No config file found named 'config.ini'`, presumably because pygwytracing.py and the config file are not under the same directory now?

bobturneruk:
Thanks @Jean-Du - I'll need to fix this.

SylviaWhittle (approved):
Works fine for me. The decision to put the core python files in one folder is good, it makes the software more understandable and sensibly structured. 

(Ran using Docker. Windows 10 up-to-date)

bobturneruk:
Thanks @SylviaWhittle ! Not sure about the config.ini thing, but *am* sure I should've put it in the package as it's necessary to run the software. I'll do some more work on it.

bobturneruk:
Hi @Jean-Du - this should now be working. Also once it's installed it should be possible to run topostats in any folder, not just the repository. Please try again and mark changes addressed if it works!

Jean-Du (changes requested):
I am still having issues with the config file. As it uses default_config.ini, which defaults the path to the current directory, this leads to an error as there are no spm files in the topostats directory.

bobturneruk:
Hi @Jean-Du - should be running in the context of the directory in which you invoke the command. What directory are you in when you type `python -m topostats` please?

Jean-Du:
@bobturneruk Directly running in a directory with spm files seems to work fine. I only had the issue when I was running it in the same directory as the script

Jean-Du (approved):
As long as the user follow the instructions in the readme it should work fine

bobturneruk:
Thanks @Jean-Du !

PR #70, opened 2022-02-23
Make topostats pip installable

bobturneruk:
This pull requests makes topostats pip installable via `pip install -e .` using the package name `topostats` (all lower case). It does not enable `pip install topostats` (i.e. from PyPi). This will require another step to push our package to the index, and we should probably wait for this until we've sorted some of the dependency issues.

The changes are:
- Put the core topostats code into a folder called `topostats` and the utility functions in `utils`.
- Add an `__init__.py` to the `topostats` folder.
- Add a `setup.py` to the repository root.
- Make the `main` function of `pywgytracing.py` into a function called `trace`.
- Rename `pywgytracing.py` `tracing.py` in the module.
- Remove several global variables from the new `tracing.py`.

This should only minimally affect installation (if you're in the cloned repo. folder you don't actually need to pip install), and should not affect how the user operates the software. **I'm keen to know if this is the case!**

This pull request enables:
- Publishing to PyPi at some future point.
- Setting up tests using `pytest` in a conventional way.


bobturneruk:
Supersedes #62

PR #72, opened 2022-03-03
Filter framework

bobturneruk:
Creates a means of adding and testing 2D image filters. May need to be discussed a bit before merging in case what I've proposed doesn't work for other people. Feedback welcome here.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


bobturneruk:
Depends on merge of #74

bobturneruk:
This was discussed at a meeting with @SylviaWhittle , @benthomasuos , @Jean-Du and @MaxGamill-Sheffield so I'm going to use admin powers to merge it.

Issue #61, opened 2022-02-04
Revise contributing guidelines


bobturneruk:
Check existing contributing guidelines and revise to reflect current good practise.

bobturneruk:
Closed by #73

Issue #22, opened 2021-01-25
Reformat readme


bobturneruk:
Add headings and other nice formatting.

ns-rse:
This has been done on the `dev` branch and merged with #73 but I can't currently see a way for me to close issues.

bobturneruk:
Specifically, I think a lot of the explanatory text should be migrated to Sphinx and form part of the readthedocs site, leaving a minimal readme on GitHub.

bobturneruk:
Resolved via multiple PRs.

PR #73, opened 2022-03-04
Contributing

bobturneruk:
Changes to readme and contributing to make them look nicer and be more easy to read and that. Also info on planned software architecture (which can totally change as we go along, just want some kind of plan). More detail on how to use the architecture to follow in #72 .

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -64,20 +64,27 @@ Object oriented approaches will be used where appropriate, but not seen as inher
 ## Using pycharm and GitHub
 
 Check the issues and draft pull requests.
+
 ![](assets/issues-prs.png)
 
 Make a branch for your contribution.
+
 ![](assets/pycharm-branch.png)
 
 Add any new files to `git` (i.e. stage them).
+
 ![](assets/stage.png)
 
 Commit the changes (with a sensible message).
+
 ![](assets/commit.png)
+
 > Add / commit multiple times to break your contribution into easy to follow chunks.
 
 Push changes.
+
 ![](assets/push.png)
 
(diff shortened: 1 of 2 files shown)


bobturneruk:
Resolves #61 insofar as such a thing can ever be fully resolved!

bobturneruk:
Resolves #22 sufficiently to close, but more work on the "How it works" sections would be helpful as part of a separate pull request.

contributing.md
```diff
@@ -5,23 +5,35 @@ This document explains how, technically, to contribute to this project. A code o
...
+## Software Architecture
+
+`topostats` is currently arranged as a Python module in the `topostats/` folder. Other scripts in this repository are not currently considered core elements of `topostats` apart from `Plotting.py`.
+
+Currently the `topostats` module consists of:
+
+* `default_config.ini` The default config file.
+* `dnatracing.py`
+* `pygwytracing.py` The "main" routine.
+* `tracingfuncs.py`
```
MaxGamill-Sheffield, line 24: do dnatracing.py and tracingfuncs.py also need brief descriptors? Something like dnatracing - applies tracing functions to each molecules, and for tracingfuncs - skeletonises and generates backbone traces from masks (I think that's what they do at least)

MaxGamill-Sheffield (approved):
Just one comment about defining the other module constituents to be consistent

SylviaWhittle (approved):

contributing.md
bobturneruk: Good idea. I think these will be subject to further revisions as well.

PR #75, opened 2022-03-09
Docker debug

bobturneruk:
Instructions and "utilities" for debugging topostats in a Python 2 Docker container, if you want to. e.g. if you're running Windows or macos and can't install it locally. I'd be keen for reviewers to feed back on whether they are able to get the debugger running using the new instructions, which are vscode specific.


bobturneruk:
This is now ready for a check, if anyone has time.

SylviaWhittle (approved):
Works well for me! Is the addition of Codecov something I should be looking at too?

bobturneruk:
@SylviaWhittle - codecov is being set up on the repo. It will make it easy to see which bits of code are covered by tests.

bobturneruk:
I think this is helpful enough to merge, so I will. All this is just "scaffolding" while we move from python 2 to 3, and remove the Gwyddion dependency. Once we've done this, topostats should be easy enough to install and debug without docker.

PR #76, opened 2022-03-10
Fix the typo with default_config.ini

Jean-Du:


bobturneruk (approved):
Good spot. Good work. Can you make another PR against `dev` please?

PR #77, opened 2022-03-11
Fix the typo in the default config file

Jean-Du:


bobturneruk (approved):
Great. Sorry we have to go against both `master` and `dev` for now.

Issue #66, opened 2022-02-17
Remove Gwyddion - Filters


bobturneruk:
## Filters

Gwyddion AFM specific `flatten_base`, `polylevel`, `align_rows`, `zero_mean`, `scars_remove`, `level` filters. These would probably have to be rewritten, but are pretty simple. 

Also Gaussian filter (in scikit image).

SylviaWhittle:
All done except `polylevel` and `scars_remove` which have been deemed nice-to-have and non-essential for now.

PR #81, opened 2022-03-31
Add raster filters

bobturneruk:
Aspects of @SylviaWhittle 's recent work on filters cherry picked from https://github.com/AFM-SPM/TopoStats/tree/dev-pySPM-tests


bobturneruk:
I think this needs unit tests and can then be merged.

bobturneruk:
Partially resolves #66, but we still need to remove the Gwyddion code.

ns-rse:
I can start writing some unit tests but was wondering if there is a sample raster image I could use to do so or should I just make a simple sample image?

bobturneruk:
There's a "randomised pixels image" fixture here https://github.com/AFM-SPM/TopoStats/blob/5dcf66d7682c50bbef8a6320a85fb5174f984942/tests/test_filters.py#L8 that could be moved to here https://github.com/AFM-SPM/TopoStats/blob/dev/tests/conftest.py and thus be available to all the tests.

Our stock test image is here:
https://github.com/AFM-SPM/TopoStats/blob/dev/minicircle.spm

May be best moved or copied to a `test_data/` folder, which can be added to, as needed. Need to be careful about how much binary data goes in the repo - could be large test sets have to go elsewhere.

For now, I'd favour trying to generate the type of noise to be removed in fixtures and then checking that the filter removes it. But I could be wrong about this approach.

ns-rse:
Thanks @bobturneruk and apologies, I'd already found the `minicircle.spm` but hadn't actually looked at it.

I'd started using it in a simple test for finding files. The general convention for `pytest` is to place files that are used in testing under `tests/resources/` so if that's ok I'll go with that. I can set up the fixture under `tests/conftest` too.

Cheers :+1:

bobturneruk:
Cool. Definitely go with convention, here!

bobturneruk:
This stuff may be useful, was hitherto only on Slack:

The code is in such a pickle it's better just to explain. Hopefully we can add an implementation to topostats, because while it's fairly trivial, I found it really useful. My model for AFM noise is that there is a mean 0 Gaussian blur from tip convolution, and also mean 0 Gaussian noise on the fast scan axis, and (line by line) on the slow scan axis. The tip convolution blur s.d. is estimated and applied as a Gaussian blur (you're doing this, too, I think). For the other noise I took some data with the AFM set up exactly as for imaging a relevant sample but with zero scan size. The idea was I just got the noise. This is how I got the noise s.d.s from one of those images:
```
slow_noise=std(mean(image,1));
fast_noise=mean(std(image,0,2));
```
This noise could be applied to a simulated image (already with a Gaussian blur tip convolution) like this:
```
image_size=size(image);
fast_noise_image=randn(image_size(1), image_size(2)).*fast_noise;
slow_noise_lines=randn(1,image_size(1));
slow_noise_image=repmat(slow_noise_lines', 1, image_size(2)).*slow_noise;
image_out=image+fast_noise_image+slow_noise_image;
```

bobturneruk:
Bit of a weird one this, in that I raised the PR but it's @SylviaWhittle and @ns-rse 's work. I'm going to take a look at it.

bobturneruk:
I think the `raster-filters` branch history has become too convoluted for it to be merged, so I've made a new one. https://github.com/AFM-SPM/TopoStats/pull/91

PR #85, opened 2022-04-01
Get a proper Codecov report and add badge

bobturneruk:
This should be working https://app.codecov.io/gh/AFM-SPM/TopoStats/


ns-rse (approved):
Not sure at what point this started working or why the addition of a blank line as in this PR might have caused that but it is working now so I'm going to approve this.

PR #86, opened 2022-04-01
Changed proposed file names of new files in contributing.md

benthomasuos:
New filename of io.py clashed with another "io" module import in zipfile. Changed to file_ip.py
Added "plottingfuncs.py" for the matplotlib plotting function that @SylviaWhittle created in dev-PySPM-tests


bobturneruk:
Thanks for this. Please can you:
- Add `build/` to .gitignore so the contents don't become part of the repo (I assume this was not planned)
- The same for `dist/`

Looks like perhaps `topostats/file_io.py` is not ready for merging yet? If this is the case, please remove from this PR.

It's fine to have a PR that only makes minor changes to a single file. Little and often is best.

bobturneruk (approved):
Well spotted and a wise change @benthomasuos. We may need to change again as we go forward depending on how code is implemented, but this is much better for now. Thanks. Please go ahead an merge when ready.

PR #88, opened 2022-04-04
Unit tests

ns-rse:
I've added unit tests for most functions, although whilst working on these the `remove_x_bowing()` function has been added, so I'll address that in subsequent work. This will help with #81 and should allow it be merged to `dev` (if I've understood the branching correctly).

Looking at this code I think this a good candidate to be made into a class as all the functionality "hangs together". A short wrapper function could then be written which would make processing a single file straight-forward and in turn this process could then be run on multiple files in parallel when a directory is given for processing.

I'll create issues for these two tasks shortly.


bobturneruk:
Hi @ns-rse - thanks for this! I'm a bit confused about the branches etc. This is probably my fault for dissecting @SylviaWhittle 's work in the way I did. 

I suspect that what we want to do here is merge [4bbff36](https://github.com/AFM-SPM/TopoStats/pull/88/commits/4bbff361692e17822037063c6991515d19a2ad04) and [6f82048](https://github.com/AFM-SPM/TopoStats/pull/88/commits/6f82048ab1de69bc70dba7b490d316e4fe05c691) into [raster-filters](https://github.com/AFM-SPM/TopoStats/tree/raster-filters), then PR [raster-filters](https://github.com/AFM-SPM/TopoStats/tree/raster-filters) against [dev](https://github.com/AFM-SPM/TopoStats/tree/dev).

Does that make sense? It barely makes sense to me! Shall I have a go at doing it?

ns-rse:
@bobturneruk Yes, I think we're on the same page here. I selected to merge this PR from the `dev-pySPM-tests` branch into `raster-filters` rather than straight to `deve` as I get the impression `raster-filters` is still a work in progress (new function added I had to incorporate and I think making a Class would be useful too). Merges to `dev` can then be made from `raster-filters` as required.

bobturneruk:
I had hoped we could get `raster-filters` reviewed and merged quite quickly. But maybe it's best to get it into a class first? As it stands, this PR will pull quite a lot of stuff from `dev-pySPM-tests` into `raster-filters`. Please either:

- Go ahead and merge this one without review and we can review `raster-filters` against `dev` later, or
- Shift a subset of stuff into `raster-filters` to get that reviewed against `dev`

ns-rse:
There's no reason I can see to hold up merging into dev with the suggestion of refactoring to a class (that can be a separate issue) so I've gone ahead and merged.

Issue #95, opened 2022-04-04
Filters.py/row_col_quantiles() : Binary_mask argument detection not working 


SylviaWhittle:
In `filters.py` , in the function `row_col_quantiles(image: np.array, binary_mask: bool = False)`, the binary mask checking no longer works, since later on, the script checks if a mask was provided using: if `binary_mask:`, which fails if `binary_mask` is of type `np.ndarray`. I am not sure how to solve this, as this argument needs to be able to take two types of argument, and then the script needs to be able to detect if a value is passed with the function for that argument versus if it has not. The previous solution I had for this was to use: `… , binary_mask = None`, along with `if binary_mask is not None`. We could use type checking to see if the `binary_mask` argument is of type `np.ndarray` , however this would still not allow for good type-hints. 

We could use `Unison()`  to allow for more than one type to be hinted, however this requires an import and may not be best practice.

bobturneruk:
@ns-rse - Is this something you can work on, or do you need more info?

ns-rse:
Sorry for not clocking this earlier.

Apologies for the mistake here, I can fix this. Where abouts is the failure occurring with `binary_mask` being an `np.ndarray` @SylviaWhittle ?

I'd noticed the `binary_mask` was used in a few different functions and it's one of the reasons I thought that converting this to a class would be useful here. If an object is binary (which it is here since it is tested for being `None` or anything else) then its best to have it as a `boolean`.

I've been working on the refactoring `filters.py` functionality to a class and starting to get things working on the multiprocessing and just pushed my work so far (see [`filters.py`](https://github.com/AFM-SPM/TopoStats/blob/ns-rse/90-refactor-filters/topostats/filters.py) and [`topotracing_v2.py`](https://github.com/AFM-SPM/TopoStats/blob/ns-rse/90-refactor-filters/topostats/topotracing_v2.py)), but clearly I have to get the logic correct so I'll go back and make sure I do that.

ns-rse:
Thanks @SylviaWhittle I'll have a go at running it and investigate and let you know how it goes.

ns-rse:
Ok, I see whats happening now and I'd not clocked everything on first look, this is essentially my notes as I walk through understanding it and fixing my mistake. I recreated the error and get...

```(topostats) ❱ python topostats/topotracing.py 
2022-04-05--18-12-15
Traceback (most recent call last):
[...]

PR #96, opened 2022-04-05
Correcting issues with masking introduced by my refactoring

ns-rse:
As per #95 I broke `topotracing.py` when I refactored `filters.py`.

As detailed in discussion with @SylviaWhittle on #95 the second flattening will _always_ require masking and that has now been addressed.

This has run successfully on my machine.


ns-rse:
Hmm, I perhaps should have done some cherry picking before creating this pull request, hadn't clocked all those commits and thought it was only the recent changes from @SylviaWhittle I was incorporating.

Whether I resolve the merge conflicts tomorrow will depend on whether I should go back through and cherry pick the commits to push (which will mean I don't need to resolve the conflicts).

Sorry for the mistake.

ns-rse:
I see my mistake, I didn't change the target branch to `dev`. Closed and will pay closer attention.

bobturneruk:
You can change this on this PR by clicking edit up top right (this may not be news!).

PR #97, opened 2022-04-05
Correcting issues with masking introduced by my refactoring

ns-rse:
As per #95 I broke topotracing.py when I refactored `filters.py`.

As detailed in discussion with @SylviaWhittle on #95 the second flattening will always require masking and that has now been addressed.

This has run successfully on my machine.

Not sure how I've picked up commits from 4 months ago though?


ns-rse:
Sorry, I thought I'd got this the wrong way round and was trying merge `dev` > `dev-pySPM-tests` based on the above resolution of merge conflicts but that isn't the case, reopening.

(For the past four years I've mainly used BitBucker with a smattering of GitLab, still finding my way around GitHub).

ns-rse:
Ok, now I understand what I've done wrong I'm closing this and will sort out the mess I've created, sorry.

PR #99, opened 2022-04-06
95 | Further reversions of the mistake I introduced to binary_mask

ns-rse:
I've corrected the mess I made and reverted the `binary_mask` to its original state across all functions but added type-hints and made the existing tests explicit that they are not using `binary_mask`. Now I understand this better I'll extend the tests for each of these functions to cover masking (hence the `FIXME` note).

Really sorry for this mistake.


bobturneruk:
Ayup! Just wanted to check this is targeting the right branch. Is it intended for `dev`, or not yet?

SylviaWhittle (approved):
This looks good :) Didn't know you can assign None to a type hinted variable

ns-rse:
> Ayup! Just wanted to check this is targeting the right branch. Is it intended for dev, or not yet?

It should be yes, I'll close it for now and work out what I need to cherry pick to merge into `dev` as if I try to merge `ns-rse/95` there are a lot of other commits that would be merged.

PR #106, opened 2022-04-13
105 | adds additional versions of Python for tests to run under

### LINKED ISSUE #105: Add newer Python versions to GitHub actions testing stages
Currently GiHub actions only tests the code under Python 3.7 on three different architectures.

We should expand this to testing under different major versions of Python to ensure comparability, these are...

* 3.8
* 3.9
* 3.10

ns-rse:
Resolves #105


bobturneruk (approved):
Looks good, and a sensible change to ensure topostats works the same on multiple Python versions. It may pragmatic to limit test platforms in future if things end up taking a long time to run. This should take users' favoured platforms into account.

PR #107, opened 2022-04-13
Cropped

MaxGamill-Sheffield:
Merging straight to master as this feature was already present but just commented out. Now I have added a config file bit for it to be turned on and off.


MaxGamill-Sheffield:
Steps to test:
1. Add `saveCroppedFigures_option = 1` to the config.ini file (similar to what is in the default_config.ini file, but default behaviour is 0).
2. Run pygwytracing.py
3. See that another folder called "Cropped" has been made which contains images of single molecules, in the correct colour scheme chosen.

PR #134, opened 2022-05-12
Tests and refactoring of grainstats.py

ns-rse:
This Pull Request contains tests and refactoring of the `grainstats.py` that @SylviaWhittle has worked on. I've broken some of the steps out into individual methods and tidied a few bits up here and there.

The intention is that merging this into `dev` will allow @SylviaWhittle and/or @MaxGamill-Sheffield to address each of the changes that improves the statistics by creating new branches from `dev` to address each change, updating or adding tests as required and I will help with this.


ns-rse:
I got massively confused trying to resolve the rebase and so created a fresh branch from current `dev` and pulled the `grainstats.py` refactoring and tests over at the same time correcting a mistake in the scale where I had used `/` instead of `*`.

PR #155, opened 2022-05-19
90 | Refactors filters and grains submodule functions to classes

### LINKED ISSUE #90: Refactor topostats/filters.py to a class
The functions defined in `topostats/filters.py` that replace some of the dependency on Gwyddion are a good candidate for refactoring into a class which can then be extended with additional methods such as the recently added `remove_x_bowing()` and subsequent processing steps that depend on Gwyddion.

ns-rse:
Resolves #90 

...and more!

Refactoring of functions in under `topostats/filters.py` and `topostats/grains.py` into classes.

It also introduces a dictionary for image filenames and titles to simplify/shorten the output of plots and for the time being a clause which only prints images of grains if grains are found.


SylviaWhittle (approved):
Great improvements, very useful OOP implementation. No problems found.

PR #163, opened 2022-05-26
Basic Flow Diagram

ns-rse:
Starts adding flow-diagrams to documentation.

Should build on ReadTheDocs too (does so locally, hopefully dependencies are pulled in).


ns-rse:
Stupidly didn't set this to merge into `dev`.

PR #166, opened 2022-05-30
Aligns grainstats output with dnatracing input

### LINKED ISSUE #156: Align output from new topotracing.py with dnatracing.py
With the removal of depending on Gwydion we are now at the next step of ensuring the output produced through processing with scikit-image is suitable to be passed on to next step in analysis which is conducted with `tracing/dnatracing.py`.

The `dnaTrace` class takes the following arguments...

* `full_image_data` 
* `gwyddion_grains`
* `afm_image_name`
* `pixel_size`
* `number_of_columns`
* `number_of_rows`

...and so we need to provide those from the output of `topotracing.py`.

To achieve this I will run the old `pygwytracing.py` based code and determine what each of these are and how to extract them from scikit-image pipeline.

ns-rse:
Resolves #156 

We had to ensure that the output from the new `topotracing` can be passed on and used as input to the `dnaTrace()` class which performs DNA tracing.

This pull request achieves that, although there is still a fair amount of work to be done on the `dnatracing.py` module as a whole as there are no unit tests to check the work done here.

At the same time I have been made aware that the results of DNA tracing are used in the `Plotting` module to simplify the process of visualising results and so I have ensured that the output from `GrainStats()` and `dnaTrace()` are combined for each image (and saved within each images output directory) and that these are then combined across all images that are found and output to a single CSV file in the config/user specified output directory.

The `Plotting.py` module will also need some attention after tests and refactoring of `dnatracing.py`.

### DIFF
--- a/run_topostats.py
+++ b/run_topostats.py
@@ -1,5 +1,6 @@
 """Topotracing"""
 import argparse as arg
+from collections import defaultdict
 from functools import partial
 from multiprocessing import Pool
 from pathlib import Path
@@ -183,11 +184,12 @@ def process_scan(
     tracing_stats.saveTraceStats(Path(output_dir) / filtered_image.filename)
 
     # Combine grainstats and tracingstats
-    results = grain_statistics["statistics"].merge(tracing_stats.pd_dataframe, on="Molecule Number", indicator=True)
+    results = grain_statistics["statistics"].merge(tracing_stats.pd_dataframe, on="Molecule Number")
     results.to_csv(output_dir / filtered_image.filename / "all_statistics.csv")
 
     # Optionally plot all stages
     if save_plots:
+        LOGGER.info("Saving plots of images at all stages of processing.")
         # Filtering stage
         for plot_name, array in filtered_image.images.items():
             if plot_name not in ["scan_raw", "extracted_channel"]:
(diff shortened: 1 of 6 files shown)


run_topostats.py
```diff
@@ -1,9 +1,13 @@
 """Topotracing"""
```
bobturneruk, line 1: Should `run_topostats.py` be part of the package?

run_topostats.py
```diff
@@ -162,19 +167,37 @@ def process_scan(
         labelled_data=grains.images["labelled_regions"],
         pixel_to_nanometre_scaling=filtered_image.pixel_to_nm_scaling,
         img_name=filtered_image.filename,
-        output_dir=output_dir,
+        output_dir=output_dir / filtered_image.filename,
     )
     grain_statistics = grainstats.calculate_stats()
 
+    # Run dnatracing
```
bobturneruk, line 174: Would it be wise to make the dna tracing step optional via the config file / command line options for now? I guess we're still deciding how to ultimately control the workflow, but I think in the short term there are several use cases where dna tracing is not needed.

bobturneruk (approved):
Looking good! A couple of questions in there.

run_topostats.py
ns-rse: As the steps being run had expanded outside of the initial "Topotracing" work to include the code and steps in `dnatracing.py` I moved and rename `topostats/topotracing.py` > `run_topostats.py`.

In light of your pointer about [entry points](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) (see also [here](https://setuptools.pypa.io/en/latest/userguide/entry_point.html) for documentation and doing this with `setup.cfg`) I've just created #170 to add this feature.

run_topostats.py
ns-rse: Should definitely make workflow flexible and happy to take guidance from users on what they would like to control.

I didn't do that in this work because `pygwytracing.py` appears to always run `dnatracing` after its extracted data using Gwyddion.

PR #167, opened 2022-05-30
Fix issue with unit conversion 

Jean-Du:
There has been an issue with unit conversion (due to direct reference of dataframes instead of copying them) when making multiple plots from the same data. This pull request fixes this and also clears up some redundant comments.


slackline (approved):
At some point `Plotting.py` will need refactoring, but for now happy to approve this.

ns-rse (approved):
At some point `Plotting.py` will need refactoring, but for now happy to approve this.

PR #173, opened 2022-06-08
170 | Adds an entry point for run_topostats

### LINKED ISSUE #170: Create an entry point for Topostats
An [entry point](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) provides a convient way of including a command line program for users to utilise when installing a package.

Currently `run_topostats.py` is the script that should be called in order to run topotracing and dnatracing using a user specified configuration file.

```
python run_topostats.py -c config/example.yaml
```

We should move this script to `topostats/run_topostats.py` and create an entry point, something along the lines of the following in `setup.cfg`

```
[options.entry_points]
my.plugins =
    run_topostats = topostats/run_topostats
```

See also [here](https://setuptools.pypa.io/en/latest/userguide/entry_point.html) for documentation and doing this with `setup.cfg`.

ns-rse:
Resolves #170 

To ease usage of topostats this pull requests adds an [entry point](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) for running TopoStats.

After (re-)installing in a virtual environment users can now run topostats at the command line by calling `run_topostats`. It works as the `run_topostats.py` script did and takes various arguments, the key required one being the path to a configuration file. To use it would be...

```
```


SylviaWhittle:
In `README.md`, to run topostats it says to use this command: `python ./run_topostats.py --config config/example.yaml`, whereas it seems this should be `python ./topostats/run_topostats.py --config config/example.yaml`. Just a tiny suggestions @ns-rse, otherwise, it looks good and works on my machine.

SylviaWhittle (approved):
Works for me, and is an excellent ease-of-use improvement, thanks @ns-rse.

PR #177, opened 2022-06-13
176 | Adds a test to GitHub tests.yaml for CLI entry point

### LINKED ISSUE #176: Add a step to GitHub Actions to ensure the CLI entry-point works
With the addition of an entry point (#170) to provide a command line interface (CLI) it would be useful to have a "smoke-test" to ensure this works as part of the GitHub actions.

**NB** [Smoke testing (software)](https://en.wikipedia.org/wiki/Smoke_testing_(software)) is a method of checking whether something works, if it doesn't its broken and needs fixing but not a huge amount of detail is indicated as to where the problem is.

ns-rse:
Resolves #176

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,7 +18,7 @@ install_requires =
   pandas
   pySPM
   ruamel.yaml
-  scikit-image
+  scikit-image==0.19.2
   scipy
   seaborn
   tqdm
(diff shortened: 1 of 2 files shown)


ns-rse:
> Looks good! Why are we pinning the scikit-image version so precisely, please?

Pinning because the new release broke the tests, I've detailed it in #178 to be investigated and addressed once the `ChangeLog` is available it wasn't when I discovered the problem. I did mention this PR in #178 so it shows up in the history but its not the best way of linking things, I'll be more explicit next time.

Issue #111, opened 2022-04-22
Metadata file output


LunchBobs:
**Is your feature request related to a problem? Please describe.**
Analysis performed is a bit of a black box. There's no way to determine what parameters were used when going back to data previously analysed (unless meticulous notes were kept at the time). 
- No list of functions performed and parameters used
- No option to select which functions are used
- Some images are not analysed, but don’t know which ones

**Describe the solution you'd like**
A clear tree of functions that were performed on the data, in the order they were performed and what parameters were used will be great. The file should also list the names of the images analysed, those that were not analysed and the number of images analysed in total. It could also function as a log of any errors the software encountered during processing.

**Describe alternatives you've considered**

It might be possible to copy the config file as a metadata file into the folder that it puts the processed images in as a short term solution. It wouldn't have everything you'd want but at least there'd be the values you'd used in the config in there. Nanoscope analysis has a visual version of this onscreen when you open the image, that might be one way to do it. 

**Additional context**

benthomasuos:
I've had a think about this alongside what could be done to open up the data formats from Bruker's incomprehensible binary formats.
HDF5 files could store the raw data directly from the numpy arrays that PySPM generates. The config data could be stored in these files as well as particular settings for each function that is run on the data. It could also store the analysis/grain stats in a standardised format for subsequent labelling and meta-analysis. The resultant ".h5" files could then be shared amongst AFM researchers and used with other programming languages other than Python, e.g. R or Matlab

The only downside is that HDF5 isn't a text format so you can't just open up the file and view the config or metadata in a program like notepad. We would need to create some utility functions to cope with that so it would either output the config and/or metadata to a text file or just to the command line. There are HDF5 GUI programs, which could be used for viewing the data too.

Creating a HDF5 intermediary file could sit in the current `dev` branch workflow between reading the .spm file with PySPM and the first action of filters. It could be a class that provides helpful get/set methods like `.get_height_array()` `.get_grain_labels()` and `.set_height_array()` `.set_grain_labels()`

bobturneruk:
Thanks @benthomasuos - here's a riff on that, it's hack, let's say it's pseudo code:

```{python}
import re

import h5py
import numpy as np
from pySPM.Bruker import Bruker

minicircle = Bruker('minicircle.spm')

with h5py.File("minicircle.hdf5", "w") as f:
    for metadata, value in minicircle.scanners[0].items():
        f.attrs.create(metadata, value)

    for layer in minicircle.layers:
[...]

PR #180, opened 2022-06-16
159 | Introduces correct "raw_height" plot; adds write_yaml()

### LINKED ISSUE #159: Dev: "01-raw_heightmap.png" plots processed data instead of raw height data
**Describe the bug**
The output plot produced to show the raw heightmap (01-raw_heightmap.png) shows data that looks like it has undergone some processing already and thus is not the raw height data from the SPM file.

**To Reproduce**

- Command used to run topostats e.g. `python pygwytracing.py` / `python topostats/topotracing.py`
- Config file (original config file used)
- View output image "01-raw_heightmap.png"

**Expected behavior**
The raw AFM data before any processing should be shown in this image.

**Output**
Output image attached below.
01-raw_heightmap

 - OS: MacOS
 - topostats version: bf56931 (HEAD -> dev, origin/dev) Merge pull request #155 from AFM-SPM/ns-rse/90a-refactor-classes.
 - Python version: 3.8.13.
 - Packages attached.
[packages.txt](https://github.com/AFM-SPM/TopoStats/files/8770764/packages.txt)

**Additional context**
N/A

ns-rse:
Resolves #159 

Plots the `raw_height` of the Bruker image to `00-raw_height.png` resolving #159.
The reason I'd not included this previously is I was confused why the `pySPM.SPM` that the `extract_channel()` plots couldn't be plotted, its because this object has the attribute `pixels` which is what should be plotted.

That said I'm not entirely sure why this, which is the plot of `pySPM.SPM.pixels` differs from the plot of pixels as I can't see why/where the `self.images["pixels"]` is modified in comparison to the extraction and so the two should match. :shrug: 

I've also introduced a short function `topostats.io.write_yaml()` that writes the configuration used to YAML in the output directory which is an interim step to addressing #111 but not an absolute solution.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -1,11 +1,12 @@
 """Functions for reading and writing data."""
 import logging
+from datetime import datetime
 from pathlib import Path
 from typing import Union, Dict
 
 from pySPM.Bruker import Bruker
-import ruamel.yaml
 from ruamel.yaml import YAML, YAMLError
+from ruamel.yaml.main import round_trip_load as yaml_load, round_trip_dump as yaml_dump
 from topostats.logs.logs import LOGGER_NAME
 
 LOGGER = logging.getLogger(LOGGER_NAME)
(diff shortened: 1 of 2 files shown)


SylviaWhittle (approved):
All good! Works for me and I think getting the metadata into the YAML file early on in development is a good idea. 

I am also clueless as to the issues with pySPM and plotting that you outlined, apologies. I remember something about encountering a similar problem in the past. 

I've left a comment in `run_topostats().py`, line `197` where warnings are displayed in the console. Not a problem, but we might want to clean up the console outputs or change the line.

PR #184, opened 2022-06-17
183 | add a regression test for dnatracing against minicircle.spm

ns-rse:
Before starting work on refactoring `dnatracing.py` I've added a "regression test" which checks what the overall output of the process should be and am putting it out for PR/review so that they are smaller and easier to review/understand.

The purpose is that as I go about refactoring the code I run this test, if it fails I've made a mistake and have incorrectly changed the process.


ns-rse:
Of course it would help if I didn't submit mistakes wouldn't it! Sorry, will correct now.

I've also just realised that I face some problems because I can't use `tmpdir` as the output directory that is stored in the `Basename` column of the CSV since that is going to change depending on where/when the test is run. :thinking:

ns-rse:
That took longer than I thought it would to resolve but tests now pass here on GitHub. Apologies for the "noise".

tests/test_dnatracing.py
```diff
@@ -0,0 +1,63 @@
+"""Tests of dnatracing methods."""
+import numpy as np
+import pytest
+
+
+def test_get_grain_array(test_dnatracing) -> None:
+    """Test extraction of grain array."""
+    TARGET = {
```
bobturneruk, line 8: Would this be better as a fixture? If only to future-proof?

tests/conftest.py
```diff
@@ -397,3 +398,48 @@ def minicircle_grainstats(
 def minicircle_grainstats_20220526() -> pd.DataFrame:
     """Statistics for minicircle for comparison."""
     return pd.read_csv(RESOURCES / "minicircle_grainstats_20220526.csv", index_col=0)
+
+
+# Derive fixtures for DNA Tracing
+GRAINS = np.array(
```
bobturneruk, line 404: Should this be a fixture itself? Maybe not.

tests/conftest.py
```diff
@@ -397,3 +398,48 @@ def minicircle_grainstats(
...
+        [0, 1, 0, 0, 1, 0, 0, 0, 0, 0, 2],
+        [0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 2],
+        [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2],
+        [0, 0, 3, 3, 3, 3, 3, 0, 0, 0, 2],
+        [0, 0, 3, 0, 0, 0, 3, 0, 0, 0, 2],
+        [0, 0, 3, 3, 3, 3, 3, 0, 0, 0, 2],
+        [0, 0, 4, 4, 4, 4, 4, 0, 0, 0, 2],
+    ]
+)
+FULL_IMAGE = RNG.random((GRAINS.shape[0], GRAINS.shape[1]))
```
bobturneruk, line 417: Should this be a fixture itself? Maybe not.

tests/test_dnatracing.py
bobturneruk, line 8: Should this be a fixture itself? Maybe not.

topostats/run_topostats.py
```diff
@@ -176,16 +176,16 @@ def process_scan(
     dna_traces = dnaTrace(
         full_image_data=grains.images["gaussian_filtered"].T,
         grains=grains.images["labelled_regions"],
-        afm_image_name=filtered_image.filename,
+        filename=filtered_image.filename,
         pixel_size=filtered_image.pixel_to_nm_scaling,
-        number_of_columns=grains.images["labelled_regions"].shape[0],
-        number_of_rows=grains.images["labelled_regions"].shape[1],
     )
+    dna_traces.trace_dna()
```
bobturneruk, line 182: Sorry if I've asked this before, but should it be a configurable option to run this? Maybe a future decsision.

tests/test_dnatracing.py
```diff
@@ -0,0 +1,63 @@
...
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                [0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0],
+            ]
+        ),
+    }
+    for x in range(1, 4):
+        np.testing.assert_array_equal(test_dnatracing._get_grain_array(x), TARGET[x])
```
bobturneruk, line 63: I've been outPythoned by this line! Maybe this isn't the forum, but it looks to me like `_get_grain_array` is a "private member of `dnatracing`, yet here it is apparently being called from `test_dnatracing`. Seems to work, but I'm curious!

bobturneruk: Ah, I get it now. There's an instance of something called `test_dnatracing`. I was confused by the filename. Silly me!

bobturneruk (approved):
Looks good and the initial refactoring seems to be going in a good direction.

tests/test_dnatracing.py
ns-rse: In `tests/test_dnatracing.py` I'm focusing on isolated testing of each method so there shouldn't be any need for each to be a `@pytest.fixture`.

However, I've setup a concurrent `tests/test_dnatracing_minicircle.py` which will take and test the `@pytest.fixture`'s defined in `tests/conftest.py` where there will be a "cascading" trail of fixtures, each built on the previous fixture. At the moment I've not started that cascade (although could have included one for `_get_grain_array()`. I'll make sure I add it in the next PR.

tests/conftest.py
ns-rse: No, `@pytest.fixture` apply to functions, whilst `GRAINS` is a constant hence it being upper case.

tests/conftest.py
ns-rse: No, `@pytest.fixture` apply to functions, whilst `GRAINS` is a constant hence it being upper case.

tests/test_dnatracing.py
ns-rse: Not the clearest of naming conventions, sorry.

You've clocked it correctly though, `_get_grain_array()` is indeed a private method of the `dnaTracing` class that takes one argument, the number of the grain, and returns an array of 0/1 for that grain alone. This loop therefore runs over all the grains and checks the correct array is returned for each grain.

tests/test_dnatracing.py
ns-rse: Think this is the same as comment above, as these are unit-tests testing methods in isolation then I don't envisage needing them to be fixtures. If fixtures are needed I'll define them in `tests/conftest.py`.

topostats/run_topostats.py
ns-rse: Its a good point and we should look at making the end-point of processing optional, I've documented it as an issue to be addressed (see #185 ).

PR #187, opened 2022-06-24
183 | Handling single grains

ns-rse:
In investigating #183 I've discovered a problem related to thresholds and filtering when only a single grain is found.

The image `20220601_NDP52_FL_10ng_PLO_HEPES_20mM_NaCl_50mM.0_00021.spm` crashed with the error...

```bash
IndexError: cannot do a non-empty take from an empty axes.
```

This arises from the `Grains` class and is caused when the method `.calc_minimum_grain_size()` is called and only one grain has been detected. The code calculates the minimum grain size by first calculating the area of all grains...

```python
        grain_areas = np.array([grain.area for grain in self.region_properties])
```

Then only a subset of grains were retained when the area exceeded a threshold...

```python
        grain_areas = grain_areas[grain_areas > threshold(grain_areas, method=self.threshold_method)]
```

Finally the `minimum_grain_size` was calculated as the median of the remaining grain sizes minus 1.5 x the Interquartile Range
```python
        self.minimum_grain_size = np.median(grain_areas) - (
            1.5 * (np.quantile(grain_areas, 0.75) - np.quantile(grain_areas, 0.25))
        )

```

The problem is that with only a single grain the threshold is the same size as the one grain that has been detected and because the inequality for retaining grains was that its area had to _exceed_ the threshold (i.e. `>`) then the grain was removed. This in turn meant that `grain_areas` was `None` (empty) and you can't calculate the median of an empty array!

The solution I've gone for here is to change the inequality so that the only grains smaller than the threshold are excluded, those that are equal are retrained, hence the very small change in this Pull Request.

It means I can now process five of the images @derollins shared with me and I've another error to investigate :smiley:

### DIFF
--- a/topostats/grains.py
+++ b/topostats/grains.py
@@ -126,7 +126,10 @@ class Grains:
         """
         self.get_region_properties()
         grain_areas = np.array([grain.area for grain in self.region_properties])
-        grain_areas = grain_areas[grain_areas > threshold(grain_areas, method=self.threshold_method)]
+        print(
+            f"threshold(grain_areas, method=self.threshold_method) : {threshold(grain_areas, method=self.threshold_method)}"
+        )
+        grain_areas = grain_areas[grain_areas >= threshold(grain_areas, method=self.threshold_method)]
         self.minimum_grain_size = np.median(grain_areas) - (
             1.5 * (np.quantile(grain_areas, 0.75) - np.quantile(grain_areas, 0.25))
         )
--- a/topostats/grains.py
+++ b/topostats/grains.py
@@ -93,7 +93,7 @@ class Grains:
     def label_regions(self, image: np.array) -> np.array:
         """Label regions.
 
-        This method is used twice, once prior to removal of small regions, and again afterwards, hence requiring and
+        This method is used twice, once prior to removal of small regions, and again afterwards, hence requiring an
         argument of what image to label.
 
         Parameters
@@ -190,9 +190,14 @@ class Grains:
         self.get_mask()
         self.tidy_border()
         self.label_regions(self.images["tidied_border"])
-        self.calc_minimum_grain_size(image=self.images["labelled_regions"])
-        self.remove_small_objects()
-        self.label_regions(self.images["objects_removed"])
-        self.get_region_properties()
-        self.colour_regions()
-        self.get_bounding_boxes()
+        try:
+            self.calc_minimum_grain_size(image=self.images["labelled_regions"])
+            self.remove_small_objects()
+            self.label_regions(self.images["objects_removed"])
+            self.get_region_properties()
+            self.colour_regions()
+            self.get_bounding_boxes()
+        # FIXME : Identify what exception is raised with images without grains and replace broad except
+        except Exception as e:
+            LOGGER.info(f"[{self.filename}] : No grains found.")
+            pass
--- a/topostats/grains.py
+++ b/topostats/grains.py
@@ -126,9 +126,6 @@ class Grains:
         """
         self.get_region_properties()
         grain_areas = np.array([grain.area for grain in self.region_properties])
-        print(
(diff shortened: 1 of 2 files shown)


bobturneruk:
So `minimum_grain_size` is not the size of the smallest grain, but the size of the hypothetical smallest grain to be included in further analysis?

topostats/grains.py
```diff
@@ -126,7 +126,7 @@ def calc_minimum_grain_size(self, image: np.array) -> float:
         """
         self.get_region_properties()
         grain_areas = np.array([grain.area for grain in self.region_properties])
-        grain_areas = grain_areas[grain_areas > threshold(grain_areas, method=self.threshold_method)]
+        grain_areas = grain_areas[grain_areas >= threshold(grain_areas, method=self.threshold_method)]
```
bobturneruk, line 129: Nice when it's this simple!

topostats/grains.py
```diff
@@ -190,9 +190,14 @@ def find_grains(self):
...
+        try:
+            self.calc_minimum_grain_size(image=self.images["labelled_regions"])
+            self.remove_small_objects()
+            self.label_regions(self.images["objects_removed"])
+            self.get_region_properties()
+            self.colour_regions()
+            self.get_bounding_boxes()
+        # FIXME : Identify what exception is raised with images without grains and replace broad except
+        except Exception as e:
+            LOGGER.info(f"[{self.filename}] : No grains found.")
```
bobturneruk, line 202: I guess this is pythonic, but logging an exception as info seems odd.

topostats/grains.py
```diff
@@ -190,9 +190,14 @@ def find_grains(self):
         self.get_mask()
         self.tidy_border()
         self.label_regions(self.images["tidied_border"])
-        self.calc_minimum_grain_size(image=self.images["labelled_regions"])
-        self.remove_small_objects()
-        self.label_regions(self.images["objects_removed"])
-        self.get_region_properties()
-        self.colour_regions()
-        self.get_bounding_boxes()
+        try:
```
bobturneruk, line 193: Pythonic, but using try catch rather than pre-empting the error always makes me feel uncomfortable! Looks good.

bobturneruk: I guess 194 is going to actually throw the error and, if so, we don't want 195-199 to execute so they're included in the block.

topostats/grains.py
ns-rse: Its not that pythonic, if I had linted it I would have been told the exception is too broad.  But this is just a place holder for now, I've requested images that have no grains and will see what exceptions are actually raised and handle them appropriately (mostly it will be a pass but I figured logging some information to say why certain steps aren't carried out would be informative).

ns-rse:
> So `minimum_grain_size` is not the size of the smallest grain, but the size of the hypothetical smallest grain to be included in further analysis?

Yes, its to get rid of noise and sets the a "threshold" for which anything below that should be excluded.

Its the difference between this which has lots of small noisy bits in...

image

...and this which has had them removed...

image

topostats/grains.py
ns-rse: My understanding is that its "Pythonic" to use `try: ... except:" and is captured in the [EAPF principle](https://docs.python.org/3.5/glossary.html#term-eafp) "_Easier to ask for forgiveness than permission_" than the contrasting "_Look Before You Leap_" that is characterised by `if: ... else:`.

And spot on, if we don't find any grains then we don't want the remaining steps to be run.

topostats/grains.py
ns-rse: If only everything in life were that simple!

> One thing I would recommend, either as part of this PR, or a later one, would be to include as test to prove that the exception is raised when appropriate data is thrown at it. `with pytest.raises` etc.

:+1: Definitely, I'm awaiting some images with zero grains so I can write tests to identify what exceptions are raised and check them in tests.

PR #189, opened 2022-06-28
186 | Fail gracefully when no grains found

ns-rse:
This is still a work in progress and subject to change when thresholding/filtering is revised, so I will update things when that is available. However it hopefully fails gracefully when no grains are found.

I think long-term we should consider bulking out the test-suite with more than just `minicircle.spm`, a `silicon.spm`, `water.spm` and more will be really useful.

Also tidied up via some linting and applied Black formatting in a few places that hadn't had that treatment.

### DIFF
--- a/tests/test_grains.py
+++ b/tests/test_grains.py
@@ -1,10 +1,14 @@
 """Test finding of grains."""
+import logging
 import numpy as np
 
 # Pylint returns this error for from skimage.filters import gaussian
 # pylint: disable=no-name-in-module
 from skimage.filters import gaussian
 from topostats.grains import Grains
+
+LOGGER = logging.getLogger(__name__)
+LOGGER.propagate = True
 
 # Specify the absolute and relattive tolerance for floating point comparison
 TOLERANCE = {"atol": 1e-07, "rtol": 1e-07}
@@ -21,3 +25,10 @@ def test_gaussian_filter(small_array_grains: Grains, grain_config: dict) -> None
 
     assert isinstance(small_array_grains.images["gaussian_filtered"], np.ndarray)
     np.testing.assert_array_equal(small_array_grains.images["gaussian_filtered"], target)
+
+
+def test_random_grains(random_grains: Grains, caplog) -> None:
+    """Test errors raised when processing images without grains."""
+    # FIXME : Uncomment once filtering excludes all noise (grains < min_size) as we should find no grains.
+    # assert "No grains found" in caplog.text
+    assert True
--- a/topostats/filters.py
+++ b/topostats/filters.py
@@ -37,7 +37,7 @@ class Filters:
             Path to a valid image to load.
         channel: str
             Channel to extract from the image.
-        amplify_level: float
+        amplify_level : float
             Factor by which to amplify the image.
         threshold_method: str
             Method for thresholding, default 'otsu'.
@@ -49,7 +49,7 @@ class Filters:
         Notes
         -----
 
-        A directory under the 'outdir' will be created using the
+        A directory under the 'outdir' will be created using the filename.
         """
         self.img_path = Path(img_path)
         self.channel = channel
@@ -160,7 +160,7 @@ class Filters:
         image = self.remove_tilt(image, mask)
         return image
 
-    def row_col_medians(self, image: np.array, mask: np.array = None) -> np.array:
+    def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarray:
         """Returns the height value medians for the rows and columns.
 
         Returns
@@ -180,7 +180,7 @@ class Filters:
         LOGGER.info(f"[{self.filename}] : Row and column medians calculated.")
         return medians
 
-    def align_rows(self, image: np.array, mask=None) -> np.array:
+    def align_rows(self, image: np.ndarray, mask=None) -> np.ndarray:
         """Returns the input image with rows aligned by median height"""
         medians = self.row_col_medians(image, mask)
         row_medians = medians["rows"]
@@ -202,7 +202,7 @@ class Filters:
         return image
 
     @staticmethod
-    def _median_row_height(array: np.array) -> float:
+    def _median_row_height(array: np.ndarray) -> float:
         """Calculate median of row medians"""
         return np.nanmedian(array)
 
@@ -211,7 +211,7 @@ class Filters:
         """Calculate difference of row medians from the median row height"""
         return row_medians - median_row_height
 
-    def remove_tilt(self, image: np.array, mask=None) -> np.array:
+    def remove_tilt(self, image: np.ndarray, mask=None) -> np.ndarray:
         """Returns the input image after removing any linear plane slant"""
         medians = self.row_col_medians(image, mask)
         gradient = {}
@@ -228,16 +228,16 @@ class Filters:
         return image
 
     @staticmethod
-    def calc_diff(array: np.array) -> np.array:
+    def calc_diff(array: np.ndarray) -> np.ndarray:
         """Calculate the difference of an array."""
         return array[-1] - array[0]
 
     @classmethod
-    def calc_gradient(self, array: np.array, shape: int) -> np.array:
+    def calc_gradient(self, array: np.ndarray, shape: int) -> np.ndarray:
         """Calculate the gradient of an array."""
         return self.calc_diff(array) / shape
 
-    def get_threshold(self, image: np.array, **kwargs) -> float:
+    def get_threshold(self, image: np.ndarray, **kwargs) -> float:
         """Returns a threshold value separating the background and foreground of a 2D heightmap.
 
         Parameters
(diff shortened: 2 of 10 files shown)


bobturneruk:
[3f9001f](https://github.com/AFM-SPM/TopoStats/pull/189/commits/3f9001fb68c124948d8e91079c597d4ee4e16bdc) is a big old commit, just saying.

topostats/filters.py
```diff
@@ -160,7 +160,7 @@ def flatten_image(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarra
         image = self.remove_tilt(image, mask)
         return image
 
-    def row_col_medians(self, image: np.array, mask: np.array = None) -> np.array:
+    def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarray:
```
bobturneruk, line 163: TIL the difference between `array` and `ndarray`. Amazed it worked without this change.

topostats/run_topostats.py
```diff
@@ -69,6 +93,20 @@ def create_parser() -> arg.ArgumentParser:
     parser.add_argument(
         "-s", "--save_plots", dest="save_plots", type=bool, required=False, help="Whether to save plots."
     )
+    parser.add_argument(
+        "--type_of_sample",
+        dest="type_of_sample",
+        type=str,
+        required=False,
+        help="Type of sample being processed.",
```
bobturneruk, line 101: Where do we list allowable values?

topostats/run_topostats.py
```diff
@@ -69,6 +93,20 @@ def create_parser() -> arg.ArgumentParser:
...
+    )
+    parser.add_argument(
+        "--minimum_grain_size",
+        dest="minimum_grain_size",
+        type=str,
+        required=False,
+        help="Minimum size of grains.",
```
bobturneruk, line 108: What range, units? Maybe this is a separate issue for the docs...

topostats/run_topostats.py
```diff
@@ -146,61 +185,77 @@ def process_scan(
...
-        filename=filtered_image.filename,
-        pixel_to_nm_scaling=filtered_image.pixel_to_nm_scaling,
-        threshold_method=threshold_method,
-        gaussian_size=gaussian_size,
-        gaussian_mode=gaussian_mode,
-        threshold_multiplier=threshold_multiplier,
-        background=background,
-    )
-    grains.find_grains()
+    try:
```
bobturneruk, line 188: This is the crux, I guess. Looks sensible.

topostats/run_topostats.py
```diff
@@ -146,61 +185,77 @@ def process_scan(
...
-        data=grains.images["gaussian_filtered"],
-        labelled_data=grains.images["labelled_regions"],
-        pixel_to_nanometre_scaling=filtered_image.pixel_to_nm_scaling,
-        img_name=filtered_image.filename,
-        output_dir=output_dir / filtered_image.filename,
-    )
-    grain_statistics = grainstats.calculate_stats()
+    if len(grains.region_properties) > 0:
+        # Grain Statistics :
+        try:
```
bobturneruk, line 208: Why might this fail if the condition at 206 is passed?

topostats/run_topostats.py
```diff
@@ -146,61 +185,77 @@ def process_scan(
...
-    results.to_csv(output_dir / filtered_image.filename / "all_statistics.csv")
+            # Combine grainstats and tracingstats
+            LOGGER.info(f"[{filtered_image.filename}] : Combining grain statistics and dnatracing statistics")
+            results = grain_statistics["statistics"].merge(tracing_stats.df, on="Molecule Number")
+            results.to_csv(output_dir / filtered_image.filename / "all_statistics.csv")
+        # Possible exceptions
+        # ValueError : Arises when lots of small grains with small areas
+        except Exception as broad_exception:
+            # If no results we need a dummy dataframe to return.
+            LOGGER.info("Errors occurred attempting to calculate grain statistics and dnnatracing statistics.")
```
bobturneruk, line 240:
```suggestion
            LOGGER.info("Errors occurred attempting to calculate grain statistics and dna tracing statistics.")
```

topostats/thresholds.py
```diff
@@ -17,17 +18,12 @@ def threshold(image: np.array, method: str = 'otsu', **kwargs: dict) -> float:
...
-    --------
-    FIXME: Add docs.
-
     """
     thresholder = _get_threshold(method)
     return thresholder(image, **kwargs)
 
 
-def _get_threshold(method: str = 'otsu'):
+def _get_threshold(method: str = "otsu"):
```
bobturneruk, line 26: Just noticed there's no return type here. I guess functions aren't somehow massaged into being a "type".

tests/test_grains.py
```diff
@@ -21,3 +25,10 @@ def test_gaussian_filter(small_array_grains: Grains, grain_config: dict) -> None
 
     assert isinstance(small_array_grains.images["gaussian_filtered"], np.ndarray)
     np.testing.assert_array_equal(small_array_grains.images["gaussian_filtered"], target)
+
+
+def test_random_grains(random_grains: Grains, caplog) -> None:
+    """Test errors raised when processing images without grains."""
+    # FIXME : Uncomment once filtering excludes all noise (grains < min_size) as we should find no grains.
+    # assert "No grains found" in caplog.text
+    assert True
```
bobturneruk, line 34: Is this to be fixed prior to merge, or post?

bobturneruk (commented):
Wasn't sure if this was ready, so just some comments for now.

topostats/run_topostats.py
ns-rse: Damn, I meant to remove this as @SylviaWhittle said she had addressed the minimum size based on sample in her thresholding work. I pulled out lots of other changes I'd made but forgot to do so with the arguments, now removed.

topostats/filters.py
ns-rse: I know, I hadn't realised the [difference](https://stackoverflow.com/a/15879527) myself really.

topostats/thresholds.py
ns-rse: Thanks for picking this up, hadn't thought about this myself so [looked it up](https://stackoverflow.com/questions/37835179/how-can-i-specify-the-function-type-in-my-type-hints). The return type here should be `Callable`, duly added.

tests/test_grains.py
ns-rse: Post, its contingent on @SylviaWhittle work which will include removing all "noise" when things are below a certain threshold.  Currently we _always_ get some grains found even on water samples. For example the `random` image used in places of this test suite results in the following HUGE number of grains being found which clearly doesn't make any sense!

image

topostats/run_topostats.py
```diff
@@ -146,61 +185,77 @@ def process_scan(
...
         for plot_name, array in filtered_image.images.items():
             if plot_name not in ["scan_raw"]:
                 if plot_name == "extracted_channel":
                     array = np.flipud(array.pixels)
                 PLOT_DICT[plot_name]["output_dir"] = Path(output_dir) / filtered_image.filename
-                plot_and_save(array, **PLOT_DICT[plot_name])
+                try:
+                    plot_and_save(array, **PLOT_DICT[plot_name])
+                except AttributeError:
+                    pass
```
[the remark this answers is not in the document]
ns-rse: Not sure, my thinking is that they are informed elsewhere that no grains have been found there and so not all plots would be produced.

I think more broadly the images I'm generating may be overkill anyway and so a discussion on what plots are actually useful should be had.

For now I've added a `LOGGER.info()` to state which plot couldn't be generated when this fail.

Thanks for pointing this out.

topostats/run_topostats.py
ns-rse: One instance I've found so far is a `ValueError` is raised as a consequence of some very small grains being detected.  At the moment these are captured with a very broad `Exception` but I expect the `ValueError` I have found to not occur once we remove "noise" based on a minimum grain size.

But I suspect there is probably a lot of other scenarios that might cause errors and would be keen to introduce regression tests for more sample types, `silicon`, `water`, `proteins` and so forth to be able to get more specific exceptions defined and handled.

For now though I've left the broad `try: .... except:` that returns an empty data frame of statistics for that image so that when processing multiple images it "fails gracefully" if one causes a problem.

ns-rse:
> [3f9001f](https://github.com/AFM-SPM/TopoStats/pull/189/commits/3f9001fb68c124948d8e91079c597d4ee4e16bdc) is a big old commit, just saying.

Yes, sorry about that I should have split it into smaller pieces for each module that was being modified.  Thanks for going through it though, I know these things aren't straight-forward.

bobturneruk:
Is there a link to an image that I can test with to see the graceful failure in action, please?

ns-rse:
Hi @bobturneruk 

Apologies for the delay in getting back to you.

> Is there a link to an image that I can test with to see the graceful failure in action, please?

If you have access to the folder @derollins shared ([this one](https://drive.google.com/drive/folders/1FQ-iI_x8lkuQk7fA5_5UEKbhSifmQTfi)) then the file `20220601_blank_PLL_water.0_00002.spm` will crash out on the `dev` branch, but should complete on `ns-rse/186-fail-gracefully` branch (which this PR seeks to merge into `dev`).

Let me know if you don't have access and I can provision the file another way.

PR #192, opened 2022-07-04
191 | Note on Anaconda version 2022.05 for M1 architecture

ns-rse:


SylviaWhittle:
@ns-rse I'm not sure if this is important or relevant, but I am developing TopoStats on an M1 architecture SOC and I haven't needed to install Anaconda, instead I am using Python's built-in `venv` virtual environment manager. This works for me, would we want to mention this to users or should we keep it to a single preferred workflow that is supported?

SylviaWhittle (approved):
Good addition for clarity for M1 SOC users, no problems.

PR #193, opened 2022-07-05
Addition of custom colormaps

benthomasuos:
Includes config yaml option for custom and matplotlib colormaps for all plots.


ns-rse:
Superseded by #210

PR #208, opened 2022-07-12
207 | Use bounding boxes for layer; exclude small grains prior to min…

ns-rse:
I think I've found what was causing the problem here and its two fold.

## Bounding Boxes

This is straight-forward, the `Grains.find_grains()` method didn't save the `region_properties` nor the `bounding_boxes` in dictionaries with keys for each layer. Changing this means tweaking the `Grains.get_region_properties()` and `Grains.get_bounding_boxes()` to return properties/bounding boxes respectively and these are saved to dictionaries in `Grains.region_properties[direction]` and `Grains.bounding_boxes[direction]` respectively.

## Failing to find grains

This is more complicated.

In the `Grains()` class (see `topostats/grains.py` the method `calc_minimum_grain_size()` is defined as such...

```python
    def calc_minimum_grain_size(self, image: np.ndarray, direction: str) -> float:
        """Calculate the minimum grain size.

        Very small objects are first removed via thresholding before calculating the lower extreme.
        """
        # self.get_region_properties(image)
        # grain_areas = np.array([grain.area for grain in self.region_properties[direction]])
        region_properties = self.get_region_properties(image)
        grain_areas = np.array([grain.area for grain in region_properties])
        if len(grain_areas > 0):
            # Exclude small objects less than a given threshold first
            grain_areas = grain_areas[
                grain_areas >= threshold(grain_areas, method=threshold_method, otsu_threshold_multiplier=1.0)
            ]
            self.minimum_grain_size = np.median(grain_areas) - (
                1.5 * (np.quantile(grain_areas, 0.75) - np.quantile(grain_areas, 0.25))
            )
        else:
            self.minimum_grain_size = -1
```
I reinstated the line...

```
grain_areas = grain_areas[grain_areas >= threshold(grain_areas, method=threshold_method, otsu_threshold_multiplier=1.0)]
```

...yesterday because, despite removing noise with the `Grains.remove_noise()` which excludes grains that are < the value `absolute_smallest_grain_size` (from the configuration and carried through with the same name) the tests failed as there were now more grains in the subsequent steps where the `median - 1.5 * IQR` (IQR = Inter-Quartile Range) was calculated and the value of `Grains.minimum_grain_size` was in turn smaller.

However...

In doing so a call is made to the `threshold()` function directly from `topostats.threshold` rather than using the `get_threshold()` method from `topostats.utils` which is the behaviour we want because we are processing a specific `lower` or `upper` threshold and wish to find the minimum grain size to exclude grains from the list `grain_areas`.  But thresholding methods in scikit-image, which does the work in `topostats.threshold`, do not include a `std_dev` method, rather when TopoStats wants to use the `mean -/+ std_dev` it uses `get_thresholds()` and gets the `mean` threshold (from `topostats.threshold`) for the image using scikit-image method and then calculates the standard deviation for the image, scales it by the desired `std_dev` value and -/+ it from the mean. 

Thus, the problem is that `threshold()` was being called with `method=self.threshold_method` which when using `std_dev` it could not handle and instead raised a `ValueError`. This was captured broadly and returned a `[<image-name>] : No grains found.` message in the logger.

This is again the situation where there are lots of small grains and some large (despite noise removal) and the small need removing _before_ the `minimum_grain_size` is calculated. To which end the Otsu method is in my view most appropriate, and because this aligns with the previous steps taken the tests pass. Another option might be to adjust the `absolute_minimum_grain_size` to be larger so that the `Grains.remove_noise()` method removes more of the smaller grains.

I'm still unsure about then calculating the `minimum_grain_size` as being the `median - 1.5 * IQR`, it would probably be okay if what is left is grains that roughly follow a Gaussian distribution but if that is the case why not use the `mean - a * std_dev`?

### DIFF
--- a/topostats/grains.py
+++ b/topostats/grains.py
@@ -90,8 +90,8 @@ class Grains:
         }
         self.directions = defaultdict()
         self.minimum_grain_size = None
-        self.region_properties = None
-        self.bounding_boxes = None
+        self.region_properties = defaultdict()
+        self.bounding_boxes = defaultdict()
         self.grainstats = None
         Path.mkdir(self.base_output_dir, parents=True, exist_ok=True)
 
@@ -147,12 +147,12 @@ class Grains:
 
         Very small objects are first removed via thresholding before calculating the lower extreme.
         """
-        self.get_region_properties(image)
-        grain_areas = np.array([grain.area for grain in self.region_properties])
+        region_properties = self.get_region_properties(image)
+        grain_areas = np.array([grain.area for grain in region_properties])
         if len(grain_areas > 0):
             # Exclude small objects less than a given threshold first
             grain_areas = grain_areas[
-                grain_areas >= threshold(grain_areas, method=self.threshold_method, otsu_threshold_multiplier=1.0)
+                grain_areas >= threshold(grain_areas, method="otsu", otsu_threshold_multiplier=1.0)
             ]
             self.minimum_grain_size = np.median(grain_areas) - (
                 1.5 * (np.quantile(grain_areas, 0.75) - np.quantile(grain_areas, 0.25))
@@ -211,7 +211,8 @@ class Grains:
         LOGGER.info(f"[{self.filename}] : Coloured regions")
         return coloured_regions
 
-    def get_region_properties(self, image: np.array, **kwargs) -> List:
+    @staticmethod
+    def get_region_properties(image: np.array, **kwargs) -> List:
         """Extract the properties of each region.
 
         Parameters
@@ -224,24 +225,26 @@ class Grains:
         List
             List of region property objects.
         """
-        self.region_properties = regionprops(image, **kwargs)
-        LOGGER.info(f"[{self.filename}] : Region properties calculated")
+        return regionprops(image, **kwargs)
 
-    def get_bounding_boxes(self) -> Dict:
+    def get_bounding_boxes(self, direction) -> Dict:
         """Derive a list of bounding boxes for each region from the derived region_properties
+
+        Parameters
+        ----------
+        direction: str
+            Direction of threshold for which bounding boxes are being calculated.
 
         Returns
         -------
         dict
             Dictionary of bounding boxes indexed by region area.
         """
-        LOGGER.info(f"[{self.filename}] : Extracting bounding boxes")
-        self.bounding_boxes = {region.area: region.area_bbox for region in self.region_properties}
+        return {region.area: region.area_bbox for region in self.region_properties[direction]}
 
     def find_grains(self):
         """Find grains."""
         LOGGER.info(f"[{self.filename}] : Thresholding method (grains) : {self.threshold_method}")
-        # self.threshold = self.get_threshold(self.image, self.threshold_method)
         self.thresholds = get_thresholds(
             image=self.image,
             threshold_method=self.threshold_method,
@@ -251,8 +254,7 @@ class Grains:
         )
         try:
             for direction, _threshold in self.thresholds.items():
-
-                # Create sub-directory for the upper / lower grains
+                LOGGER.info(f"[{self.filename}] : Processing {direction} threshold ({_threshold})")
                 self.directions[direction] = defaultdict()
                 self.gaussian_filter()
                 self.directions[direction]["mask_grains"] = _get_mask(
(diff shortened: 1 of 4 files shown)


topostats/grains.py
```diff
@@ -277,12 +279,16 @@ def find_grains(self):
                 self.directions[direction]["labelled_regions_02"] = self.label_regions(
                     self.directions[direction]["removed_small_objects"]
                 )
-                self.get_region_properties(self.directions[direction]["labelled_regions_02"])
-
+                self.region_properties[direction] = self.get_region_properties(
+                    self.directions[direction]["labelled_regions_02"]
+                )
+                LOGGER.info(f"[{self.filename}] : Region properties calculated ({direction})")
+                print(f"#### {direction} region_properties : {self.region_properties[direction]}")
```
MaxGamill-Sheffield, line 286: Line 286 in `grains.py` still has the print statement meaning the following still appears:
```
[Wed, 13 Jul 2022 11:39:17] [INFO    ] [topostats] [minicircle] : Labelling Regions
[Wed, 13 Jul 2022 11:39:17] [INFO    ] [topostats] [minicircle] : Region properties calculated (upper)
#### upper region_properties : [<skimage.measure._regionprops.RegionProperties object at 0x156891f10>, <skimage.measure._regionprops.RegionProperties object at 0x156891b20>, <skimage.measure._regionprops.RegionProperties object at 0x156891ee0>, <skimage.measure._regionprops.RegionProperties object at 0x156891c40>, <skimage.measure._regionprops.RegionProperties object at 0x156891d60>, <skimage.measure._regionprops.RegionProperties object at 0x156891d30>, <skimage.measure._regionprops.RegionProperties object at 0x156891cd0>, <skimage.measure._regionprops.RegionProperties object at 0x156891ca0>, <skimage.measure._regionprops.RegionProperties object at 0x156891c70>, <skimage.measure._regionprops.RegionProperties object at 0x156891be0>, <skimage.measure._regionprops.RegionProperties object at 0x156891bb0>, <skimage.measure._regionprops.RegionProperties object at 0x156891f40>, <skimage.measure._regionprops.RegionProperties object at 0x156891fa0>, <skimage.measure._regionprops.RegionProperties object at 0x156891b80>, <skimage.measure._regionprops.RegionProperties object at 0x156891b50>, <skimage.measure._regionprops.RegionProperties object at 0x156891af0>, <skimage.measure._regionprops.RegionProperties object at 0x156891ac0>, <skimage.measure._regionprops.RegionProperties object at 0x156891a90>, <skimage.measure._regionprops.RegionProperties object at 0x156891a30>, <skimage.measure._regionprops.RegionProperties object at 0x156891a00>, <skimage.measure._regionprops.RegionProperties object at 0x1568919d0>]
[Wed, 13 Jul 2022 11:39:17] [INFO    ] [topostats] [minicircle] : Coloured regions
[Wed, 13 Jul 2022 11:39:17] [INFO    ] [topostats] [minicircle] : Extracted bounding boxes (upper)
[Wed, 13 Jul 2022 11:39:17] [INFO    ] [topostats] [minicircle] : *** Grain Statistics ***
[Wed, 13 Jul 2022 11:39:17] [INFO    ] [topostats] [minicircle] : Processing grain: 0
```
Other than that it seems good to go!

topostats/run_topostats.py
```diff
@@ -255,7 +255,7 @@ def process_scan(
     if grains.region_properties is not None:
         # Grain Statistics :
         try:
-            LOGGER.info(f"[{filtered_image.filename}] : Grain Statistics")
+            LOGGER.info(f"[{filtered_image.filename}] : *** Grain Statistics ***")
```
MaxGamill-Sheffield, line 258: The "***...***" is not consistent across processing, grain finding and grain statistics. 

Is this by design?

MaxGamill-Sheffield (changes requested):
Works for my testing and also sorts the issue where the 'colour regions' key isn't found and causes crashes. 

Just a print statement to remove and add consistency across logging for each of the sections.

SylviaWhittle:
I agree with the use of Otsu for size based thresholding.

SylviaWhittle (approved):
I have tested this against a large-ish dataset and found no problems. These are really great changes, thank you @ns-rse.

PR #209, opened 2022-07-13
Sylvia whittle/204 nanometre scaling

SylviaWhittle:
This changes the scaling on image plots from pixels to nanometres. Also adds the units [nanometres] as labels on the x and y axes of images.

Changes internally are that now `pixel_to_nm_scaling` is now passed to `plot_and_save` in all cases. Tests have been updated accordingly.


ns-rse (commented):
Excellent work incorporating all this @SylviaWhittle :+1:

PR #210, opened 2022-07-13
Adds colormap, updates test images, simplifies passing plotting opts

ns-rse:
I got confused trying to resolve all the conflicts between changes to `dev` that had been merged since #193 was drafted.

I therefore pulled across the `topostats/themes.py` from @benthomasuos branch and updated tests and modified the plotting to work with the colourmap. The tests now make plots which...

* Include colorbars
* Use Nanometers on the axes
* Use `nanoscope` colormap except a new test which explicitly uses `afmhot` to check that the option in the `plot_and_save()` function works correctly.

If you run this now the bounding boxes are not correct but that should be resolved by #208 which I'm just going to tidy up now after feedback from @MaxGamill-Sheffield.


SylviaWhittle (approved):
Works for me, no problems found. 

The plots are looking good!

MaxGamill-Sheffield:
Works for me too! But I'd suggest adding a label to the Z scale "<channel name> nm" or something. I was also going to share a comparison to the experimentalists of the old NS and new NS for comparison with the following image
[nanoscope colour changes.pdf](https://github.com/AFM-SPM/TopoStats/files/9110990/nanoscope.colour.changes.pdf)

SylviaWhittle:
Just my opinion, but the old image's thinner colour bar looks cleaner. It's probably an unimportant aesthetic change - also might be worse from an accessibility standpoint?

ns-rse:
@MaxGamill-Sheffield & @SylviaWhittle  This is where Git, branches/issues come in and the beauty of these is that they are small and specific, so make an issue for this, branch, make the changes, put a Pull Request out for review and then merge.

The width of the bar is likely controlled by some aspect of ...

```python
66            cax = divider.append_axes("right", size="5%", pad=0.05)
```

Rather than making it fixed add it as an option to the function and include it to the `PLOT_DICT` automatically from the `example.yaml`.

I'd have to look up how to add `nm` to the colorbar scale.

PR #215, opened 2022-07-14
Scale Bounding boxes and add test & fix DNA tracing

### LINKED ISSUE #212: Bounding boxes need scaling to nanometers
**Describe the bug**
Only three bounding boxes appear on the plots now they are scaled to nanometers, this is because the bounding box locations are based on pixels and need scaling too.

**To Reproduce**

```bash
```

**Expected behavior**

Should be bounding boxes around each image.

**Output**
image

ns-rse:
Resolves #212 

Whilst the x and y-axis were scaled with PR #209 the location of bounding boxes were still in pixels and as a result only a portion of the boxes were drawn and they were way off scale.

This has been fixed and a test added.

Weirdly though I have had to correct the y-axis co-ordinates. Without the correction the image plots the boxes inverted on the y-axis as per...

image

The correction puts them back in place

image

At the same time I noticed that the DNA tracing was failing and have fixed that. The wrong itteration was being used to pass over a list (`.keys()` when it should have been `items()` and corrected referring to the appropriate item).


SylviaWhittle (approved):
Works for me. I tried making the tests fail by editing how the bounding box coordinate calculations work and the test successfully failed. ✅

PR #217, opened 2022-07-14
Adding topostats/logs/__init__.py

### LINKED ISSUE #216: Fix installation
## Describe the bug
An install with `pip install .` does not work modules are not found (specifically the `topostats.logs.logs` module, however `pip install -e .` does work. This will need fixing before releasing to PyPI.

## To Reproduce

- Clone the repository
- Install with `pip install . && pup install '.[tests]'`
- Run `pytest` and it fails.

## Expected behavior

Tests should run.

### Adding `__init__.py`

Dropping an empty `topostats/logs/__init__.py` should work but didn't, despite this being the case for `topostats/tracing/__init__.py`.

### Removing extra `__init__.py`

Similarly removing `topostats/tracing/__init__.py` and `topostats/logs/__init__.py` didn't work either.

## Proposed solution

Move `topostats/logs/logs.py` > `topostats/logs.py` and update all imports.

## Relevant reading

* [Namespace package use case](https://stackoverflow.com/a/48804718)
* [Traps for the Unwary in Python’s Import System](https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html)

ns-rse:
Resolves #216

This should hopefully fix installation. Whilst Python 3.3 did introduce the ability not to need `__init__.py` in every directory this is useful when using namespace packages. For regular packages its is better to include the `__init__.py` in module directories as @bobturneruk correctly advised when the issue was first discussed.

If testing create a new virtual environment and 
```
pip install . && pip install .[tests]
pytest
```

If it fails check carefully what virtual environments are activated. I found when investigating this I had activated a standard virtual environment from a Conda virtual environment and some tools were found in the later.

See #216 for links to relevant reading.


MaxGamill-Sheffield (approved):
This makes `pip install . && .[tests]` now work when running Topostats by solving the "unable to find modules ..." error. There may be merge conflict issues with PR #214 as the resultant images have no Z-scale bar labels

PR #218, opened 2022-07-14
211 | Test of bounding box with type binary excluding colorbar

ns-rse:
Had a strange merge conflict between the test added for bounding boxes on @MaxGamill-Sheffield recent PR so I removed it. Reinstating here with his updated functionality to exclude the colorbar from binary plots.


MaxGamill-Sheffield (approved):
Tests now pass and the image in the temp dir under `test_plot_and_save_bounding_bo0` is this one showing it works

PR #220, opened 2022-07-18
Dictionary update

Jean-Du:
Updating the dictionary and unit conversion to accommodate the output of new TopoStats. Fixing the lack of brackets in some printing statements.

### DIFF
--- a/Plotting.py
+++ b/Plotting.py
@@ -44,6 +44,21 @@ colname2label = {
     'grain_laplace_volume': 'Laplacian Volume / $\mathregular{%s^3}$',
     'End to End Distance': 'End to End Distance / nm',
     'Contour Lengths': 'Contour Lengths / nm',
+    'raidus_min': 'Minimum Radius / nm',
+    'radius_max': 'Maximum Radius / nm',
+    'radius_mean': 'Mean Radius / nm',
+    'radius_median': 'Median Radius / nm',
+    'height_min': 'Minimum Height / nm',
+    'height_max': 'Maximum Height / nm',
+    'height_median': 'Median Height / nm',
+    'height_mean': 'Mean Height / nm',
+    'volume': 'Volume / $\mathregular{nm^3}$',
+    'area': 'Area / $\mathregular{nm^2}$',
+    'area_cartesian_bbox': 'Cartesian Bounding Box Area / $\mathregular{nm^2}$',
+    'smallest_bounding_width': 'Smallest Bounding Width / nm',
+    'smallest_bounding_length': 'Smallest Bounding Length / nm',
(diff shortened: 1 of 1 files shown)


Plotting.py
```diff
@@ -456,7 +475,7 @@ def computeStats(data, columns, min, max):
 
 
 if __name__ == '__main__':
-    # Path to the json file, e.g. C:\\Users\\username\\Documents\\Data\\Data.json
+    # Path to the json file, e.g. C:\\Users\\username\\Documents\\Data\\Data.csv
```
MaxGamill-Sheffield, line 478: Should be "Path to the csv file".

Plotting.py
```diff
@@ -456,7 +475,7 @@ def computeStats(data, columns, min, max):
...
 
     path = ''
```
MaxGamill-Sheffield, line 480: Perhaps add an example path like "./output/all_statistics.csv"?

MaxGamill-Sheffield (approved):
Works for me and produces both the area KDE plot and contour length histogram when pointing to `all_statistics.csv`.

Although this entire module needs to be refactored and become pip installable with the rest of TopoStats, this is a very good current solution to generate plots from the newer output files.

A few very minor changes changes but this will all change when refactoring takes place.

SylviaWhittle:
Hi @Jean-Du, I have some changes/suggestions that I would like to request - small edits like bringing in the use of `pathlib` for example. Could you hold off on merging for a little bit?

SylviaWhittle:
@Jean-Du I have made some edits in `SylviaWhittle/plotting-edits` that might be of interest to you. I could make a PR against this branch, or I could make a PR against `dev` after this is merged. What do you think the best course would be @ns-rse? I'm leaning towards the latter but would like to know your thoughts on best practice.

Plotting.py
Jean-Du: @SylviaWhittle has noticed issues with using relative paths and is working on adding pathlib. It might be better to have her add that example path in her new pull request?

Jean-Du:
> Very good, works for me and I concur with the comments that have been made previously.
> 
> I have some edits in https://github.com/AFM-SPM/TopoStats/tree/SylviaWhittle/plotting-edits that might be useful, however I think I shall make a separate PR for those against `dev`

@SylviaWhittle Thank you very much for doing this! Since you'll be making a separate PR, would it be ok for me to merge this one for now?

ns-rse:
Sorry for getting back so late to this, generally the best approach is to merge into `dev` and resolve any conflicts. As you say @SylviaWhittle the whole module needs refactoring and moving under `topostats` so that it becomes part of the installation process.

PR #221, opened 2022-07-18
Maxgamill sheffield/198 max and min z range imgs

MaxGamill-Sheffield:
There should be a new field called `zrange` in the config file which takes a lower bound and upper bound. These values are used to make a new image which confines z-axis range of the gaussian filtered image.

To test, run TS with different values of `zrange` and see if the out image `08_5-z_thresholded.png` (I tried to make it sound like 8.5), looks as expected.


build/lib/topostats/grains.py
```diff
@@ -0,0 +1,307 @@
+"""Find grains in an image."""
```
ns-rse, line 1: Not sure how this file got added but it can safely be removed.

The `build/` directory is for making package releases and we don't need to include it in the repository (I'll add it into `.gitignore` when I remove the `.png`).

build/lib/topostats/run_topostats.py
```diff
@@ -0,0 +1,459 @@
+"""Topotracing"""
```
ns-rse, line 1: Ditto, this file can be removed.

```
git rm --cache build/lib/topostats/grains.py
git rm --cache build/lib/topostats/run_topostats.py
```

tests/conftest.py
```diff
@@ -411,6 +411,14 @@ def minicircle_grain_gaussian_filter(minicircle_grain_threshold_otsu: np.array)
     return minicircle_grain_threshold_otsu
 
 
+@pytest.fixture
+def minicircle_grain_z_threshold(minicircle_grain_gaussian_filter: np.array) -> Grains:
+    """Apply Z-scale thresholding."""
+    minicircle_grain_gaussian_filter.zrange = [0, 3]
```
ns-rse, line 417: If you add the following after line 370 when you instantiate the `Grains()` class you shouldn't need this here.

```python
        zrange = grain_config["zrange"]     
```

tests/test_grains_minicircle.py
```diff
@@ -45,6 +45,22 @@ def test_gaussian_filter_minicircle(minicircle_grain_gaussian_filter: Grains, pl
     return fig
 
 
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_z_threshold_minicircle(minicircle_grain_z_threshold: Grains, plotting_config: dict, tmpdir) -> None:
+    """Test Z threshold filter."""
+    assert isinstance(minicircle_grain_z_threshold.images["z_threshed"], np.ndarray)
+    assert minicircle_grain_z_threshold.images["z_threshed"].shape == (1024, 1024)
+    assert minicircle_grain_z_threshold.images["z_threshed"].sum() == 255794.52478913686
```
ns-rse, line 53: As discussed in chat you could...

```
def test_z_threshold_minicircle(minicircle_grain_z_threshold: Grains, plotting_config: dict, grain_config: dict, tmpdir) -> None:
    ...
    assert minicircle_grain_z_threshold.images["z_threshed"].min() == grain_config["zrange"][0]
    assert minicircle_grain_z_threshold.images["z_threshed"].max() == grain_config["zrange"][1]
```

topostats/grains.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         absolute_smallest_grain_size: float = None,
         background: float = 0.0,
         base_output_dir: Union[str, Path] = None,
+        zrange: list = None
```
ns-rse, line 45: Could you please add a doc-string under `Parameters` for what this option is. That way users know and its included in the auto-generated Sphinx documentation which is built from the docstrings automatically.

topostats/grains.py
```diff
@@ -106,6 +109,16 @@ def gaussian_filter(self, **kwargs) -> np.array:
             mode=self.gaussian_mode,
             **kwargs,
         )
+    
+    def z_thresholding(self, **kwargs) -> np.array:
+        """Apply height thresholding so outliers are capped at config values"""
```
ns-rse, line 114: Pedantic but "capped" usually means or is interpreted as an upper limit, but outliers are also removed at the lower limit. Perhaps "clipped" would be more appropriate.

topostats/run_topostats.py
```diff
@@ -180,9 +182,11 @@ def process_scan(
     gaussian_mode : str
         Mode for filtering (default is 'nearest').
     otsu_threshold_multiplier : Union[int, float]
-
         Factor by which lower threshold is to be scaled prior to masking.
     background : float
+        The value to average the background around.
```
ns-rse, line 187: Thanks for adding the description here as well as the docstring for `zrange` :+1:

ns-rse (changes requested):
Great work @MaxGamill-Sheffield a few minor changes but once those are tweaked it'll be ready to go :+1:

build/lib/topostats/grains.py
MaxGamill-Sheffield: Ah I see whoops. I'm not sure how that happened.

build/lib/topostats/run_topostats.py
MaxGamill-Sheffield: Done in a recent commit

tests/conftest.py
MaxGamill-Sheffield: Ah perfect! I never saw that Grain() part!

topostats/grains.py
MaxGamill-Sheffield: Yep - we aim to make it as easy to read as possible!

topostats/grains.py
MaxGamill-Sheffield: yep, done!

topostats/run_topostats.py
MaxGamill-Sheffield: No problem! I forgot that when we moved it in the config I'd have to add it to grains too.

build/lib/topostats/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Topostats"""
```
ns-rse: Pulled in a bunch of `build/lib/*` files again.

Feel free to add the following to `.gitignore` once these are removed, then you and no one else will pull them in again.

```
build/*
```

tests/test_grains_minicircle.py
```diff
@@ -45,6 +45,24 @@ def test_gaussian_filter_minicircle(minicircle_grain_gaussian_filter: Grains, pl
     return fig
 
 
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_z_threshold_minicircle(minicircle_grain_z_threshold: Grains, plotting_config: dict, grain_config: dict, tmpdir) -> None:
```
ns-rse, line 49: Minor but you can remove `grain_config: dict` here as its no longer used.

tests/test_grains_minicircle.py
```diff
@@ -45,6 +45,24 @@ def test_gaussian_filter_minicircle(minicircle_grain_gaussian_filter: Grains, pl
...
+    """Test Z threshold filter."""
+    assert isinstance(minicircle_grain_z_threshold.images["z_threshed"], np.ndarray)
+    assert minicircle_grain_z_threshold.images["z_threshed"].shape == (1024, 1024)
+    assert minicircle_grain_z_threshold.images["z_threshed"].sum() == 255794.52478913686
+    assert minicircle_grain_z_threshold.images["z_threshed"].min() >= grain_config["zrange"][0]
```
ns-rse, line 54: Ah, sensible inequality just in case nothing reaches the min/max :+1:

ns-rse (approved):
Brilliant ta, good to merge.

build/lib/topostats/__init__.py
MaxGamill-Sheffield: Cool, done!

tests/test_grains_minicircle.py
MaxGamill-Sheffield: Cool, I've removed `grain_config` and used `minicircle_grain_z_threshold.zrange[X]` instead

Issue #222, opened 2022-07-19
Colorbar "False" option doesn't work


MaxGamill-Sheffield:
**Describe the bug**
When setting `colourbar : False` in the config file, all images still have a colour bar.

**To Reproduce**
Setting `colourbar : False` in the config file, all images still have a colour bar.

**Expected behavior**
No images should have a colour bar.

**Output**
 All images still have a colour bar.

**Additional context**
The test for this option passes meaning its not being passed through the main script correctly.

I think it's because `colorbar` is not referenced within the `process_scan` function in `run_topostats.py`

SylviaWhittle:
Resolved as of #223 ✅

PR #225, opened 2022-07-19
222 | Test for colorbar =False

ns-rse:
Further to #223 this adds a test for issue #222.

Apologies for not clocking the review on #223 earlier otherwise my comment would have been to add a test.


ns-rse:
Stupidly didn't change the target branch to `dev`. :facepalm:

PR #226, opened 2022-07-19
222 | Adding test for colorbar = False

ns-rse:
Further to #223 this adds a test for issue #222.

Apologies for not clocking the review on #223 earlier otherwise my comment would have been to add a test.


tests/test_plottingfuncs.py
```diff
@@ -24,6 +24,15 @@ def test_plot_and_save_colorbar(minicircle_pixels: Filters, plotting_config: dic
     )
     return fig
 
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_plot_and_save_colorbar_false(minicircle_pixels: Filters, plotting_config: dict, tmpdir) -> None:
```
MaxGamill-Sheffield, line 28: @ns-rse is this necessary due to a very similar, almost identical test beginning at line 10?

MaxGamill-Sheffield (commented):
I believe the test was already implemented at line 10, just the parameter not being passed from the config into that `run_topostats` function which made it work in the test (where it was updated) but not in the main running of TopoStats

ns-rse:
Already implemented such a test, just hadn't paid attention to the existing tests.

PR #229, opened 2022-07-20
Maxgamill sheffield/224 mask overlay image

MaxGamill-Sheffield:
This should now create an image combining the z-thresholded image and the upper/lower (you choose in the config) removed_small_objects grains. Additionally it includes a test that checks the produced image hasn't changed.

The config addition was added into 'Grains' because Alice mentioned she wanted this as a config option instead of producing both upper and lower grains when only one might be needed. Although this has not been implemented here, it may be useful for when it is implemented.

Here when 'upper' is specified in the config:
Screenshot 2022-07-20 at 13 44 08

And here is 'lower':
Screenshot 2022-07-20 at 13 40 30


MaxGamill-Sheffield:
Closed and re-opened because I initially tried to pull to master and after swapping to dev, the tests didn't initiate. 
Now they have done.

topostats/run_topostats.py
```diff
@@ -54,6 +54,7 @@
     "removed_noise": {"filename": "11-noise_removed.png", "title": "Noise removed", "type": "binary"},
     "labelled_regions_01": {"filename": "12-labelled_regions.png", "title": "Labelled Regions", "type": "binary"},
     "removed_small_objects": {"filename": "13-small_objects_removed.png", "title": "Small Objects Removed", "type": "binary"},
+    "mask_overlay": {"filename": "13_5-mask_overlay.png", "title": "Hight Thresholded with Mask", "type": "non-binary"},
```
ns-rse, line 57: Typo
```suggestion
    "mask_overlay": {"filename": "13_5-mask_overlay.png", "title": "Height Thresholded with Mask", "type": "non-binary"},
```

ns-rse (changes requested):
Just a typo, but otherwise good to go, thanks again for getting this done so quickly.

ns-rse (approved):
All good to go :+1:

PR #231, opened 2022-07-21
Maxgamill sheffield/224 output file dir structure

MaxGamill-Sheffield:
This PR orders the output directories into a structure similar to how the files are found.

Here, using "./data/" as the `base_dir` in the config file, we have multiple directories containing multiple files, i.e:
Screenshot 2022-07-21 at 12 41 14

The PR replaces the `base_dir` of the `image_path` with the `output_dir`, keeping the subsequent directory structure after.
This creates an output mirroring the inputs, i.e:
Screenshot 2022-07-21 at 12 43 36

Further work will be done to remove the `minicircle<#>` directories but this should only happen after:
1) The output files are truncated.
2) The output files are re-named to reflect the file they originated from
In order to avoid overwriting output files.

### DIFF
--- a/topostats/utils.py
+++ b/topostats/utils.py
@@ -84,7 +84,7 @@ def get_out_path(image_path: Union[str, Path] = None, base_dir: Union[str, Path]
     image_path: Union[str, Path]
         The path of the current image.
     base_dir: Union[str, Path]
-        The directory which recursively searched for files.
+        Directory to recursively search for files, if not specified the current directory is scanned.
     output_dir: Union[str, Path]
         The output directory specified in the configuration file.
 
--- a/topostats/utils.py
+++ b/topostats/utils.py
@@ -95,12 +95,7 @@ def get_out_path(image_path: Union[str, Path] = None, base_dir: Union[str, Path]
     """
     pathparts = list(image_path.parts)
     inparts = list(base_dir.parts)
-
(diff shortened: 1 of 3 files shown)


MaxGamill-Sheffield:
Apologies for the post PR commit. I realised I could obtain the output path faster by not using for loops

bobturneruk:
Fine by me! Might as well make things better where there's an opportunity, and it hasn't been reviewed!

topostats/utils.py
```diff
@@ -76,6 +76,27 @@ def find_images(base_dir: Union[str, Path] = None, file_ext: str = ".spm") -> Li
     return list(base_dir.glob("**/*" + file_ext))
 
 
+def get_out_path(image_path: Union[str, Path] = None, base_dir: Union[str, Path] = None, output_dir: Union[str, Path] = None):
```
ns-rse, line 79: You can add a return type here...

```suggestion
def get_out_path(image_path: Union[str, Path] = None, base_dir: Union[str, Path] = None, output_dir: Union[str, Path] = None) -> Path:
```

ns-rse (changes requested):
Thanks for getting this done so quickly Max, one very minor addition of a return type hint.

ns-rse (approved):
Also good to go :+1:

PR #235, opened 2022-07-27
Output data dictionary

bobturneruk:
This will define topostats output columns.


ns-rse (approved):
I think this should be merged, don't think anything has changed and perhaps others with more knowledge on the descriptors could fill in the blanks.

Units are likely to change in the not too distant future when moving to #243 is addressed.

ns-rse:
Closing as I've incorporated into #307

Issue #224, opened 2022-07-19
Reorder output file structure


MaxGamill-Sheffield:
Just spoke to Alice, Eddie and Libby and there are concerns about the output files. Mainly to increase the speed of finding what you need and the labelling of images. See below:

When dealing with multiple files, and multiple folders of files, the outputs should:
- [x]  1. Carry the name of the parent folder with them.

2. Produce a '_processed' image and '_processed_masked' image for every .spm, and a grainstats.csv of all .spm's within that folder.
    - [x] Produce a mask overlay image
    - [x] Produce these in the correct directory (no `_output_dir/<filename>/...`,)
    - [x] Include filename in the image filename

- [x]  3. File names should be contained within all the produced images.

4. Order grain outputs
    - [x] Folder of all grains from all files in the directory
    - [x] Rename grain files to be <fname>_grain_#
    - [x] Make grains optional

4. Order grainstats outputs
    - [x] Global grainstats file at location of `base_dir`
    - [x] Include per directory grainstats
    - [x] Remove per file grainstats

An example structure is shown below:
```
Input:
-- <data> --- folder1 ---- one.spm
		      ---- two.spm
		      ---- three.spm
	
	  --- folder2 ---- four.spm
		      ---- five.spm


Output:
-- <output> --- folder1_processed ---- one_processed.png (image 08_5)
		      		  ---- one_processed_masked.png (image 08_5 + 13)
				  ---- two_processed.png 
		      		  ---- two_processed_masked.png 
				  ---- three_processed.png 
		      		  ---- three_processed_masked.png
				  ---- folder1_grainstats.csv (only containing one, two and three)
				  ---- cropped...
	
	    --- folder2_processed ---- four_processed.png
				  ---- four_processed_masked.png
			   	  ---- five_processed.png
			   	  ---- five_processed_masked.png
				  ---- folder2_grainstats.csv
				  ---- cropped...
            all_grainstats.csv

```

ns-rse:
One thing I found particularly confusing when first getting involved was the order in which plots should be considered.

I appreciate that many are redundant and not required (based on the above example) but it is why I introduced the numerical prefixes to output images.

Also `grainstats.csv` will always be produced but there is also output from `dnatracing.py` which is combined. I set it up so that they were combined, should these be included in the output of the file `grainstats.csv`?

[...]

PR #239, opened 2022-07-28
Maxgamill sheffield/224 truncate output and img names

MaxGamill-Sheffield:
This should address all the issues regarding file output structure from #224 and have a few tests implemented.

Additionally, saving the grains is now optional, and "core" and "all" image sets have allowed the reduction of the number of output files to only those essential to experimentalists.

Some lines that are no longer used have been commented out, such as those creating the `grain{index}` directories. However, if people believe they should be outright removed, please point them out and let me know.

Some tests are a bit lackluster 
- The T/F save_cropped_grains test only checks that a "grains" directory is populated and not that grain images exist. 
- The image_set config option test only checks that the additional grain images are output/not output, and doesn't check for the presence of additional filtered images. However, this is because the plots for these are in the main body of "run_topostats" and I do not know where tests for these go. I suggest the Filter image generation be moved into the Filters module and class.


ns-rse (approved):
I have one test fail for me when I run this locally, but the tests pass here on GitHub which is strange.

A few linting errors have crept in (mostly trailing whitespace/newlines), but some of the code still needs linting (e.g. `topostats/tracing/dnatracing.py` in particular).

Happy to approve and work on these in due course.

MaxGamill-Sheffield:
I think you mentioned that some refactoring still needs to happen with DNA tracing so I guess the linting could be done then.

Other than that, how do you recursively lint each file? can you just run `pylint */*.py`?

ns-rse:
Yes, that was my plan to ensure `dnatracing` is linted when I refactor it. :+1:

PR #244, opened 2022-08-02
Maxgamill sheffield/238 enforce zrange and legend outside img

MaxGamill-Sheffield:
This pull request should fix a few things. It should now
1. Have the legend for the `mask_overlay` image outside the image itself.
2. Should make it so the raw grain images are also thresholded according to the `zrange` (now found in the plotting section of the config file)
3. Enforce the `zrange` so even if height data lies within the range, the colour scale is the same as it would be for that range.

Tests have also been implemented.


MaxGamill-Sheffield:
Not to self: 
Currently the part of the testing suite that fails is the `.github/workflows/test.yaml` and the failing test is `pip install -e .` followed by `run_topostats --config config/example.yaml`.

This works on my machine however, it fails with the error: `got multiple values for keyword argument 'zrange'`.

Want to check zrange is not in the `PLOT_DICT` dictionary that is being passed into the `plot_and_save` function, as it should be passed via the inclusion of the `zrange` parameter directly, in `zrange=zrange`.

output/config.yaml
```diff
@@ -0,0 +1,43 @@
+# Configuration from TopoStats run completed : 2022-08-09 11:09:18
```
ns-rse, line 1: Don't think this file should be included as its part of the output.

tests/test_grains_minicircle.py
```diff
@@ -45,24 +45,6 @@ def test_gaussian_filter_minicircle(minicircle_grain_gaussian_filter: Grains, pl
     return fig
 
 
-@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
```
ns-rse, line 48: If this test is being removed we should also remove the `tests/img/test_z_threshold_minicircle.png` that is associated with it.

```
git rm tests/img/test_z_threshold_minicircle.png
```

tests/test_grains_minicircle.py
```diff
@@ -135,7 +117,7 @@ def test_remove_small_objects(minicircle_small_objects_removed: Grains, plotting
     )
     return fig
 
-
+'''
```
ns-rse, line 120: Is there a reason this test has been commented out?

If it is no longer required it can be removed (along with the associated `test/img/test_mask_overlay.png`), otherwise the quotes around can be removed and the test re-instated.

EDIT : Ah, I've read through `grains.py` now and there is no `z_range` passed in and no thresholding done to produce a `z_threshed` image so this test is no longer required.

ns-rse (changes requested):
Some minor tidying up but otherwise good to go. Great work as always @MaxGamill-Sheffield :+1:

output/config.yaml
MaxGamill-Sheffield: Yep absolutely! Don't remember how it snuck in. Resolved in f092894

tests/test_grains_minicircle.py
MaxGamill-Sheffield: Yep it was changed but the image is still being used for a test elsewhere. Removed the redundant test in c328770

tests/test_grains_minicircle.py
ns-rse: I've removed this image and that associated with the `test_mask_overlay`.

ns-rse (approved):
Thanks @MaxGamill-Sheffield 

I took the liberty of removing `tests/resources/img/test_mask_overlay.png` as well and corrected a minor formatting issue in a doc string.

`tests/test_grain_minicircles.py` passes locally (as do all other tests) and so good to merge once the tests here on GitHub complete.

PR #252, opened 2022-08-04
249 | Adds regression tests using pytest-regtest

ns-rse:
Resolves #249

Rather than comparing an existing CSV which is read in to a fixture we now check output using pytest-regtest which saves output to `tests/_regtest_outputs/` and compares the test that is run against these.

PR #258, opened 2022-08-05
DNA tracing optional

### LINKED ISSUE #241: Make DNA Tracing Optional
Make DNA tracing optional as a parameter in the config file.

When `True`, DNA tracing should run and output the DNA tracing stats into the folder-wise and all_statistics.csv along with the grainstats.

When `False`, it should not run but still output the grainstats files but only for the grain stats.

N/A

**Additional context**
N/A

ns-rse:
Resolves #241 

Makes DNA tracing optional based on the configuration file option...

```
dnatracing:
  run: true # Whether to run DNA Tracing. Options : true, false
```

If _not_ run then just the grainstats data frame is saved.


tests/resources/sample_config.yaml
```diff
@@ -30,6 +30,8 @@ grains:
     absolute:
       - 1
       - -1
+dnatracing:
+  run: true # Whether to run DNA Tracing. Options : true, false
 plotting:
   save: True
```
MaxGamill-Sheffield, line 36: consistency in bools here too if applicable

config/example.yaml
```diff
@@ -30,6 +30,8 @@ grains:
     absolute:
       - 1
       - -1
+dnatracing:
+  run: true # Whether to run DNA Tracing. Options : true, false
 plotting:
   save: True
```
MaxGamill-Sheffield, line 36: Does `read_yaml` read true/false as Python bool True/False? if so should we make these the same so they are consistent?

MaxGamill-Sheffield (commented):
Yep works great, good job! Just a few consistency things.

I've also noticed that the "lower" values have disappeared from `all_stats.csv`. I don't think this has happened now so should I create a new issue to fix it or fix it here?

topostats/utils.py
```diff
@@ -149,10 +152,10 @@ def _get_mask(image: np.ndarray, threshold: float, threshold_direction: str, img
     if threshold_direction == "upper":
         LOGGER.info(f"[{img_name}] : Masking (upper) Threshold: {threshold}")
         return image > threshold
-    elif threshold_direction == "lower":
+    if threshold_direction == "lower":
         LOGGER.info(f"[{img_name}] : Masking (lower) Threshold: {threshold}")
         return image < threshold
-    LOGGER.fatal(f"[{img_name}] : Threshold direction invalid: {threshold_direction}")
+    # LOGGER.fatal(f"[{img_name}] : Threshold direction invalid: {threshold_direction}")
```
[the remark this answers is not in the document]
ns-rse: I was getting a linting error that either all statements should `return` something or none at all. If neither of these conditions are met there is nothing to return so I commented out the logging statement.

config/example.yaml
ns-rse: Depends on the [version of YAML](https://stackoverflow.com/questions/42283732/are-on-and-off-supposed-to-be-interpreted-as-true-or-false-in-yaml-1-2), out of habit I've always used `true` as that seems to be what 1.2 uses. I'll update them all

ns-rse:
> Yep works great, good job! Just a few consistency things.
> 
> I've also noticed that the "lower" values have disappeared from `all_stats.csv`. I don't think this has happened now so should I create a new issue to fix it or fix it here?

No I'll have a quick look now. Typically if a PR breaks something it shouldn't be merged. In one sense this points to the fact that our tests aren't covering everything. I'll think about how to improve that but only have time to address the immediate issue as I have to start getting ready to go on holiday tomorrow.

ns-rse:
Well that took longer than expected and some strange unexpected issues (see comments on ee78fc7 ) but tests all pass, and output includes `lower` and `upper` now. Strange that the `inplace=True` option had disappeared from `create_trace_stats()`. The only thing I can think of is perhaps a branch was made from `dev` that wasn't up-to-date.

Anyway, this is likely going to mean that #252 will need to be rebased and the output the regression tests are compared to there updated. I'll do that when back from holiday.

MaxGamill-Sheffield:
Very tidy and works for me on one and multiple images. Great job cleaning up the files and making things consistent.

MaxGamill-Sheffield (approved):
Very tidy and works for me on one and multiple images. Great job cleaning up the files and making things consistent.

PR #261, opened 2022-08-10
pixel to nm scaling no longer assumes xy in nm

MaxGamill-Sheffield:
.spm data seems like it is encoded as a value and a unit. This change adds a dictionary to the `extract_pixel_to_nm_scaling` function in the `Filters` class and uses the units to rescale produced value of `self.pixel_to_nm_scaling`.

Tests have also been added via `parameterize` so there is room for non-square tests too.

Currently, this is only a quick fix but has a framework in place to address some other issues I've found associated with this:
- It is currently all in reference to "nm" where it may be best to have SI units which can then propagate to the grainstats and dnatracing stats.
- Framework is present within the function to obtain px-to-nm values for non-square images but further changes would be needed to propagate these values through to later functions as the output would differ (no longer one value).
- Always using nm in plots looks weird for large values (10000's of nm) so would ideally pull the units through to the plots.
- Pixel values (heights) have no units that I can see, a similar issue may occur like above but for the height scale of images.

What are people's thoughts on making the above changes in other branches/PR's? Or just waiting until my other PR is merged to then make changes on this branch because of the merge conflicts with changing so much function input/output?

### DIFF
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -74,7 +74,6 @@ def test_extract_pixel_to_nm_scaling(test_filters_random: Filters, unit, x, y , 
     'x': x,
     'y': y
     }
-    #test_filters_random.images["extracted_channel"].pxs()
     test_filters_random.extract_pixel_to_nm_scaling()
     assert test_filters_random.pixel_to_nm_scaling == expected
 
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -63,6 +63,22 @@ def test_extract_pixels(test_filters: Filters) -> None:
     assert test_filters.images["pixels"].shape == (1024, 1024)
 
 
+@pytest.mark.parametrize("unit, x, y, expected", [
+    ('um', 100, 100, 97.65625),
+    ('nm', 50, 50, 0.048828125),
+    ])
+def test_extract_pixel_to_nm_scaling(test_filters_random: Filters, unit, x, y , expected) -> None:
+    """Test extraction of pixels to nanometer scaling."""
+    test_filters_random.images["extracted_channel"].size['real'] = {
+    'unit': unit,
+    'x': x,
+    'y': y
+    }
+    #test_filters_random.images["extracted_channel"].pxs()
+    test_filters_random.extract_pixel_to_nm_scaling()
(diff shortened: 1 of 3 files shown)


tests/test_grainstats.py
```diff
@@ -15,7 +15,7 @@
 EDGES = np.array([POINT1, POINT2, POINT3, POINT4])
 
 
-def test_get_angle(grainstats) -> None:
+def test_get_angle(grainstats: GrainStats) -> None:
```
ns-rse, line 18: Thanks for catching this omission :+1:

tests/test_grainstats.py
```diff
@@ -24,23 +24,23 @@ def test_get_angle(grainstats) -> None:
     assert angle == target
 
 
-def test_is_clockwise_clockwise(grainstats) -> None:
+def test_is_clockwise_clockwise(grainstats: GrainStats) -> None:
```
ns-rse, line 27: Ditto for this and other missing type-hints.

tests/test_filters.py
```diff
@@ -63,6 +63,22 @@ def test_extract_pixels(test_filters: Filters) -> None:
...
+    ('nm', 50, 50, 0.048828125),
+    ])
+def test_extract_pixel_to_nm_scaling(test_filters_random: Filters, unit, x, y , expected) -> None:
+    """Test extraction of pixels to nanometer scaling."""
+    test_filters_random.images["extracted_channel"].size['real'] = {
+    'unit': unit,
+    'x': x,
+    'y': y
+    }
+    #test_filters_random.images["extracted_channel"].pxs()
```
ns-rse, line 77: I think this can probably be removed since the PySPM method (`.pxs()`) is called within the `Filters.extract_pixel_to_nm_scaling()` method.

ns-rse (changes requested):
I think there is one extraneous line in `tests/test_filters.py` but otherwise good to go.

tests/test_filters.py
MaxGamill-Sheffield: Done and test still work

ns-rse (approved):
Thanks @MaxGamill-Sheffield all good to go :+1:

PR #262, opened 2022-08-16
Fixing regression test

ns-rse:
Somehow, despite the tests passing in the GitHub Actions, @SylviaWhittle found that `dev` was broken, same for me too when I checked.

For some reason the target of regression test `tests/test_dnatracing_minicircle.py::test_dnatracing_regression` (the file `tests/_regtest_outputs/test_dnatracing_minicircle.test_dnatracing_regression.out` had the wrong index staring from 1 instead of 0. This file has been updated in this pull request so the tests now pass locally and will do so on `dev` once merged.


SylviaWhittle (approved):
Thank you for fixing this

PR #263, opened 2022-08-17
Maxgamill sheffield/237 save different output formats

MaxGamill-Sheffield:
This change includes a config file option to:
- Optionally add axes to the produced images
- Save the images as a different format (formats provided by matplotlib)

Tests have been added to ensure axes config options work well, documentation has been cleaned up a little, a loose test to check the file format (mainly extension) added, and grain images added to the plot dict.

The default behaviour when there is no axes or colormap is to produce images which are fill the plot (i.e. no whitespace). The only issue with this is that it would be preferable for future processing pipelines to produce an image and not a plot. This would require using `imsave` over `savefig` but has the caveat of not being able to plot masks. This may need to be addressed in the future as the images are solely aesthetic.


MaxGamill-Sheffield:
Finished cleaning up so should be good for review

MaxGamill-Sheffield:
Changed to draft to modify no axes, no colorbar images to imsave instead of savefig

tests/test_plottingfuncs.py
```diff
@@ -58,13 +80,14 @@ def test_plot_and_save_bounding_box(
...
 def test_plot_and_save_zrange(minicircle_grain_gaussian_filter: Grains, plotting_config: dict, tmpdir) -> None:
-    """Test plotting with colorbar"""
+    """Tests plotting of the zrange scaled image"""
     plotting_config["zrange"] = [-10, 10]
     plotting_config["core_set"] = True
     fig, _ = plot_and_save(
         minicircle_grain_gaussian_filter.images["gaussian_filtered"],
-         tmpdir, "08_5-z_threshold.png",
-         title="Raw Height", 
+         Path(tmpdir),
```
ns-rse, line 88: Perhaps pedantic but I've been going through and putting the explicit arguments into all the tests to help avoid confusion and so readers don't have to remember the positional arguments.

```suggestion
         data=minicircle_grain_gaussian_filter.images["gaussian_filtered"],
         output_dir=Path(tmpdir),
         filename="08_5-z_threshold.png",
```

Mainly because my memory isn't great and having had a break I was having to look up the options.

topostats/plottingfuncs.py
```diff
@@ -89,7 +108,17 @@ def plot_and_save(
 
         if save:
             if image_set == "all" or core_set:
-                plt.savefig(output_dir / filename)
+                if not axes and not colorbar:
+                    plt.title('')
+                    fig.frameon=False
+                    plt.savefig(
+                        (output_dir / filename).with_suffix('.'+save_format),
```
ns-rse, line 115: Probably think I'm crazy banging on about f-strings but is there any reason not to use one here?
```suggestion
                        (output_dir / filename).with_suffix(f".{save_format}"), 
```

topostats/plottingfuncs.py
```diff
@@ -89,7 +108,17 @@ def plot_and_save(
 
         if save:
             if image_set == "all" or core_set:
-                plt.savefig(output_dir / filename)
+                if not axes and not colorbar:
+                    plt.title('')
```
ns-rse, line 112: Have you linted this code with [Black](https://github.com/psf/black)? I think it would have replaced single quotes with double quotes in most instances if so.

topostats/plottingfuncs.py
```diff
@@ -29,30 +29,48 @@ def plot_and_save(
...
+        Output directory to save the file to.
     filename : Union[str, Path]
         Filename to save image as.
+    pixel_to_nm_scaling_factor : float
+        The scaling factor showing the real length of 1 pixel, in nm.
+    data2 : np.ndarray
+        Optional mask array to overlay onto an image.
     title : str
         Title for plot.
+    type : str
```
ns-rse, line 52: Thanks for adding all these missing docstrings in. :+1:

topostats/grainstats.py
```diff
@@ -179,39 +175,29 @@ def calculate_stats(self) -> Dict:
 
             if self.save_cropped_grains:
                 output_grain.mkdir(parents=True, exist_ok=True)
-
                 if self.cropped_size == -1:
                     # Plot the cropped grain image
                     plot_and_save(
```
ns-rse, line 180: I think we might be able to reduce the amount of code here with a simple loop...

```
for name, image in {'processed_grain': grain_image, 'grainmask':grain_mask, 'grain_image': masked_grain)image].items():
    plot_and_save(data=image,
                  output_dir=output_grain,
                  filename=f"{self.image_name}_{name}_{index}"
                  pixel_to_nm_scaling_factor=self.pixel_to_nanometre_scaling,
                  **self.plot_opts[name]
                  )
```

This saves repeating essentially the same code three times.

I've purposefully left out the file extension as, in light of making the format flexible, it may be sensible to add in the file extension to the `self.plot_opts` for each image type.

tests/test_plottingfuncs.py
MaxGamill-Sheffield: @ns-rse I'll keep this in mind and make the changes!

I've also seen that pathlib can use a full path variable called `tmp_path` instead of the python LocalPath `tmpdir`. This has been an issue in a few of my tests where using the LocalPath with pathlib doesn't work hence I've used `Path(tmpdir)`. Do you want me to make the switch to make all the tests neater?

topostats/plottingfuncs.py
MaxGamill-Sheffield: Apologies, just not used to using them yet

topostats/plottingfuncs.py
MaxGamill-Sheffield: My black in VS code doesn't seem to do quotes. I think I need to look at you pylint docs again.

I've done this manually for now

topostats/grainstats.py
MaxGamill-Sheffield: Amazing spot! Added this in now for both the `cropped_size=-1` and `cropped_sizes=<value>`

MaxGamill-Sheffield:
I think this is good to go now. Main changes are:
1) Restructured `plotting_funcs.py` to classes and functions to be a bit more readable so different image types can be generated depending on config params. Additionally this should speed things up as plots are only made if `save=True` and `core_set=True` or `image_set='all'`, instead of always being plotted and only saved if the above is true.
2) When no axis or colorbar are specified in the config, the *array* is saved as an image (using plt.imsave so importing this image for other pipelines etc will have the correct pixel size etc), apart from multi-plot plots such as the bbox or masked image which are saved via plt.savefig().
3) In tests, `tmpdir` changed to `tmp_path` as to avoid using a LocalPath object in favour of a Path object that plays nicely with pathlib.
4) Image tests have params specified, and tests made for the separation of the `plot_and_save` functions.

Let me know if you agree/disagree with any of the changes

MaxGamill-Sheffield:
I think this is good to go now. Main changes are:
1) Restructured `plotting_funcs.py` to classes and functions to be a bit more readable so different image types can be generated depending on config params. Additionally this should speed things up as plots are only made if `save=True` and `core_set=True` or `image_set='all'`, instead of always being plotted and only saved if the above is true.
2) When no axis or colorbar are specified in the config, the *array* is saved as an image (using plt.imsave so importing this image for other pipelines etc will have the correct pixel size etc), apart from multi-plot plots such as the bbox or masked image which are saved via plt.savefig().
3) In tests, `tmpdir` changed to `tmp_path` as to avoid using a LocalPath object in favour of a Path object that plays nicely with pathlib.
4) Image tests have params specified, and tests made for the separation of the `plot_and_save` functions.

Let me know if you agree/disagree with any of the changes :)

tests/test_plottingfuncs.py
SylviaWhittle: I would say that using the full path variable would be better, as I believe there are issues with cross-compatibility between different and more niche installation environments (eg: in personal projects, I find that using local paths breaks on my Windows 10 system but functions fine on my MacOS system)

SylviaWhittle:
If you wanted to, you could add a more in-depth check of filetype using the inbuilt python module `imghdr`. This module will check the file type by looking at the data contained in the file:

```
import imghdr
imghdr.what('test.png')
> 'png'
```

SylviaWhittle (commented):
A few minor things.

I would suggest running Black if you haven't done so already, I tried it and it presented a few very minor formatting changes.

topostats/plottingfuncs.py
```diff
@@ -14,96 +14,182 @@
...
+        self.cmap=cmap
+        self.region_properties=region_properties
+        self.zrange=zrange
+        self.colorbar=colorbar
+        self.axes=axes
+        self.save=save
+        self.save_format=save_format
+
+    def plot_and_save(self) -> None:
+        """Plot and save the images with savefig or imsave depending on config file parameters"""
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Good catch thank you

MaxGamill-Sheffield:
Merge with dev complete although I'm not too hopeful about the results

topostats/plottingfuncs.py
```diff
@@ -14,108 +14,212 @@
...
-        plt.xlabel("Nanometres")
-        plt.ylabel("Nanometres")
-        if colorbar and type == "non-binary":
-            divider = make_axes_locatable(ax)
-            cax = divider.append_axes("right", size="5%", pad=0.05)
-            plt.colorbar(im, cax=cax, label="Height (Nanometres)")
-        if region_properties:
-            fig, ax = add_bounding_boxes_to_plot(fig, ax, region_properties, pixel_to_nm_scaling_factor)
+            plt.title(self.title)
+            plt.xlabel("Nanometres")
```
ns-rse, line 159: With an eye on the future SI units the `xlabel` and `ylabel` values should probably be configurable options, whether that's when you `__init__` the class or for the `save_figure` method itself.

ns-rse (changes requested):
Great work refactoring `plot_and_save()` to a class. A few minor things with comments associated above, but thank you for all the work on this it looks great.

tests/test_plottingfuncs.py
```diff
@@ -1,53 +1,129 @@
...
         colorbar=True,
-    )
+        axes=True,
+        image_set="all",
+    ).plot_and_save()
+    return fig
+
+
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_plot_and_save_axes(minicircle_pixels: Filters, plotting_config: dict, tmp_path: Path) -> None:
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: yep

tests/test_plottingfuncs.py
```diff
@@ -1,53 +1,129 @@
...
+    ).save_figure()
+    assert Path(tmp_path / "result.png").exists()
+    assert fig is not None
+    assert ax is not None
+
+
+def test_save_array_figure(tmp_path: Path):
+    """Tests that the image array is saved"""
+    Images(
+        data=np.random.rand(10,10),
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: The idea here was that the method linked to this is test basically just a nice wrapper for `plt.imsave` so all that would be tested with a mpl compare its that its arguments work and this runs, hence why I thought the `pytest-mpl` might have been unnecessary as this would test an image that is already tested elsewhere (in `test_plot_and_save_no_axes_no_colorbar`). 

I'll add the `image_random` to make sure its consistent but if you disagree with the need for the mpl-compare, let me know and I'll add it in

tests/test_plottingfuncs.py
```diff
@@ -1,53 +1,129 @@
...
+
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_plot_and_save_axes(minicircle_pixels: Filters, plotting_config: dict, tmp_path: Path) -> None:
+    """Test plotting without axes"""
+    plotting_config["axes"] = False
+    fig, _ = Images(
+        data=minicircle_pixels.images["pixels"],
+        output_dir=tmp_path,
+        filename="01-raw_heightmap.png",
+        title="Raw Height",
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: I think it's because they're not taken from the plotting_config["plot_dict"][image_name]

topostats/grainstats.py
```diff
@@ -177,40 +173,19 @@ def calculate_stats(self) -> Dict:
...
-                        masked_grain_image,
-                        output_grain,
-                        f"{self.image_name}_grain_image_{index}.png",
-                        pixel_to_nm_scaling_factor=self.pixel_to_nanometre_scaling,
-                        type="non-binary",
-                        image_set=self.image_set,
-                        zrange=self.zrange,
-                        core_set=False,
-                    )
+                    for name, image in {
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: I agree but think this should be left for a later PR due to the size of this one. I'll add an issue now

topostats/plottingfuncs.py
MaxGamill-Sheffield: I've been thinking about this too and think it would be useful to use the dictionary made in `Filters.extract_pixel_to_nm_scaling()` to carry units with the scaling objects to then place into plotting.

But again, I think this should be left for a future PR to not mix up things. I believe it's only been pulled up here due to the restructuring.

topostats/run_topostats.py
```diff
@@ -351,6 +360,7 @@ def process_scan(
     # There are two layers to process those above the given threshold and those below, use dictionary comprehension
     # to pass over these.
     if grains.region_properties is not None:
+        grain_plot_dict = {key:value for key, value in PLOT_DICT.items() if key in ["grain_image","grain_mask", "grain_mask_image"]}
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: The filetype should be passed in via line 360

tests/test_plottingfuncs.py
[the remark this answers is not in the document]
ns-rse: No, not at all, as I say I thought the suggestion might be overkill, and even reusing the `image_random` is to some extent excessive. I just clocked it and knew that the fixture existed.

tests/test_plottingfuncs.py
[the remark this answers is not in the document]
ns-rse: Don't mind, I think I recall modifying some tests the other day to use `**plotting_config` at your suggestion (but I may be imagining that I've been bouncing around between things a lot of late!).

In general my view is if the tests produce the output we are expecting and work then that is fine.

ns-rse (approved):
All good, :crossed_fingers: that my splitting of the config file doesn't have too many conflicts! I don't think it should since I modify the loading of configurations early on in `contest.py` but there is only one way to find out!

PR #275, opened 2022-08-24
Default config and splits plotting_dictionary to its own file.

### LINKED ISSUE #274: Default configuration and split plot_dict to its own file.
Configuration is made through a YAML file and #266 changed this structure so that options aligned with class arguments and at the same time moved the hard-coded `PLOT_DICT` into a configuration.

However, most won't touch the `plot_dict` section and so it should be moved to its own file so as not to intimidate users.

At the same time having a default configuration file loaded with reasonable parameters that is used when one is not specified would be useful.

ns-rse:
Resolves #274 

Splits the `plot_dict` section of `config/example.yaml` out to its own file and includes it under package data at `topostats/plotting_dictionary.yaml`. Also includes a default configuration (`topostats/default_config.yaml`) that is loaded and used if `run_topostats` is invoked without a `-c`/`--config` argument.


ns-rse:
Damn, forgot to change the target branch.

PR #276, opened 2022-08-24
Default config and splits plotting_dictionary to its own file.

### LINKED ISSUE #274: Default configuration and split plot_dict to its own file.

ns-rse:
Resolves #274 

Splits the `plot_dict` section of `config/example.yaml` out to its own file and includes it under package data at `topostats/plotting_dictionary.yaml`. Also includes a default configuration (`topostats/default_config.yaml`) that is loaded and used if `run_topostats` is invoked without a `-c`/`--config` argument.


topostats/default_config.yaml
```diff
@@ -0,0 +1,40 @@
...
+  threshold_method: std_dev # Options : otsu, std_dev, absolute
+  otsu_threshold_multiplier: 1.0
+  threshold_std_dev: 1.0
+  threshold_absolute_lower: -1.0
+  threshold_absolute_upper: 1.0
+  direction: upper # Options: upper, lower, both (defines whether to look for grains above or below thresholds or both)
+  background: 0.0
+grainstats:
+  run: true # Options : true, false
+  zrange: [0, 3]
```
MaxGamill-Sheffield, line 30: Might want to check the config file just merged as the zrange has moved along with some other bits

MaxGamill-Sheffield (dismissed):
Seems like a `pip install .` followed by a `run_topostats` fails due to missing items in the default config file. They've been pointed out in the review.

SylviaWhittle (approved):
Hi @ns-rse. This is great stuff and it works for me, though two minor things I would like to mention:

- Would `plotting_dictionary.yaml` be better placed in the `config` directory? 

- When attempting to run this branch, it failed. I believe this was due to something that is included in the .gitignore being required, however sadly I cannot confirm, as I stupidly forgot to make a backup of my old configuration before doing a full re-install of TopoStats (which solved the issue). The error log stated that the channel "Height" could not be found in `minicircle.spm` which is weird. Though it is unlikely, users updating may experience difficulties. I don't think this will be a big issue as there won't be many users updating from before this PR, just thought I would mention it on here.

ns-rse:
Thanks @SylviaWhittle 

> Hi @ns-rse. This is great stuff and it works for me, though two minor things I would like to mention:
> 
> * Would `plotting_dictionary.yaml` be better placed in the `config` directory?

I would like it to be but my understanding of how `[options.package_data]` works is that it needs to be included [under the package source](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files)  so that when builds are made available on PyPI they will be included (I had some trouble loading from a dedicated directory under there so ended up keeping them in the root directory). Currently you only get `config/` included if you clone the Git repository, but convention is to keep other files ("data") in the package source.

> * When attempting to run this branch, it failed. I believe this was due to something that is included in the .gitignore being required, however sadly I cannot confirm, as I stupidly forgot to make a backup of my old configuration before doing a full re-install of TopoStats (which solved the issue). The error log stated that the channel "Height" could not be found in `minicircle.spm` which is weird. Though it is unlikely, users updating may experience difficulties. I don't think this will be a big issue as there won't be many users updating from before this PR, just thought I would mention it on here.

Weird that `Height` couldn't be found in `minicircle.spm` :confused: 

I introduced regression tests for `process_scan()` when filtering `minicircle.spm` for `upper`, `lower` and `both` (see `tests/test_run_topostats.py`) and they have passed so the Grainstats and DNATracing should be consistent.

PR #279, opened 2022-08-24
Processing non-square images

MaxGamill-Sheffield:
One line change and test added for non-square image processing in Filters()


MaxGamill-Sheffield:
Tests should now be sorted. Due to the change causing very very slight differences in values, some assert statements and regtests were altered. 

The regtests still produced extremely similar values but were different nonetheless and so had to be reset

tests/test_filters.py
```diff
@@ -172,3 +173,12 @@ def test_row_col_medians_with_mask(
...
     np.testing.assert_array_equal(medians["cols"], image_random_col_medians_masked)
+
+
+def test_non_square_img(test_filters_random: Filters):
+    test_filters_random.images["pixels"]=test_filters_random.images["pixels"][:,0:512]
+    test_filters_random.images["zero_averaged_background"] = test_filters_random.average_background(
+        test_filters_random.images["pixels"], test_filters_random.images["pixels"]
+    )
+    assert isinstance(test_filters_random.images["zero_averaged_background"], np.ndarray)
+    assert test_filters_random.images["zero_averaged_background"].shape==(1024, 512)
```
ns-rse, line 184: I think it's useful to test that the array we've received is the same as we are expecting and a simple way of doing that is to `.sum()` it up (we could use `np.testing.assert_array_equal()` but then have to have all of the arrays stored somewhere, read them in and compare them).

This led to the above comment about masking the whole image which prevents us from doing so.

tests/test_filters.py
```diff
@@ -172,3 +173,12 @@ def test_row_col_medians_with_mask(
     assert isinstance(medians["cols"], np.ndarray)
     np.testing.assert_array_equal(medians["rows"], image_random_row_medians_masked)
     np.testing.assert_array_equal(medians["cols"], image_random_col_medians_masked)
+
+
+def test_non_square_img(test_filters_random: Filters):
+    test_filters_random.images["pixels"]=test_filters_random.images["pixels"][:,0:512]
+    test_filters_random.images["zero_averaged_background"] = test_filters_random.average_background(
+        test_filters_random.images["pixels"], test_filters_random.images["pixels"]
```
ns-rse, line 181: Been a while since I've looked at this but why do you pass the same image in as a mask here?

The `average_background()` method passes it to `row_col_medians()` where it is conditionally used to mask values using `np.ma.masked_array()` and the `mask` argument to this is expected to be Boolean. Since most values are non-zero the whole image gets masked.

In turn this means that we can't test the values stored in `test_filters_random.images["zero_averaged_background"]` and test what they are rather than just whether its a numpy array and has the desired shape (see below comment).

I'm not sure if I've articulated this well so have tried to create a Minimal Reproducible Example based on the code in `Filters()` class to show what is happening...

```python
import numpy as np
square_array = np.array([[10,11,12,13,14],
                         [24,26,29,40,21],
                         [39,89,43,11,10],
                         [56,57,20,95,11],
                         [15,89,35,46,45]])
non_square_array = square_array[0:3,]

def row_medians(x):
    return np.nanmedian(x, axis=1)

def average_background(x, mask=False):
    if mask is False:
        return (x.T - row_medians(x)).T
    else:
        masked = np.ma.masked_array(x, mask=x)
        return (masked.T - row_medians(masked)).T

# Square and no mask
average_background(square_array, mask=False)
array([[ -2.,  -1.,   0.,   1.,   2.],
       [ -2.,   0.,   3.,  14.,  -5.],
       [  0.,  50.,   4., -28., -29.],
       [  0.,   1., -36.,  39., -45.],
       [-30.,  44., -10.,   1.,   0.]])
# This is ok, its what we expect.

# Square with mask
average_background(square_array, mask=True)
masked_array(
  data=[[--, --, --, --, --],
        [--, --, --, --, --],
        [--, --, --, --, --],
        [--, --, --, --, --],
        [--, --, --, --, --]],
  mask=[[ True,  True,  True,  True,  True],
        [ True,  True,  True,  True,  True],
        [ True,  True,  True,  True,  True],
        [ True,  True,  True,  True,  True],
        [ True,  True,  True,  True,  True]],
  fill_value=1e+20,
  dtype=float64)
# A masked array isn't what we want returned, in Filters() masked values are filled with np.nan and so it would sum() to nan

# Non-square no mask
average_background(non_square_array, mask=False)
array([[ -2.,  -1.,   0.,   1.,   2.],
       [ -2.,   0.,   3.,  14.,  -5.],
       [  0.,  50.,   4., -28., -29.]])
# Again good, this is what we want

# Non-square with mask
average_background(non_square_array, mask=True)
masked_array(
  data=[[--, --, --, --, --],
        [--, --, --, --, --],
        [--, --, --, --, --]],
  mask=[[ True,  True,  True,  True,  True],
        [ True,  True,  True,  True,  True],
        [ True,  True,  True,  True,  True]],
  fill_value=1e+20,
  dtype=float64)
# Not good, we've masked everything and again in Filters() masked values are filled with np.nan and so it would sum() to nan
```

This is a test scenario so I guess one question might be "_Why does this matter, when using we won't ever use the image itself to mask, we use the derived mask_" which is true, but as I've asked below, I think we should be testing the `.sum()` of the returned array to ensure that we always get the same answer rather than just an array with the expected size.

ns-rse:
Not surprising that the values change in the regression test, but are the newer values closer or further from the values observed with Gwyddion?

tests/test_filters.py
MaxGamill-Sheffield: So my thought process here and for below was that other tests (mainly the mpl-image-compare tests) would check if the functions work as expected and produce the same output as expected. I didn't know if duplicate tests are frowned upon/common practice.

As a result, and to prevent duplicate tests, here we are just testing that the `average_background` function throws no errors when the image is not square, and returns something of an expected type and size - where the contents of the array don't matter due to testing elsewhere.  I should also mention that only the `average_background` is tested as this was the cause of the initial error.

Please let me know if you agree/disagree and I can use `test_filters_random_with_mask` fixture and overwrite the arrays for the "mask" and "pixels" image with `[:,0:512]` and add the sums.

ns-rse: Thanks for the explanation @MaxGamill-Sheffield 

Duplicating testing is indeed a bit of a waste of time, but as I tried to convey in my other comment I do think tests should be explicit and check the returned values and not just the properties of the object.

I don't think its much of a change since the default is for `mask=None` and wouldn't slow the tests down dramatically summing an array and testing its value.

```
    test_filters_random.images["zero_averaged_background"] = test_filters_random.average_background(test_filters_random.images["pixels"])
    ...
     assert test_filters_random.images["zero_averaged_background"].sum() == XXX
```

It might even b possible to parameterise the test with the original 1024x1024 `test_filters_random.images["pixels"]` and a chunk of it (although not sure how the fixtures would play with parameterisation as I've never tried.

MaxGamill-Sheffield:
@ns-rse So in comparison with the current dev branch (using the same old grains as old topo and compared to old topo), statistic values produced are ever-so-slightly better - the largest worse difference being roughly 0.047% further away from the Gwyddion value than seen in dev,  largest better difference being roughly 0.057% closer and on average, closer by 0.0079%.

Despite the changes, I think these are negligible and that a review of the filtering process needs to be done to better flatten images as some provided by Eddie have not been flattened well.

MaxGamill-Sheffield:
> Not surprising that the values change in the regression test, but are the newer values closer or further from the values observed with Gwyddion?

@ns-rse So in comparison with the current dev branch (using the same old grains as old topo and compared to old topo), statistic values produced are ever-so-slightly better - the largest worse difference being roughly 0.047% further away from the Gwyddion value than seen in dev,  largest better difference being roughly 0.057% closer and on average, closer by 0.0079%.

Despite the changes, I think these are negligible and that a review of the filtering process needs to be done to better flatten images as some provided by Eddie have not been flattened well.

tests/test_filters.py
MaxGamill-Sheffield: I've tried the fixture with parameterisation before but despite the numpy seed being set, the arrays were different.

I've replaced the `average_background` mask input with `None` and added `assert np.sum(test_filters_random.images["zero_averaged_background"]) == 44426.48188033322` which should resolve these comments

MaxGamill-Sheffield:
Regression tests will fail due to merge of Gaussian changes and non-square changes

MaxGamill-Sheffield:
Note: Please review #287 first to merge into this branch and PR

tests/test_filters.py
MaxGamill-Sheffield: Test resolved via mask=None

tests/test_filters.py
```diff
@@ -175,6 +176,16 @@ def test_row_col_medians_with_mask(
...
 
 
+def test_non_square_img(test_filters_random: Filters):
+    test_filters_random.images["pixels"]=test_filters_random.images["pixels"][:,0:512]
+    test_filters_random.images["zero_averaged_background"] = test_filters_random.average_background(
+        image=test_filters_random.images["pixels"], mask=None
+    )
+    assert isinstance(test_filters_random.images["zero_averaged_background"], np.ndarray)
+    assert test_filters_random.images["zero_averaged_background"].shape==(1024, 512)
+    assert np.sum(test_filters_random.images["zero_averaged_background"]) == 44426.48188033322
```
ns-rse, line 186: Very minor and pretty much inconsequential, but arrays have the `.sum()` method that you can apply to them so you could here do the following which would be consistent with the other tests...

```suggestion
    assert test_filters_random.images["zero_averaged_background"].sum() == 44426.48188033322
```

But its incredibly pedantic of my to highlight this and doesn't really need changing.

tests/test_grains_minicircle.py
```diff
@@ -46,13 +46,13 @@ def test_mask_minicircle(minicircle_grain_mask: Grains, plotting_config: dict, p
     return fig
 
 
-@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+#@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
```
ns-rse, line 49: Sorry, I missed this when I looked at it previously but why are the fixtures commented out here and in other tests that generate images?

tests/test_grains_minicircle.py
```diff
@@ -63,13 +63,13 @@ def test_clear_border(minicircle_grain_clear_border: Grains, plotting_config: di
...
```
ns-rse, line 66: Ditto, why is the fixture commented out here?

tests/test_grains_minicircle.py
```diff
@@ -83,18 +83,18 @@ def test_remove_noise(minicircle_grain_remove_noise: Grains, plotting_config: di
 def test_calc_minimum_grain_size_pixels(minicircle_minimum_grain_size) -> None:
     """Test calculation of minimum grain size in pixels."""
     assert isinstance(minicircle_minimum_grain_size.minimum_grain_size, float)
-    assert minicircle_minimum_grain_size.minimum_grain_size == 1571.75
+    assert minicircle_minimum_grain_size.minimum_grain_size == 1570.25
 
 
-@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+#@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
```
ns-rse, line 89: Ditto, why is the fixture commented out here?

tests/test_grains_minicircle.py
```diff
@@ -105,15 +105,15 @@ def test_remove_small_objects(
     return fig
 
 
-@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+#@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
```
ns-rse, line 108: Ditto, why is the fixture commented out here?

tests/test_grains_minicircle.py
```diff
@@ -132,13 +132,13 @@ def test_region_properties(minicircle_grain_region_properties_post_removal: np.a
         assert isinstance(x, RegionProperties)
 
 
-@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+#@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
```
ns-rse, line 135: Ditto, why is the fixture commented out here?

ns-rse (changes requested):
Very minor comment about a method `.sum()` for summing arrays and quite inconsequential, but curious why the `pytest-mpl` fixtures have been removed in one file, I think this means that the tests of the images will no longer run as `pytest` won't know to look for a figure for that test.

tests/test_filters.py
MaxGamill-Sheffield: Ah I didn't realise this. I've made the appropriate changes

tests/test_grains_minicircle.py
MaxGamill-Sheffield: Oops, I think it was back when I was looking at specific test outcomes and I forgot to change them back. Ammended now.

ns-rse (approved):
Thanks @MaxGamill-Sheffield looks good to go, great work getting this done and as noted in the meeting earlier this week, a fairly critical aspect to making `dev` > `main` so thank you. :+1:

PR #280, opened 2022-08-31
Maxgamill sheffield/232 move gaussian filter

MaxGamill-Sheffield:
This PR moves the `gaussian_filter()` function from the `grains` module into the `filters` module and outputs the final Filters image as the "gaussian_filtered" one. This is then used as a starting point for other modules such as grains and grainstats.

Additionally, the plotting section has now been added to each individual Filters/Grains section, tests have been created and updated, and regression tests reset.

The difference in produced test images, statistics and the number of lower grains found comes from the fact that previously, topostats used `zero_averaged_background` as the image passed into `Grains()`. This meant that all the thresholds were taken from this image instead of the gaussian filtered image (as obtaining thresholds was done at the start of `Grains.find_grains`).
As the passed image is now `gaussian_filtered`, this now generates a different threshold value and thus the produced images and statistics are slightly different.


tests/conftest.py
```diff
@@ -218,6 +218,20 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
     return grains
 
 
+@pytest.fixture
+def small_array_filters(small_array: np.ndarray, filter_config: dict, tmp_path) -> Grains:
+    """Grains object based on small_array."""
```
ns-rse, line 223: This is returning a `Filters` object not a `Grains`.

```suggestion
    """Filters object based on small_array."""
```

tests/conftest.py
```diff
@@ -218,6 +218,20 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
     return grains
 
 
+@pytest.fixture
+def small_array_filters(small_array: np.ndarray, filter_config: dict, tmp_path) -> Grains:
```
ns-rse, line 222: This is returning a `Filters` object not a `Grains`.

```suggestion
def small_array_filters(small_array: np.ndarray, filter_config: dict, tmp_path) -> Filters:
```

tests/conftest.py
```diff
@@ -267,7 +281,7 @@ def minicircle_pixels(minicircle_channel) -> Filters:
 
 
 @pytest.fixture
-def minicircle_initial_align(minicircle_pixels: np.array) -> Filters:
+def minicircle_initial_align(minicircle_pixels: Filters) -> Filters:
```
ns-rse, line 284: Thanks for adding in missing typehints here and in other places :+1:

tests/test_grains.py
```diff
@@ -14,19 +13,6 @@
 TOLERANCE = {"atol": 1e-07, "rtol": 1e-07}
 
 
-def test_gaussian_filter(small_array_grains: Grains, grains_config: dict) -> None:
```
ns-rse, line 17: I think this test should be moved to `test_filters_minicircle.py` so that we retain a unittest of the `Filters.gaussian_filter()` against `minicircle.spm`.

topostats/filters.py
```diff
@@ -166,14 +172,18 @@ def flatten_image(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarra
         image = self.remove_tilt(image, mask)
         return image
 
-    def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarray:
+    def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> dict:
```
ns-rse, line 175: Thanks also for correcting the types and docstrings here. :+1:

topostats/filters.py
```diff
@@ -202,12 +224,12 @@ def align_rows(self, image: np.ndarray, mask=None) -> np.ndarray:
 
         # Adjust the row medians accordingly
         # FIXME : I think this can be done using arrays directly, no need to loop.
-        for i in range(image.shape[0]):
+        for i in range(image_cp.shape[0]):
             # if np.isnan(row_median_diffs[i]):
             #     LOGGER.info(f"{i} Row_median is nan! : {row_median_diffs[i]}")
-            image[i] -= row_median_diffs[i]
+            image_cp[i] -= row_median_diffs[i]
```
ns-rse, line 230: Surprised this doesn't fail, where have we defined `row_median_diffs` within this function? :thinking:

topostats/filters.py
```diff
@@ -187,12 +197,24 @@ def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndar
...
 
     def align_rows(self, image: np.ndarray, mask=None) -> np.ndarray:
-        """Returns the input image with rows aligned by median height"""
+        """Returns a copy of the input image with rows aligned by median height.
+        
+        Parameters
+        ----------
+        image: np.ndarray
+            2-D image to align rows.
+
```
ns-rse, line 206:
```suggestion
        mask: np.ndarray
            Boolean array of points to mask.
```

topostats/filters.py
```diff
@@ -187,12 +197,24 @@ def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndar
         return medians
 
     def align_rows(self, image: np.ndarray, mask=None) -> np.ndarray:
```
ns-rse, line 199:
```suggestion
    def align_rows(self, image: np.ndarray, mask: np.ndarray=None) -> np.ndarray:
```

topostats/filters.py
```diff
@@ -166,14 +172,18 @@ def flatten_image(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarra
...
 
-    def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> np.ndarray:
+    def row_col_medians(self, image: np.ndarray, mask: np.ndarray = None) -> dict:
         """Returns the height value medians for the rows and columns.
 
+        Parameters
+        ----------
+        image: np.ndarray
+            2-D image to calculate row and column medians.
+
```
ns-rse, line 182:
```suggestion
        mask: np.ndarray
            Boolean array of points to mask.
```

topostats/filters.py
```diff
@@ -220,20 +242,32 @@ def _row_median_diffs(row_medians: np.array, median_row_height: float) -> np.arr
         return row_medians - median_row_height
 
     def remove_tilt(self, image: np.ndarray, mask=None) -> np.ndarray:
```
ns-rse, line 244:
```suggestion
    def remove_tilt(self, image: np.ndarray, mask: np.ndarray=None) -> np.ndarray:
```

topostats/filters.py
```diff
@@ -220,20 +242,32 @@ def _row_median_diffs(row_medians: np.array, median_row_height: float) -> np.arr
...
-        """Returns the input image after removing any linear plane slant"""
-        medians = self.row_col_medians(image, mask)
+        """Returns a copy of the input image after removing any linear plane slant.
+    
+        Parameters
+        ----------
+        image: np.ndarray
+            2-D image to align rows.
+
```
ns-rse, line 251:
```suggestion
        mask: np.ndarray
            Boolean array of points to mask.
```

ns-rse (changes requested):
Thanks for tackling this @MaxGamill-Sheffield notionally "small" changes are becoming quite involved!

I noticed a lot of the arrays have changed in `test_grains_minicircle.py`. Moving the code shouldn't affect this, but I noticed the `gaussian_size` defined in the `tests/resources/sample_config.yaml` (and `tests/resources/process_scan_config.yaml`) has changed (from `2.0` to `0.5`). Is there any particular reason for doing this? It might be a more rational default value for the default configuration but for tests we just want to know that we get the same answer with the same input and so its not really to big an issue if its a larger value than would be sensible. In the long run its kind of created more work updating all the tests to work with a new value. 

With this in mind and an eye on future work I think we should try and come up with reasonable defaults to use both in the default configuration, but particularly in tests as it creates unnecessary work to be updating tests in this manner.

tests/conftest.py
MaxGamill-Sheffield: My bad, must have been overlooked when moving bits around

tests/test_grains.py
MaxGamill-Sheffield: This test has been moved to `test_filters.py` already (lines 178-189). Although, neither this nor the earlier test were agains `minicircle.spm`, only the random array.

topostats/filters.py
MaxGamill-Sheffield: Thanks for adding this!

topostats/filters.py
MaxGamill-Sheffield: It is defined 7 lines above:
`row_median_diffs = self._row_median_diffs(row_medians, median_row_height)`
Although the var `row_median_diffs` and the func `row_median_diffs` make things a little confusing. Should I rewrite these names?

MaxGamill-Sheffield:
> Thanks for tackling this @MaxGamill-Sheffield notionally "small" changes are becoming quite involved!
> 
> I noticed a lot of the arrays have changed in `test_grains_minicircle.py`. Moving the code shouldn't affect this, but I noticed the `gaussian_size` defined in the `tests/resources/sample_config.yaml` (and `tests/resources/process_scan_config.yaml`) has changed (from `2.0` to `0.5`). Is there any particular reason for doing this? It might be a more rational default value for the default configuration but for tests we just want to know that we get the same answer with the same input and so its not really to big an issue if its a larger value than would be sensible. In the long run its kind of created more work updating all the tests to work with a new value.
> 
> With this in mind and an eye on future work I think we should try and come up with reasonable defaults to use both in the default configuration, but particularly in tests as it creates unnecessary work to be updating tests in this manner.

@ns-rse It turns out that there *were* changes in the arrays due to the move as this changed the array to which thresholds were applied. The array that was passed into `Grains()` used `zero_averaged_background` as the `self.image` (line 261-267 of grains), whereas now this is the `gaussian_filtered` array which has slightly different values and thus thresholds. Maybe it is worth a discussion with the experimentalists about wether the raw or smoothed data should be used for this step.

Additionally, as this was a big merge I didn't want to commit to making large changes on this branch and having to readjust smaller changes/fixes later on. So, I took your advice and created branches against this one to make small fixes (should I have left these for review too before merging them?). One of these fixes was to change the incorrectly calculated gaussian size (`*` when it should have been `/`). This made the gaussians very blurry (as expected as the gaussian size of 2nm covered 4 pixels), so I then reduced this from 2 -> 0.5nm in order to regain similar images, however these were still different.

tests/test_grains.py
MaxGamill-Sheffield: The more tests the merrier - I noticed there wasn't one so I've added a test here

MaxGamill-Sheffield:
To add, in terms of trying to match the Gwyddion results, the overall results get worse when using the `gaussian_filtered` image in-place of the `zero_averaged_background` for obtaining the initial thresholds. Only buy a max of 0.6% and overall 0.14% compared to the current `dev` values.

ns-rse (approved):
Thanks as always for the explanation @MaxGamill-Sheffield 

I guess checking with the experimentalists is going to be key here, whether that should diverge from what Gwyddion does is really not something I feel qualified to answer.

MaxGamill-Sheffield:
I'll discuss with Eddie in tomorrows meeting and wait to merge until then, although I think Alice has a recommended Gwyddion procedure in the lab we can use to compare workflows.

MaxGamill-Sheffield:
Speaking with Jean and Eddie this morning, we looked through the Nature Methods paper, Old TopoStats code and this code to identify differences in filtering steps.

Seen below we see that all three differ *but* all end using a gaussian filter to be passed into the grain finding functions:
Screenshot 2022-09-08 at 10 53 00

Other flattening methods such as polynomial levelling may have to be added later and the Filters order re-jigged to get values closer to old TopoStats in the future.

For now, I will merge this to dev.

PR #284, opened 2022-09-01
Maxgamill sheffield/281 area thresholding

MaxGamill-Sheffield:
So far I have implemented area thresholding (in nm^2) so that this could be applied to images of various scales (as opposed to a `std_dev`-type method which would give different means and thus results depending on the amount/type of impurities within the sample).

In doing so, I have essentially "overwritten" the `remove_small_objects` function. Default behaviour is to use these absolute threshold values in place of the aforementioned function although if the sample is of unknown size, one can set the config area threshold values to `None` and the `remove_small_objects` function will be used instead.


MaxGamill-Sheffield:
After meeting, this area replacement would work well as a factory method where values are specified and if not, the already present method using OTSU should be done

tests/conftest.py
```diff
@@ -494,6 +493,20 @@ def minicircle_small_objects_removed(minicircle_minimum_grain_size: np.array) ->
...
+def minicircle_area_thresholding(minicircle_grain_labelled_all: np.array, grains_config: dict) -> Grains:
+    """Small objects removed."""
+    minicircle_grain_labelled_all.directions["upper"][
+        "removed_small_objects"
+    ] = minicircle_grain_labelled_all.area_thresholding(
+        minicircle_grain_labelled_all.directions["upper"]["labelled_regions_01"],
+        grains_config["area_absolute_threshold_upper"],
+        grains_config["area_absolute_threshold_lower"]
+    )
+    # for some reason this is returning an image with most molecules gone
```
ns-rse, line 506: Is that because they are outside of the area threshold?

tests/test_grains_minicircle.py
```diff
@@ -105,6 +105,25 @@ def test_remove_small_objects(
...
+    assert minicircle_area_thresholding.directions["upper"]["removed_small_objects"].shape == (1024, 1024)
+    assert minicircle_area_thresholding.directions["upper"]["removed_small_objects"].sum() == 42378
+    plotting_config = {**plotting_config, **plot_dict["removed_small_objects"]}
+    fig, _ = Images(
+        data=minicircle_area_thresholding.directions["upper"]["removed_small_objects"],
+        output_dir=tmp_path,
+        pixel_to_nm_scaling_factor=minicircle_area_thresholding.pixel_to_nm_scaling,
+        **plotting_config,
+    ).plot_and_save()
+    return fig
```
ns-rse, line 124: Could you add the image to the repository and commit it please?

I think it should be something like...

```
git add tests/resources/img/test_area_thresholding.png
```

...assuming you've generated it with the `pytest tests/test_grains_minicircle.py --mpl-generate-path=tests/resources/img`.

tests/resources/sample_config.yaml
```diff
@@ -23,6 +23,8 @@ grains:
   threshold_std_dev: 1.0
   threshold_absolute_lower: -1.0
```
ns-rse, line 24: We could correct this too...

```suggestion
  threshold_absolute_lower: 1.0
```

ns-rse (commented):
Sounds as though you're going to re-write this based on your comments but some minor feedback.

ns-rse:
> After meeting, this area replacement would work well as a factory method where values are specified and if not, the already present method using OTSU should be done

If I've understood the code correctly in this PR the area threshold being used is based on a bounding box. The current methods of thresholding based on size are, and hopefully @SylviaWhittle can correct me if I'm wrong, based on the number of connected pixels.

Whilst both approaches will result in a number (one an area, the other the number of connected pixels) their type and meaning are subtly different, to which end careful consideration of whether this should be included in the current thresholding functions (/factory method) should be given.

tests/conftest.py
MaxGamill-Sheffield: I don't believe so as these are the current values I use when normally running TS to gather the expected amount of grains.

Using some print statements I find the following:
Normal TS run:
Labelled grains found: 126
Initial grains found: 126
Final grains found: 21
Thresh: (500, 800)

Test TS run:
Labelled grains found: 41
Initial grains found: 41
Final grains found: 3
Thresh: (500, 800)

So I'm thinking I might be using the wrong image?

tests/test_grains_minicircle.py
MaxGamill-Sheffield: > > After meeting, this area replacement would work well as a factory method where values are specified and if not, the already present method using OTSU should be done
> 
> If I've understood the code correctly in this PR the area threshold being used is based on a bounding box. The current methods of thresholding based on size are, and hopefully @SylviaWhittle can correct me if I'm wrong, based on the number of connected pixels.
> 
> Whilst both approaches will result in a number (one an area, the other the number of connected pixels) their type and meaning are subtly different, to which end careful consideration of whether this should be included in the current thresholding functions (/factory method) should be given.

So this method counts the number of pixels each grain has and multiplies it by `px_to_nm_scaling ** 2` to obtain an area, then it uses this to threshold. However, I've just realised I could go the other way and reuse the `remove_noise` function to do this too instead of the one I made.

@SylviaWhittle approach uses `remove_small_objects` which first calculates a `minimum_grain_size` via finding grains which are greater than a threshold obtained via otsu (i.e. reduces grain area list to those of bigger objects). Then this `minimum _grain_size is calculated by `np.median(grain_areas) - (1.5 * (np.quantile(grain_areas, 0.75) - np.quantile(grain_areas, 0.25)))` and finally `remove_small_objects` removes objects < this value. Let me know if I've got that right?

MaxGamill-Sheffield: Actually, skimage's remove small objects func can't remove objects larger than a value nor is there a similar func to do so so I'll keep mine

MaxGamill-Sheffield:
Should now test well. It should be noted that changes to the regtests are due to lower grains no longer being found as they don't lie within the area threshold of 400-600nm2.

Is this behaviour we want? do we want a lower threshold when both or lower directions are specified?

SylviaWhittle:
That's an interesting question - yeah perhaps we would want a lower threshold for lower grains? I would defer to the experimentalists on this though.

SylviaWhittle:
> > After meeting, this area replacement would work well as a factory method where values are specified and if not, the already present method using OTSU should be done
> 
> If I've understood the code correctly in this PR the area threshold being used is based on a bounding box. The current methods of thresholding based on size are, and hopefully @SylviaWhittle can correct me if I'm wrong, based on the number of connected pixels.
> 
> Whilst both approaches will result in a number (one an area, the other the number of connected pixels) their type and meaning are subtly different, to which end careful consideration of whether this should be included in the current thresholding functions (/factory method) should be given.

Yes, it should be based on the number of pixels.

tests/test_grains_minicircle.py
SylviaWhittle: > > > After meeting, this area replacement would work well as a factory method where values are specified and if not, the already present method using OTSU should be done
> > 
> > 
> > If I've understood the code correctly in this PR the area threshold being used is based on a bounding box. The current methods of thresholding based on size are, and hopefully @SylviaWhittle can correct me if I'm wrong, based on the number of connected pixels.
> > Whilst both approaches will result in a number (one an area, the other the number of connected pixels) their type and meaning are subtly different, to which end careful consideration of whether this should be included in the current thresholding functions (/factory method) should be given.
> 
> So this method counts the number of pixels each grain has and multiplies it by `px_to_nm_scaling ** 2` to obtain an area, then it uses this to threshold. However, I've just realised I could go the other way and reuse the `remove_noise` function to do this too instead of the one I made.
> 
> @SylviaWhittle approach uses `remove_small_objects` which first calculates a `minimum_grain_size` via finding grains which are greater than a threshold obtained via otsu (i.e. reduces grain area list to those of bigger objects). Then this `minimum _grain_size is calculated by `np.median(grain_areas) - (1.5 * (np.quantile(grain_areas, 0.75) - np.quantile(grain_areas, 0.25)))`and finally`remove_small_objects` removes objects < this value. Let me know if I've got that right?

Correct. But it can't remove larger objects, only smaller ones. I think we might want to shift to just using our own custom Python function rather than using an incomplete (can't do larger grains) version from scikit?

topostats/grains.py
```diff
@@ -183,6 +187,38 @@ def remove_small_objects(self, image: np.array, **kwargs):
             return small_objects_removed > 0.0
         return image
 
+    def area_thresholding(self, image: np.ndarray, upper: float, lower: float):
```
SylviaWhittle, line 190: I know you're already working on this, but just adding a comment for transparency. 

Perhaps we should remove the Scikit area thresholding method and replace it with Max's method using only the lower size limit. This would make the code's behaviour more consistent between configurations. (Not having two different methods of thresholding size).

MaxGamill-Sheffield:
Updated a few things:
- `area_thresholding` using more optimised numpy code
- Now there is an extra area threshold so different molecules can be found in upper and lower masks
- The `area_thresholding` func now takes a labelled image and returns a labelled image too (instead of a binary mask)
- I've shuffled the order of the grain finding as `tidy_boarder` and `remove noise` labelled the image anyway so if this is done first it should save some time re-labelling things (this meant the test labelled image changed). This means I've also shuffled the order of the plot_dict too so numbers are correct
- An issue where for `direction = both` only produces the lower image has been resolved
- An extra test for the area threshold added
- Regtest changed due to finding a lower grain now which I think should be kept as I think finding no lower grains is a recipe for something to go wrong

I think that's it - any Q's feel free to ask

topostats/grains.py
MaxGamill-Sheffield: Done

MaxGamill-Sheffield:
Oooo and the default config has been updated

tests/conftest.py
```diff
@@ -458,7 +457,7 @@ def minicircle_grain_clear_border(minicircle_grain_mask: np.array) -> Grains:
 
 @pytest.fixture
 def minicircle_grain_remove_noise(minicircle_grain_clear_border: np.array) -> Grains:
-    """Cleared borders."""
+    """Remove Noise."""
```
ns-rse, line 460: I thought the agreed term was "crap" :wink: 

```suggestion
    """Test Removing Noise."""
```

tests/resources/sample_config.yaml
```diff
@@ -21,8 +21,10 @@ grains:
   threshold_method: otsu # Options : otsu, std_dev, absolute
   otsu_threshold_multiplier: 1.7
   threshold_std_dev: 1.0
-  threshold_absolute_lower: -1.0
+  threshold_absolute_lower: 1.0
   threshold_absolute_upper: 1.0
+  upper_absolute_area_threshold: [400,600] # above surface [Low, High] in nm^2 (also takes null)
```
ns-rse, line 26: The other two configs have `[500, 800]` just wanted to check that these are meant to be as they are, I'd expect the tests you've written work with these.

MaxGamill-Sheffield: So yep this took me a while to figure out. But basically the differences are due to the grain finding method in this file being "otsu" which with the default values makes the grains a little skinnier and thus reduces their area compared to the other files which use the standard deviation method.

ns-rse: Great, thanks for the clarification.

It sounds as though having sensible "`default`" configuration is going to be challenging!

tests/test_grains.py
```diff
@@ -13,6 +14,61 @@
...
+    "area_thresh, expected",
+    [
+        ([None, None], grain_array),
+        ([None, 8], grain_array2),
+        ([3,6], grain_array3),
+        ([8,11], grain_array4)
+    ]
+)
+def test_known_array_threshold(area_thresh, expected) -> None:
+    "Tests that arrays the thresholded on size as expected"
```
ns-rse, line 67: Sorry another pedantic typo
```suggestion
    "Tests that arrays are thresholded on size as expected."
```

topostats/grains.py
```diff
@@ -76,10 +78,14 @@ def __init__(
         self.threshold_std_dev = threshold_std_dev
         self.threshold_absolute_lower = threshold_absolute_lower
         self.threshold_absolute_upper = threshold_absolute_upper
+        self.absolute_area_threshold_dict = {
```
ns-rse, line 81: Would it be easier/simpler to make the configuration file provide a dictionary?

Something like

```yaml
  absolute_area_thresholds:
    upper: [500, 800] # above surface [Low, High] in nm^2 (also takes null)
    lower: [null, null] # above surface [Low, High] in nm^2 (also takes null)
```

Would mean you can pass in `absolute_area_threshold` as a dictionary directly.

```python
        self.absolute_area_threshold = absolute_area_threshold     
```

topostats/run_topostats.py
```diff
@@ -264,7 +264,7 @@ def process_scan(
                 plotting_config["plot_dict"][plot_name]["output_dir"] = Path(_output_dir)
                 Images(
                     filtered_image.images["gaussian_filtered"],
-                    filename=filtered_image.filename + "_processed_masked",
+                    filename=filtered_image.filename + f"_{direction}_processed_masked",
```
ns-rse, line 267: Could just be a single f-string

```suggestion
                    filename=f"{filtered_image.filename}_{direction}_processed_masked",
```

ns-rse (commented):
Brilliant work as always @MaxGamill-Sheffield 

Mostly very minor semantics (several of which you've already addressed/resolved), but perhaps worth considering pushing the creation of a dictionary back into the configuration file. I'm not beholden to doing so though.

Sorry for the disjointed reviewing, was trying to multi-task.

topostats/grains.py
MaxGamill-Sheffield: Yes I like this idea a lot! I'll make this change in a sec.

ns-rse (approved):
Thanks for incorporating the suggestion of area thresholds as a dictionary in the config @MaxGamill-Sheffield , considerably quicker than the time it took me to review (although I was on leave last week).

All good to merge :+1:

Issue #271, opened 2022-08-23
Handle filenames with multiple periods in


ns-rse:
**Describe the bug**

When a filename has two periods in it we get crashes e.g. `mini.circle.spm`...

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/home/neil/work/projects/topostats/TopoStats/topostats/run_topostats.py", line 305, in process_scan
    plot_and_save(
  File "/home/neil/work/projects/topostats/TopoStats/topostats/plottingfuncs.py", line 92, in plot_and_save
    plt.savefig(output_dir / filename)
  File "/home/neil/.virtualenvs/topostats/lib/python3.10/site-packages/matplotlib/pyplot.py", line 979, in savefig
    res = fig.savefig(*args, **kwargs)
  File "/home/neil/.virtualenvs/topostats/lib/python3.10/site-packages/matplotlib/figure.py", line 3046, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/home/neil/.virtualenvs/topostats/lib/python3.10/site-packages/matplotlib/backend_bases.py", line 2259, in print_figure
    canvas = self._get_output_canvas(backend, format)
  File "/home/neil/.virtualenvs/topostats/lib/python3.10/site-packages/matplotlib/backend_bases.py", line 2188, in _get_output_canvas
    raise ValueError(
ValueError: Format 'circle_processed' is not supported (supported formats: eps, jpeg, jpg, pdf, pgf, png, ps, raw, rgba, svg, svgz, tif, tiff)
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/neil/.virtualenvs/topostats/bin/run_topostats", line 33, in <module>
    sys.exit(load_entry_point('topostats', 'console_scripts', 'run_topostats')())
  File "/home/neil/work/projects/topostats/TopoStats/topostats/run_topostats.py", line 394, in main
    for img, result in pool.imap_unordered(processing_function, img_files):
  File "/usr/lib/python3.10/multiprocessing/pool.py", line 873, in next
    raise value
ValueError: Format 'circle_processed' is not supported (supported formats: eps, jpeg, jpg, pdf, pgf, png, ps, raw, rgba, svg, svgz, tif, tiff)
```

This should be handled automatically.

[...]

PR #286, opened 2022-09-06
Maxgamill sheffield/hotfix overwriting plots

MaxGamill-Sheffield:
This fixes an issues where image overwriting can occur due to the filenames containing an extension i.e:

`abc.0_001.spm` is passed into the filename argument as `abc.0_001`.
So in `plottingfuncs.py`, the `<path>.with_suffix(<ext>)` replaces the `.0_001` with `<ext>`.
This replaces the image if an image for `abc.0_000.spm` already exists.

The fix is to ensure no file extensions are input into filename, hence the removal from `plotting_dictionary.yaml` and in the tests.

### DIFF
--- a/tests/test_plottingfuncs.py
+++ b/tests/test_plottingfuncs.py
@@ -26,7 +26,7 @@ def test_save_figure(
     fig, ax = Images(
         data=image_random,
         output_dir=tmp_path,
-        filename="result.png",
+        filename="result",
         data2=data2,
         colorbar=axes_colorbar,
         axes=axes_colorbar,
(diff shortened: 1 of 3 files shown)


MaxGamill-Sheffield:
> Does this solve the problems of multiple periods in filenames as described in #271 ?
> 
> When I started work on that #ns-rse/271-filenames I converged on the same solution as you have here of removing the file extension from the output filenames too (e.g. see [example.yaml](https://github.com/AFM-SPM/TopoStats/blob/ns-rse/271-filenames/config/example.yaml) although we now have that split into the default YAML files in the package that are loaded automatically so my branch is effectively obsolete, or needs rebasing).
> 
> I think I recall you included the changes for handling #271 in a recent PR.

@ns-rse  yep you're right,I believe #271 was fixed in an earlier PR when we refactored `plotting_funcs.py`, I just hadn't realised then that *that* was the PR which prevented the errors causing that issue from appearing. This however was an oversight on my part in which no explicit error if raised, just that files are overwritten as they have the same name. This PR should resolve that.

ns-rse (approved):
Thanks for the clarification and sensible to tidy this up by removing the file extension which is explicit from the config.

PR #287, opened 2022-09-07
Maxgamill sheffield/hotfix stretched non sq imgs

MaxGamill-Sheffield:
This fix stops non-square images that are plotted (not `imsave`'d) from being stretched by swapping the extents of the images around - this might be due to the way `imshow` plots are made. This also meant modifying the bounding boxes and removing hardcoded `1024`'s.

An additional test has been added to check the plot of a non-square image and its bounding boxes.


tests/test_plottingfuncs.py
```diff
@@ -164,3 +164,25 @@ def test_plot_and_save_zrange(minicircle_grain_gaussian_filter: Grains, plotting
...
+
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_plot_and_save_non_square_bounding_box(
+    minicircle_grain_coloured: Grains,
+    minicircle_grain_region_properties_post_removal: Grains,
+    plotting_config: dict,
+    tmp_path: Path,
+) -> None:
+    """Test plotting bounding boxes"""
+    print(minicircle_grain_region_properties_post_removal)
```
ns-rse, line 177: Extraneous `print()` from debugging?

ns-rse (dismissed):
Very minor `print()` needs removal, then good to go :+1:

PR #292, opened 2022-09-16
289 | Factory pattern to skeletonize molecules

ns-rse:
Introducing a factory pattern to skeletonize molecules using methods from [skimage.morphology](https://scikit-image.org/docs/stable/api/skimage.morphology.html).

Its not actually used anywhere yet but there are unit-tests in place to ensure the different methods work with both circular and linear (dummy) molecules. Next step will be replacing the existing methods with calls to this and introducing configuration options to the config files.


tests/tracing/test_skeletonize.py
```diff
@@ -0,0 +1,281 @@
...
+    assert isinstance(test, np.ndarray)
+    assert test.ndim == 2
+    assert test.shape == (21, 21)
+    assert test.sum() == 36
+    np.testing.assert_array_equal(test, CIRCULAR_TARGET)
+
+
+def test_skeletonize_circular_medial_axis(skeletonize_circular_bool_int: np.ndarray) -> None:
+    """Test the medial axis method of skeletonization on a circular object."""
+    CIRCULAR_MEDIAL_AXIS_TARGET = np.array(
```
MaxGamill-Sheffield, line 64: This array looks unusual (not what I'd expect from skeletonising a square), have we tested this method on anything else to see if its results are useful for our data? If not we might not need it.

MaxGamill-Sheffield (approved):
Good job on cleaning up conftest and thresholds files!

And this looks good although I have a silly question - why have we made a wrapper for the skimage skeletonise method with all the extra `method='lee'` etc when we could use `method=<method>` and shorten the amount of code?

Also will skimage not have tests for this themselves? Is testing skimage's code in our code best practice?

ns-rse:
> Good job on cleaning up conftest and thresholds files!

Ta was mostly just applying `black` and then checking with `pylint` which had thrown an unnecessary `elif` error in the `_get_skeletonize()` function and I knew I'd made the same mistake in `thresholds()`. Easy enough to tidy up and include here.

> And this looks good although I have a silly question - why have we made a wrapper for the skimage skeletonise method with all the extra `method='lee'` etc when we could use `method=<method>` and shorten the amount of code?

Not silly at all, this is another instance of a "Factory Method" [design pattern](https://refactoring.guru/design-patterns/python) that makes it flexible as its not just the `zhang` or `lee` method of `skeletonize()` from scikit-image but also two completely separate functions `medial_axis()` and `thin()`.

Such a pattern means that our function `get_skeleton()` conforms to the *O*pen component of the [SOLID design principles](https://devopedia.org/solid-design-principles), that is it is "Open" to extension if for example we wanted to make skeletonizing available using the [OpneCV](https://medium.com/analytics-vidhya/skeletonization-in-python-using-opencv-b7fa16867331) method. We would then modify the code to add an option for `method='opencv'` without having to modify existing code which remains "Closed".



> Also will skimage not have tests for this themselves? Is testing skimage's code in our code best practice?

It does indeed duplicate testing to a large extent, whether this is "best" practice I'm unsure, but I think its "good" practice as it means we can quickly and easily identify if something changes  in `skimage` that impacts _our_ expected results. It can be a pain keeping track of changes in packages, and we'd have to put significant active effort into tracking the methods in skimage and watching out for changes. These sorts of simple, automated, tests throws up anything that actually impacts our workflow.

tests/tracing/test_skeletonize.py
ns-rse: Yes, I wasn't sure if this method would be useful but it is what the method of [medial axis](https://homepages.inf.ed.ac.uk/rbf/HIPR2/skeleton.htm) should look like (if I've understood the method correctly).

Happy to have it removed, but didn't see any harm in including it, perhaps it will be worth noting with a WARNING when I get round to writing documentation.

PR #301, opened 2022-10-04
Maxgamill sheffield/#243 si units

MaxGamill-Sheffield:
This SI unit PR will be put on hold. 

Currently all statistics are in SI, and the extracted `px_to_m` value is in SI and propagated through to everywhere it needs to be and all looks correct. Additionally, plots use a function which extracts the non-SI units from very small numbers to input into the graphs so they now say "um" on the axis if the image should be in um instead of nm.

The reason for the hold is that when it comes to extracting a unit from the header for the pixel values (width and lengths are fine), I cannot find a function for this in pySPM. As a result, I've hardcoded this into "nm" i.e. 1e-9 but this is not known if this is consistent across different data that may have an image zrange > 1um and as a result it would break.

First I will move io functions to the `topostats_io.py` file to standardise their output and try to implement other file types to see if this is consistent across different file types too.

Note: the tests are also not complete for this PR

### DIFF
--- a/config/example.yaml
+++ b/config/example.yaml
@@ -13,11 +13,11 @@ filter:
   threshold_std_dev: 1.0
   threshold_absolute_lower: -1.0
   threshold_absolute_upper: 1.0
-  gaussian_size: 0.5
+  gaussian_size: 0.5e-9 # extent of blurring in m
   gaussian_mode: nearest
 grains:
   run: true # Options : true, false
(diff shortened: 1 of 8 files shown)


MaxGamill-Sheffield:
Closing as solved with #445. 

#445 also addresses the floating point number issue when internal calculations are in SI and thus keeps internal calculations in `nm`, and only changes the output stats to SI units.

Issue #304, opened 2022-10-05
Support  ASD file format


ns-rse:
**Is your feature request related to a problem? Please describe.**

Splitting down #137 this issue requests support for `.asd` file format. This can be achieved using the [libasd](https://github.com/ToruNiina/libasd) package which reads the files and allows access to headers.

Currently loading of Bruker formats is done via `topostats.io.load_scan()` to avoid breaking functionality this should be converted to (another!) Factory Method that calls the appropriate routine for loading the given file format.

**Describe the solution you'd like**

Be able to load and process `.asd` files.

**Describe alternatives you've considered**
 N/A

ns-rse:
`libasd` is not available for the ARM architecture some OSX systems use (see [available builds](https://pypi.org/project/libasd/#files)).

@SylviaWhittle has requested a build of `libasd` for this architecture (see [here](https://github.com/ToruNiina/libasd/issues/11)).

For now the `libasd` Python package is an optional dependency. On none-ARM systems it can be installed from this cloned repository with...

```bash
pip install .[libasd]
```

...or directly with...

```bash
pip install libasd
```

On ARM architectures the solution is to build and install the package yourself as described in the [building yourself](https://github.com/ToruNiina/libasd#building-by-yourself) instructions.

Issue #300, opened 2022-10-04
Support JPK formats


ns-rse:
As we work towards supporting more fileformats (see #137) we have decided to prioritise being able to load and process JPK file formats. The [afmformats](https://github.com/AFM-analysis/afmformats/) package supports this format and so the `topostats.io` sub-module should be expanded to support this.

Currently loading of Bruker formats is done via `topostats.io.load_scan()` to avoid breaking functionality this should be converted to (another!) Factory Method that calls the appropriate routine for loading the given file format.

**Describe the solution you'd like**

Be able to load JPK files.

**Describe alternatives you've considered**
N/A

MaxGamill-Sheffield:
I've tried renaming the extension (as I recall it came with none) to all the JPK formats above but it still couldn't open. 

I get the message "jpk format not supported" when the file extension is `.jpk`, and the message "Bad zip file" when copying the extension above. This refers to the JPK file structure as it is actually a zipped archive with a top level header folder and subfolders containing data.

I have tried with another JPK loading library (https://gitlab.gwdg.de/ikuhlem/jpkfile) with the same things but I also get "Bad zip file".

However, Gwyddion (grrr) is able to load this with no issues, and can tell it is a jpk file:
<img width="149" alt="Screenshot 2022-10-12 at 16 12 55" src="https://user-images.githubusercontent.com/91465918/195381131-f00c58c2-ce38-42dd-9a4f-2d7ca4650114.png">

So I'm at a bit of a loss

ns-rse:
I have two files the original I got from the shared drive...

`force-save-2019.01.11-12.23.32.518.jpk-force` 

...this is a zip archive of sorts (Google Drive recognises it as such).

```
import jpkfile

sample_jpk = jpkfile.JPKFile("force-save-2019.01.11-12.23.32.518.jpk-force")
/home/neil/.virtualenvs/topostats/lib/python3.10/site-packages/jpkfile/jpkfile.py:92: UserWarning: Did not find encoder parameters for channel smoothedCapacitiveSensorHeight!
  warnings.warn("Did not find encoder parameters for channel {}!".format(split[3][:-4]))
/home/neil/.virtualenvs/topostats/lib/python3.10/site-packages/jpkfile/jpkfile.py:92: UserWarning: Did not find encoder parameters for channel smoothedMeasuredHeight!
  warnings.warn("Did not find encoder parameters for channel {}!".format(split[3][:-4]))
 dir(sample_jpk)
[...]

Issue #137, opened 2022-05-13
Support loading and processing multiple file formats


ns-rse:
**Is your feature request related to a problem? Please describe.**
The ability to handle multiple file formats would be very useful.

**Describe the solution you'd like**
A single function or class to `io.load_scan(file_path)` that uses a [Factory Method](https://realpython.com/factory-method-python/) design pattern to use the appropriate functions/packages to load the file and return a NumPy array ready to be processed.

**Describe alternatives you've considered**
Currently we support Bruker formats using [pySPM](https://github.com/scholi/pySPM) but what other formats are commonly used and are there other Python packages out there that load these formats?

**Additional context**
At the same time this should address #132 

Specific issues pertaining to individual file formats or individual issues.

* #300 
* #304
* #309 
* #414

bobturneruk:
Several High Speed AFM users have expressed an interest in using TopoStats - LibASD would enable High Speed AFM file formats to be opened, so might be a priority.

bobturneruk:
Should `io.py` hold some of the stuff that's currently in `filters.py`? Probably, yes. This means all the stuff that has to know about file format and headers is in one place and `io.py` is more of a self contained module.

MaxGamill-Sheffield:
e.g. a `pySPM` function is currently (`.get_channel`) being used in the filters class within the `extract_channel` function, this will probably break if loading via a different library as it will not be a `pySPM.Bruker` object.

I think the `Filters` class should have a universal input (image, px_to_nm and general config file stuff) and thus the `io` should be universal output of the image and anything other scripts need from the image headers.

ns-rse:
> can we point to examples of files we would like to be able to open? 

The `afmformats` package has sample files that it uses for tests, we could perhaps borrow those?

> I think also we may wish to create some subsidiary issues pertaining to specific file formats as we get closer to implementation.

Agree with this, I've created a specific issue for supporting JPK (see #300 ) and added it to the Milestone 3. I'll edit the top comment to have a list of issues for specific file formats.

> I think the Filters class should have a universal input (image, px_to_nm and general config file stuff) and thus the io should be universal output of the image and anything other scripts need from the image headers.

Agree with this whole-heartedly. Aside from the image (`np.ndarray` of heights) and the Pixel-to-nm scaling what other parameters might we want to extract across images consistently and pass onto the next step in processing (i.e. `Filters()`).

ns-rse:
[...]

PR #306, opened 2022-10-07
Switch to broader LoadScan class

ns-rse:
This is the start of work on #137 and associated issues #300 and #304

@MaxGamill-Sheffield did the hard work of removing functionality that loads scans from `Filters()` and replaced the original `io.load_scan()` function with the `LoadScan()` class.

I have sorted out all the tests in light of changes as many were dependent on fixtures that loaded data via `Filters()` and this no longer happens.

I've taken the time to lint some of the files too, applying Black formatting so there are afew others included here with things like trailing white space being removed or import orders being moved around etc.

By having this first step in the refactoring as a separate pull request beyond incorporating new methods (the jpw stuff doesn't work yet!) we hopefully keep PRs smaller, more focused and a little easier to review.


tests/conftest.py
```diff
@@ -159,42 +166,73 @@ def image_random_col_medians_masked() -> np.array:
...
-    filters.load_scan()
-    filters.extract_channel()
-    filters.extract_pixels()
-    filters.images["pixels"] = image_random
-    # filters.get_threshold(filters.images["pixels"])
+    test_filters.images["pixels"] = image_random
+    return test_filters
+
+
+# @pytest.fixture
```
MaxGamill-Sheffield, line 206: Should this be removed if commented out?

tests/conftest.py
```diff
@@ -234,68 +272,98 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
...
+    )
     return filters
 
 
-@pytest.fixture
-def minicircle_filename(minicircle) -> Filters:
-    """Extract the filename."""
-    minicircle.extract_filename()
-    return minicircle
+# @pytest.fixture
```
MaxGamill-Sheffield, line 317: Is this also able to be removed?

tests/conftest.py
```diff
@@ -234,68 +272,98 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
...
+#     minicircle.load_scan()
+#     return minicircle
 
 
-@pytest.fixture
-def minicircle_make_output_directory(minicircle) -> Filters:
-    """Make output directory."""
-    minicircle.make_output_directory()
-    return minicircle
+# @pytest.fixture
```
MaxGamill-Sheffield, line 331: ditto

tests/conftest.py
```diff
@@ -234,68 +272,98 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
...
+#     minicircle.make_output_directory()
+#     return minicircle
 
 
-@pytest.fixture
-def minicircle_channel(minicircle) -> Filters:
-    """Extract the image channel."""
-    minicircle.extract_channel()
-    return minicircle
+# @pytest.fixture
```
MaxGamill-Sheffield, line 338: ditto

tests/conftest.py
```diff
@@ -234,68 +272,98 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
...
+#     return minicircle
 
 
-@pytest.fixture
-def minicircle_pixels(minicircle_channel) -> Filters:
-    """Extract Pixels"""
-    minicircle_channel.extract_pixels()
-    minicircle_channel.extract_pixel_to_nm_scaling()
-    return minicircle_channel
+# @pytest.fixture
```
MaxGamill-Sheffield, line 345: ditto

tests/conftest.py
```diff
@@ -554,17 +622,18 @@ def grainstats(image_random: np.array, minicircle_filename: str, grainstats_conf
 def minicircle_grainstats(
     minicircle_grain_gaussian_filter: Filters,
     minicircle_grain_labelled_post_removal: Grains,
-    minicircle_pixels: float,
-    minicircle_filename: str,
+    # minicircle_pixels: float,
```
MaxGamill-Sheffield, line 625: can these commented lines also be removed?

tests/conftest.py
```diff
@@ -554,17 +622,18 @@ def grainstats(image_random: np.array, minicircle_filename: str, grainstats_conf
...
     grainstats_config: dict,
-    plotting_config: dict,
     tmp_path: Path,
 ) -> GrainStats:
     """GrainStats object."""
     return GrainStats(
         data=minicircle_grain_gaussian_filter.images["gaussian_filtered"],
         labelled_data=minicircle_grain_labelled_post_removal.directions["upper"]["labelled_regions_02"],
-        pixel_to_nanometre_scaling=minicircle_pixels.pixel_to_nm_scaling,
+        # pixel_to_nanometre_scaling=minicircle_pixels.pixel_to_nm_scaling,
```
MaxGamill-Sheffield, line 635: ditto

tests/test_filters.py
```diff
@@ -176,23 +113,25 @@ def test_row_col_medians_with_mask(
...
-    test_filters_random.images["pixels"]=test_filters_random.images["pixels"][:,0:512]
+    test_filters_random.images["pixels"] = test_filters_random.images["pixels"][:, 0:512]
     test_filters_random.images["zero_averaged_background"] = test_filters_random.average_background(
         image=test_filters_random.images["pixels"], mask=None
     )
     assert isinstance(test_filters_random.images["zero_averaged_background"], np.ndarray)
-    assert test_filters_random.images["zero_averaged_background"].shape==(1024, 512)
-    assert test_filters_random.images["zero_averaged_background"].sum() == 44426.48188033322
+    assert test_filters_random.images["zero_averaged_background"].shape == (1024, 512)
+    # assert test_filters_random.images["zero_averaged_background"].sum() == 44426.48188033322
```
MaxGamill-Sheffield, line 125: Is there a reason this is commented out?

tests/test_grainstats.py
```diff
@@ -195,6 +196,6 @@ def test_grainstats_get_triangle_height(base_point_1, base_point_2, top_point, e
 
 
 @pytest.mark.parametrize("edge_points, expected", [([[0, 0], [0, 1], [1, 0], [1, 1]], (1.0, 1.4142135623730951))])
-def test_get_min_max_ferets(edge_points, expected) -> None:
+def test_get_mini_max_ferets(edge_points, expected) -> None:
```
MaxGamill-Sheffield, line 199: is there a reason for this name change? The actual function is `get_max_min_ferets` if that is better?

topostats/filters.py
```diff
@@ -22,43 +21,36 @@ class Filters:
...
         ----------
-        img_path: Union[str, Path]
-            Path to a valid image to load.
-        channel: str
-            Channel to extract from the image.
+        image: np.ndarray
+            The raw image from the AFM.
+        filename: str
+            The filename (used for logging outputs only).
         amplify_level : float
```
MaxGamill-Sheffield, line 45: I should have added a docstring here for `px_to_nm_scaling` my bad

topostats/io.py
```diff
@@ -60,20 +65,101 @@ def write_yaml(config: dict, output_dir: Union[str, Path]) -> None:
...
+        -------
+        float
+            Pixel to nm scaling factor.
+        """
+        unit_dict = {
+            "nm": 1,
+            "um": 1e3,
+        }
+        px_to_real = channel_data.pxs()
+        # Why are both nm and um calculated here and then only nm used?
```
MaxGamill-Sheffield, line 129: To answer this, it doesn't (or at least it's not meant to (its been a while since I've looked at this)).

`px_to_real` I think returns `{x_coord: [value, unit], y_coord: [value, unit]`. So, the list that is returned for this is one of the x_coord `px_to_nm_scaling` and the y_coord `px_to_nm_scaling`. 

Usually these are the same, and this is what I've seen with the test data but with the high speed afm, there may be a difference so this is here as I thought it may be useful in the future because if there were a difference we could more easily propagate this through.

topostats/run_topostats.py
```diff
@@ -168,29 +171,39 @@ def process_scan(
...
-        filter_out_path = _output_dir / image_path.stem / "filters"
-        grain_out_path = _output_dir / image_path.stem / "grains"
+        filter_out_path = Path(_output_dir) / image_path.stem / "filters"
+        grain_out_path = Path(_output_dir) / image_path.stem / "grains"
+        filter_out_path.mkdir(exist_ok=True, parents=True)
         Path.mkdir(_output_dir / image_path.stem / "grains" / "upper", parents=True, exist_ok=True)
         Path.mkdir(_output_dir / image_path.stem / "grains" / "lower", parents=True, exist_ok=True)
 
+    # Extract image and image params
+    #   I don't know if this should be Load_scan(...) and
```
MaxGamill-Sheffield, line 181: Feel free to remove this comment too

MaxGamill-Sheffield (changes requested):
Great job on sorting out the tests for this change, and capturing some mistakes in the early tests.

A few comments about removing comments and commented code but mostly good!

One thing I'd mention is that to avoid confusion, would it be better to input into the test functions `image, px_to_nm, filename` rather than the whole `LoadScan` object or does this add diversity to upcoming tests/make it easier to implement?

Also as always thanks for the linting!

Also also, changing the files to find in the config to `.jpk` instead of `.spm` still finds the `minicircle.spm` so is the file listing working properly? (not toi do with this commit but I've only just noticed)
Screenshot 2022-10-10 at 16 53 26

Great work with the tests as always!

tests/test_grainstats.py
ns-rse: Fat fingers and being in the wrong buffer. Thanks for highlighting.

topostats/filters.py
ns-rse: Ah, good spot, added the other missing docstring fields too.

topostats/io.py
```diff
@@ -60,20 +65,101 @@ def write_yaml(config: dict, output_dir: Union[str, Path]) -> None:
...
+        # Has potential for non-square pixels but not yet implimented
+        pixel_to_nm_scaling = (
+            px_to_real[0][0] * unit_dict[px_to_real[0][1]],
+            px_to_real[1][0] * unit_dict[px_to_real[1][1]],
+        )[0]
+        LOGGER.info(f"[{self.filename}] : Pixel to nm scaling : {pixel_to_nm_scaling}")
+        return pixel_to_nm_scaling
+
+    # def load_jpk(self) -> None:
```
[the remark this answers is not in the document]
ns-rse: Not for now, I'm trying to make each function really small and split the loading of the file into `_load_jpk()` and the extraction of in `_extract_jpk()` but then call each with the method `load_jpk()`. It hopefully makes testing each method slightly easier. I left in as a place holder.

ns-rse:
> One thing I'd mention is that to avoid confusion, would it be better to input into the test functions `image`, `px_to_nm`, `filename` rather than the whole `LoadScan` object or does this add diversity to upcoming tests/make it easier to implement?

I don't have a strong opinion on that. Thinking about it (albeit briefly) it may be desirable to modify `Filters()` to accept a `LoadScan` object rather than three individual objects and to test that the `LoadScan` always has the required attributes. But there are multiple ways of skinning a cat and I can't think of strong arguments for or against at the moment.

> Also also, changing the files to find in the config to .jpk instead of .spm still finds the minicircle.spm so is the file listing working properly? (not toi do with this commit but I've only just noticed)

Ah, good catch, I'll look into that and write an appropriate test.

tests/test_filters.py
ns-rse: Sorry, wasn't paying attention, yes I left a note in the `FIXME` at the start of the test and was going to address this in a separate issue as I wanted to get these tests updated and ready to go before further work was done on other file formats. Its really important to take time to update tests when refactoring before then going onto implement new features otherwise you could end up without tests.

ns-rse:
Ok, props to @MaxGamill-Sheffield for testing with different file extensions and discovering `.spm` were still picked up.

Fairly important feature as we move to supporting more file formats and the reason was that the default value of `utils.find_images(base_dir, file_ext=".spm")` was being used in `run_topostats.main()` to find files, i.e.

```python
    img_files = find_images(config["base_dir"])
```

...when it _should_ have been...

```python
    img_files = find_images(config["base_dir"], file_ext=config["file_ext"])
```

This mean that whatever the value for `file_ext` was in the config was irrelevant, we only looked for `.spm`. I've incorporated the above change and tested locally and do not pick up any files when `file_ext: jpk` in the configuration and the `base_dir` only has `.spm`.

There is one outstanding issue with the non-square array where I disabled the `assert` that summed the array. I couldn't work out why these didn't match given the parent arrays did.

MaxGamill-Sheffield:
> > One thing I'd mention is that to avoid confusion, would it be better to input into the test functions `image`, `px_to_nm`, `filename` rather than the whole `LoadScan` object or does this add diversity to upcoming tests/make it easier to implement?
> 
> I don't have a strong opinion on that. Thinking about it (albeit briefly) it may be desirable to modify `Filters()` to accept a `LoadScan` object rather than three individual objects and to test that the `LoadScan` always has the required attributes. But there are multiple ways of skinning a cat and I can't think of strong arguments for or against at the moment.

I think I'd like to avoid this to make it more universal. While yes it would help when running in `run_topostats`, but if someone's preferred method of using the library and the `Filters` function is via notebooks or if they have another software which imports their files, they can still use `Filters` without instantiating a LoadScan object.

tests/test_filters.py
MaxGamill-Sheffield: cool, sounds good

MaxGamill-Sheffield (approved):
Looks good but just one question:
I see you've left space for tests of the other file types but does this mean we will have to add a file of each filetype to the repo just for the tests?

ns-rse:
> I see you've left space for tests of the other file types but does this mean we will have to add a file of each filetype to the repo just for the tests?

Possibly, I think it might be possible to [mock](https://changhsinlee.com/pytest-mock/) the images although mocking is not something I'm currently very familiar with, although I should be. Thus I think we will initially be able to write tests to load files, but for subsequent tests use mocked versions.  Can likely rewrite current tests to take advantage of this too.

PR #313, opened 2022-10-11
310 | Removes amplify_level from codebase

### LINKED ISSUE #310: Remove amplify_level from configuration and code.
**Describe the bug**

Currently `amplify_level` is a configuration option but its not actually used, all config files have the value set to `1.0` which has no impact on the actual data.

It should therefore be removed from the configuration and code base.

ns-rse:
Closes #310


MaxGamill-Sheffield (approved):
Looks good to me

PR #307, opened 2022-10-10
Adds contributing guidelines & refreshes documentation

ns-rse:
Resolves #259 

Adds a contribution guideline along with detailing the configuration options (see #201) and updating the documentation to be in-line with #305 which will reduce the need to update so many files when introducing new config options.

Includes the data dictionary from #235.

Also switches the theme used in the pages to be a bit more modern.

I've attempted to bring the introduction up-to-date but at the same time simplify it as I felt the existing text which mentioned all the classes and functions that are used was unnecessarily detailed and potentially confusing. I'd really appreciate expansion on this area. At the same time I think the flow-diagrams can also be improved, I sketched one out for DNATracing to help me understand the refactoring process that is on-going in the background and include it here. Also sketched one out for the `LoadScan` class in [137](https://github.com/AFM-SPM/TopoStats/issues/137#issuecomment-1271256444) although haven't yet incorporated it here since that is work in progress.

PR #314, opened 2022-10-11
305 | Removing config/example.yaml

### LINKED ISSUE #305: Remove config/example.yaml
Currently we have a default configuration file that is loaded automatically from `topostats/default_config.yaml` if no configuration file is specified when invoking `run_topostats`.

We also have `config/example.yaml` which pre-dates this and was used early on when developing and there was no default and it was invoked with `run_topostats -c config/example.yaml`.

Now that we have a default as well it means when new configuration options are added that we have to update both files. This is a duplication of effort and possibly error prone as @SylviaWhittle found the other day (see #296 and #297).

To which end I propose removing `config/example.yaml` to reduce this burden. At the same time the documentation will need updating, currently this resides in `README.md` but I think we should add a section to `docs/usage.md` to document the usage as well as the notebooks (#242) that should probably be merged in and made available.

ns-rse:
Closes #305 

I _haven't_ touched the `README.md` or other documentation as they are already reflected in #307 

I will update the as yet unmerged notebooks to reflect these changes (and likely put a PR out to merge the notebooks into `dev` after that).


MaxGamill-Sheffield (approved):
As you can imagine, this PR is working well.

I just want to make note of a behaviour change in that previously running `run_topostats -c conf/example.yaml` the config file could be changed without needing to `pip install .` again, however, now it cannot.

I realise that one could just install using `pip install -e .` for interactive updates but I'm curious which one will be used when on PyPi. Also I guess this file shouldn't really be changed, only copied and then the `-c` config specified which will get around this anyway.

ns-rse:
> As you can imagine, this PR is working well.
> 
> I just want to make note of a behaviour change in that previously running `run_topostats -c conf/example.yaml` the config file could be changed without needing to `pip install .` again, however, now it cannot.
> 
> I realise that one could just install using `pip install -e .` for interactive updates but I'm curious which one will be used when on PyPi. Also I guess this file shouldn't really be changed, only copied and then the `-c` config specified which will get around this anyway.

I've never seen `pip install -e <package-on-pypi>` used, that's not to say it can't be.

The documentation PR (#307) that is out for review clarifies that `topostats/default_config.yaml` should, as you note, be copied and modified.

PR #318, opened 2022-10-12
Maxgamill sheffield/#288 remove background

MaxGamill-Sheffield:
This work removes `background` from the config files. 

I have not updated the example.yaml file as this is in the process of being removed.

This is because it is unnecessary because its used in the [skimage.morphology.labels](https://scikit-image.org/docs/stable/api/skimage.morphology.html#skimage.morphology.label) function where the input image to this is either the `mask` (in which the background is 0 by design) or a labelled image from `area_threshold` (which has a 0 background) or `remove_small_objects` (which also has 0 background by default)


MaxGamill-Sheffield:
Tests failed due to network connection drop - re-running now

PR #321, opened 2022-10-13
Make libasd an optional install dependency

ns-rse:


MaxGamill-Sheffield (approved):
One question though, how would one continue to make the appropriate functions in io.LoadScan? Woulnd't an error be thrown when trying to `import libasd` at the top of the file if the user has only done `pip install .`?

ns-rse:
> One question though, how would one continue to make the appropriate functions in io.LoadScan? Woulnd't an error be thrown when trying to `import libasd` at the top of the file if the user has only done `pip install .`?

Correct. In this case I think we simply need to remove `libasd` completely and only have it on development branches until we have a resolution and are ready to release a version of TopoStats that supports such files.

PR #325, opened 2022-10-14
Housekeeping and tidying up

ns-rse:
* Rename `Image()` class option `type` to `image_type` as `type` is a core Python function.
* Rename `Image()` class option `pixel_to_nm_scaling_factor` to `pixel_to_nm_scaling` to be consistent across classes.
* Remove `channel` as an option to `Filter()` class.
* Remove `background` from `validate_config()`
* Remove `output_dir` from `Grains()` as it was only being used to make a directory and that is now done in `run_topostats.main()` / `process_scan()`.
* Tidied up docstring for `Grains()` class removing no longer used option `background`.

This also involved modifications to the validation schema and various tests.

(Noticed these as I was spending a little time updating the Notebook to work with the current codebase and started making the changes without creating an issue first).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -242,7 +242,6 @@ def random_grains(grains_config: dict, random_filters: Filters, tmp_path) -> Gra
         image=random_filters.images["zero_averaged_background"],
         filename="random",
         pixel_to_nm_scaling=0.5,
-        base_output_dir=Path(tmp_path),
         **grains_config,
     )
     grains.find_grains()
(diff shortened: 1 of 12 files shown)


SylviaWhittle (approved):
This is a wonderful set of changes, thank you @ns-rse. I've had a look and see no issues here.

ns-rse:
Thanks @SylviaWhittle most I should have caught when reviewing PRs but I don't always check everything, its one of the reasons I'm keen to get some linting applied (ideally `pylint` but that may take a fair bit of work).

Issue #302, opened 2022-10-05
Improve validation of configuration


ns-rse:
#299 addresses the need for the validation of our YAML input configuration and has been addressed by `INSERT`.

However, not all values are complete as valid ranges for each of the fields are not something @ns-rse can provide. Now that the framework is in place it should be refined and the purpose of this ticket is to record discussion of valid ranges for each of the fields prior to implementing.

Fields that are yet to have ranges defined...

| Field                                  | Type    | Current value in `config/example.yaml`                                                                                                                                                                                                                    |
|:---------------------------------------|:--------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `filter.amplify_level`                 | `float` | `1.0`                                                                                                                                                                                                                                                     |
| `filter.otsu_threshold_multiplier`     | `float` | `1.0`                                                                                                                                                                                                                                                     |
| `filter.threshold_absolute_lower`      | `float` | `-1.0`                                                                                                                                                                                                                                                    |
| `filter.threshold_absolute_upper`      | `float` | `1.0`                                                                                                                                                                                                                                                     |
| `filter.gaussian_size`                 | `float` | `0.5`                                                                                                                                                                                                                                                     |
| `grains.absolute_smallest_grain_size`  | `int`   | `100`                                                                                                                                                                                                                                                     |
| `grains.otsu_threshold_multiplier`     | `float` | `1.0`                                                                                                                                                                                                                                                     |
| `grains.otsu_threshold_multiplier`     | `float` | `1.0`                                                                                                                                                                                                                                                     |
| `grains.threshold_absolute_lower`      | `float` | `1.0`  **NB** Differs from `filter.threshold_absolute_lower`, was changed in [284](https://github.com/AFM-SPM/TopoStats/pull/284/files#diff-64a1933b775fbf7d3dc5f338c5b2fdb6d334aa1401172abc10f8bcb9184f6715) but not in `topostats/default_config.yaml`. |
| `grains.threshold_absolute_upper`      | `float` | `1.0`                                                                                                                                                                                                                                                     |
| `grains.absolute_area_threshold.upper` | `int`   |  `500-800`                                                                                                                                                                                                                                                         |
| `grains.absolute_area_threshold.lower` | `int`   |  `null-null`                                                                                                                                                                                                                                                         |
| `grains.background`                    | `float` |          `0.0`                                                                                                                                                                                                                                                 |
[...]

PR #328, opened 2022-10-17
Reordered the 'grains' section of the config file

### LINKED ISSUE #319: Reorder the 'grains' section of the config file
At the moment, the 'grains' section contains options for both height-based selection and area-based selection, and it is not very clear which one each line refers to, and this could cause confusion to users unfamiliar with the project.

Reorder the parameters in the 'grains' section of the config so that all height-based entries are in one place and all area-based entries are in another, with clearer commenting showing which one is which.

I considered creating nested subsections within the 'grains' section, but this would cause unecessary complications. 

**Additional context**

Jean-Du:
Reordered the 'grains' section of the config file so that it's clearer to users

closes #319

### DIFF
--- a/topostats/default_config.yaml
+++ b/topostats/default_config.yaml
@@ -17,16 +17,19 @@ filter:
   gaussian_mode: nearest
 grains:
   run: true # Options : true, false
-  absolute_smallest_grain_size: 100
-  threshold_method: otsu # Options : otsu, std_dev, absolute
+  # Thresholding by height
+  threshold_method: std_dev # Options : otsu, std_dev, absolute
   otsu_threshold_multiplier: 1.7
   threshold_std_dev: 1.0
   threshold_absolute_lower: 1.0
   threshold_absolute_upper: 1.0
-  absolute_area_threshold:
-    upper: [400,600] # above surface [Low, High] in nm^2 (also takes null)
-    lower: [null, null] # below surface [Low, High] in nm^2 (also takes null)
   direction: upper # Options: upper, lower, both (defines whether to look for grains above or below thresholds or both)
+  # Thresholding by area
+  absolute_smallest_grain_size: 100
+  absolute_area_threshold:
+    upper: [ 400,600 ] # above surface [Low, High] in nm^2 (also takes null)
+    lower: [ null, null ] # below surface [Low, High] in nm^2 (also takes null)
+
 grainstats:
   run: true # Options : true, false
   cropped_size: 40.0 # Length (in nm) of square cropped images (can take -1 for grain-sized box)
--- a/topostats/default_config.yaml
+++ b/topostats/default_config.yaml
@@ -18,7 +18,7 @@ filter:
 grains:
   run: true # Options : true, false
   # Thresholding by height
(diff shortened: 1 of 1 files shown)


topostats/default_config.yaml
```diff
@@ -17,16 +17,18 @@ filter:
   gaussian_mode: nearest
 grains:
   run: true # Options : true, false
-  absolute_smallest_grain_size: 100
-  threshold_method: otsu # Options : otsu, std_dev, absolute
+  # Thresholding by height
+  threshold_method: std_dev # Options : otsu, std_dev, absolute
```
ns-rse, line 21: I suspect the change in the default method will be the cause of tests failing.

ns-rse (changes requested):
Thanks @Jean-Du and sensible addition of comments to delimit sections :+1: 

A number of tests fail when I run them locally (I use the `pytest-xdist` to run them in parallel so can check faster than the GitHub Actions).

I think it will be because of the change in the `threshold_method` from `otsu` to `std_dev`. Whether this should be the default is a discussion to be had but it will require updating tests (or tweaking a fixture in `tests/conftest.py` so that the tests run with the current `otsu` value they are expecting.

ns-rse (approved):
Tests pass, it was my system that wasn't up-to-date, now it is they pass too, sorry for the mistake.

topostats/default_config.yaml
Jean-Du: I changed the default back to otsu before I saw your update. I'll keep it this way for now, because the current dev branch has otsu as default for both the 'grains' and the 'filter' sections, and it'd make more sense for this pull request to be only about improving readability.
However, I do recall that we have decided in previous meetings that standard deviation should be the default for both sections, and this has been the case in previous versions of the dev branch when example.yaml was still used. I suspect the default was (intentionally or unintentionally) reverted to otsu during the removal of the duplicate config file. Is it worth creating another issue?

Jean-Du: Also, if we are certain that standard deviation is the preferable default method, are there any tests that need to be changed to reflect that?

ns-rse: Thanks @Jean-Du 

Currently the regression tests under `tests/test_run_topostats.py` use the `threshold_method = "std_dev"` for both `Filter()` and `Grains()` processing.

The beauty of git is that we can go back through the [history of `default_config.yaml`]( https://github.com/AFM-SPM/TopoStats/commits/dev/topostats/default_config.yaml) and find where the change was introduced.

Doing so shows it was me who changed it, specifically so the tests could pass (see [this commit](https://github.com/AFM-SPM/TopoStats/commit/ec453ce3993dc78baab4b2acfd6bc447152a45e7#diff-969c222dca4d7238d7e79e8a6ac14b700a670ff1f85ec540ec4b7ed3d0ccb8e0)).

There is another way of doing this though and I shall now revert that in the configuration and modify the tests have used `otsu` in the past to use `std_dev`.

With regards to defaults across all values, not just threshold methods I think #302 which seeks to refine the schema would be a sufficient place to record this. It should probably be documented clearly why its the default and the implications of choosing other methods.

PR #331, opened 2022-10-17
Reverts default threshold_method to std_dev; updates tests

### LINKED ISSUE #330: Correct default threshold method and update tests.
#328 highlighted that the default `threshold_method` defined in the single configuration file we now have `topostats/default_config.yaml` had been switched to `otsu` in order to get tests to past.

Pending discussion of the schema (/default values) across all config parameters (see #302) the default value for `threshold_method` under both `grains` and `filters` should be reverted to `std_dev` and fixtures defined in `tests/conftest.py` updated to adjust these parameters accordingly so that tests pass.

ns-rse:
Closes #330

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,6 +35,7 @@ def default_config() -> Dict:
     config["plotting"]["plot_dict"] = yaml.safe_load(plotting_dictionary.read())
     config["filter"]["threshold_method"] = "otsu"
     config["grains"]["threshold_method"] = "otsu"
+    config["grains"]["otsu_threshold_multiplier"] = 1.7
     return config
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,6 +33,8 @@ def default_config() -> Dict:
     config = read_yaml(BASE_DIR / "topostats" / "default_config.yaml")
     plotting_dictionary = pkg_resources.open_text(topostats, "plotting_dictionary.yaml")
     config["plotting"]["plot_dict"] = yaml.safe_load(plotting_dictionary.read())
+    config["filter"]["threshold_method"] = "otsu"
+    config["grains"]["threshold_method"] = "otsu"
     return config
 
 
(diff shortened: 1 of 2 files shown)


Jean-Du:
During the commit that changed the default method from standard deviation to otsu, the default value for 'otsu_threshold_multiplier' in the grains section was also changed from 1.0 to 1.7. Was there a reason behind that, or does this need to be reverted to 1.0 too?

ns-rse:
> During the commit that changed the default method from standard deviation to otsu, the default value for 'otsu_threshold_multiplier' in the grains section was also changed from 1.0 to 1.7. Was there a reason behind that, or does this need to be reverted to 1.0 too?

`1.7` was the value the tests were developed with. If it should be `1.0` then I can revert that too and make the change in setting up the `@pytest.fixture` for `sample_config()` in the same way as I adjust the `threshold_method`.

I don't know what good values are (hence #302) and having tinkered a little don't see much objective difference in the images. Whether it has a quantitative impact on the resulting statistics would perhaps be the most important thing to assess and this may be something worth investigating on a wider range of samples than `minicircle.spm`.

SylviaWhittle:
> > During the commit that changed the default method from standard deviation to otsu, the default value for 'otsu_threshold_multiplier' in the grains section was also changed from 1.0 to 1.7. Was there a reason behind that, or does this need to be reverted to 1.0 too?
> 
> `1.7` was the value the tests were developed with. If it should be `1.0` then I can revert that too and make the change in setting up the `@pytest.fixture` for `sample_config()` in the same way as I adjust the `threshold_method`.
> 
> I don't know what good values are (hence #302) and having tinkered a little don't see much objective difference in the images. Whether it has a quantitative impact on the resulting statistics would perhaps be the most important thing to assess and this may be something worth investigating on a wider range of samples than `minicircle.spm`.

`1.7` is a good default for Otsu if I recall correctly

ns-rse:
> `1.7` is a good default for Otsu if I recall correctly

Ok, I was sort of unsure as I thought perhaps the removal of very small pieces via `absolute_smallest_grain_size` may alter the distribution (removing a heavy left-hand tail) meaning no scaling of the Otsu threshold was required. Perhaps this is something that some investigative data analysis might inform. :thinking: 

For now rather than continuing the discussion of defaults here could we perhaps continue it on #302 and/or next weeks meeting.

Jean-Du (approved):

Issue #334, opened 2022-10-20
Gaussian Size should be in px not nm


MaxGamill-Sheffield:
**Describe the bug**
Globally setting the `gaussian_size` parameter in nm means that when processing both large and smaller scans, smoothing can occur in smaller scans whereas pixelation can occur in larger images due to the gaussian size being smaller than the pixel length. 

As seen, the blurring by the gaussian also helps to remove noise caused by the high gain.

**To Reproduce**

Run topostats with 2 very different scan sizes and see the differences in blurring.

**Expected behavior**
Blurring factor should be consistent across images.

**Output**
See below for an example of `gaussian_size` = 0.5nm:
Large scan size (2µm)
![LIN_339bp_processed](https://user-images.githubusercontent.com/91465918/196923750-75472a68-2706-423d-8025-3b0d17c0a45e.png)
Small scan size (500nm)
![339bp_3_processed](https://user-images.githubusercontent.com/91465918/196924762-fc4bf694-79bd-4b2e-a0e3-13fdb5fb933b.png)

**(Not required if using Docker) Your computer configuration (please complete the following information):**
 - OS: e.g. windows, MacOS 12.4
 - topostats version: 404fa3c (HEAD -> dev, origin/dev) Merge pull request #328 from AFM-SPM/JeanDu/319-reorder-config-file
 - Python version: 3.10.5

PR #336, opened 2022-10-20
Changed `gaussian_size` to px not nm

MaxGamill-Sheffield:
This addresses #334 to change the config variable `gaussian_size` to us pixels instead of nm as units, which was causing pixelation in larger images.

Not many tests have been re-written due to identifying the pixel value that matches the current nm value.


topostats/default_config.yaml
```diff
@@ -13,7 +13,7 @@ filter:
   threshold_std_dev: 1.0
   threshold_absolute_lower: -1.0
   threshold_absolute_upper: 1.0
-  gaussian_size: 0.5
+  gaussian_size: 1.0121397464510862 # in px
```
ns-rse, line 16: I'm naive (or ignorant!), what is the reason for setting the `gaussian_size` to this value?

Could we perhaps document why this has been chosen as a default value somewhere? If not in the formal `docs/` then a comment here in the `default_config.yaml` so that when we get round to writing up more documentation we have a record of why its chosen.

Better still an explanation of what `gaussian_size` is actually doing would be useful, as I say I don't know what it means.

topostats/filters.py
```diff
@@ -255,11 +255,12 @@ def gaussian_filter(self, image: np.ndarray, **kwargs) -> np.array:
 
         """
         LOGGER.info(
-            f"[{self.filename}] : Applying Gaussian filter (mode : {self.gaussian_mode}; Gaussian blur (nm) : {self.gaussian_size})."
+            f"[{self.filename}] : Applying Gaussian filter (mode : {self.gaussian_mode}; Gaussian blur (px) : {self.gaussian_size})."
         )
+        print(f"Sigma = {self.gaussian_size / self.pixel_to_nm_scaling}")
```
ns-rse, line 260: Erroneous `print()` should be `LOGGER.info()` if its meant to be output during processing, otherwise remove.

ns-rse (commented):
Minor comments, mainly about documenting what is being done (I don't know what `gaussian_size` and its application to an image actually do, users would benefit from having an explanation available).

As an aside, if you start your Pull Request comment with...

```
Closes #334 
```

Then GitHub recognises this and should automatically close the issue given. See [Using keywords in issues and pull requests](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).

topostats/default_config.yaml
MaxGamill-Sheffield: This value is used as it is the same value as before, just in different units. i.e. a `gaussian_size` of 1.0121397464510862 px == 0.5 nm. This value means that the tests do not need to be changed, but it could just as easily assume any other value, but the tests would need to be modified.

I'll add the below to the `docs/configuration.md` file:
"The number of standard deviations to build the Gaussian kernel and thus affects the degree of blurring. See [skimage.filters.gaussian](https://scikit-image.org/docs/dev/api/skimage.filters.html#skimage.filters.gaussian) and `sigma` for more information."

Does that description cover it?

ns-rse: Great thanks. :+1: 

Will this vary depending on the resolution/size of the scan and/or instrument and/or the `pixel_to_nm_scaling` that is extracted from the scans?

ns-rse (approved):
Thanks for adding to the documentation, all good to merge. :+1:

topostats/default_config.yaml
MaxGamill-Sheffield: Nope. So the idea is that this is now set across all the images using a single value which does not change between images and thus doesn't use the `pixel_to_nm_scaling` or anything else.

PR #338, opened 2022-10-21
Use pathlib methods to resolve directory paths in get_out_path()

### LINKED ISSUE #337: Reconcile linting and get_out_path() function
@MaxGamill-Sheffield found that there is some conflict between `black` and `flake8` when linting code that uses a colon to slice a list...

```
[2022-10-20T17:06:38.781Z] > git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file - [775ms]
[2022-10-20T17:06:38.781Z] [WARNING] Unstaged files detected.
[INFO] Stashing unstaged files to /Users/Maxgamill/.cache/pre-commit/patch1666285598-17876.
trim trailing whitespace.................................................Passed
check docstring is first.................................................Passed
check for case conflicts.................................................Passed
fix end of files.........................................................Passed
fix requirements.txt.................................(no files to check)Skipped
check BOM - deprecated: use fix-byte-order-marker........................Passed
check yaml...........................................(no files to check)Skipped
black....................................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

topostats/utils.py:100:58: E203 whitespace before ':'
topostats/utils.py:286:21: F541 f-string is missing placeholders
```

This is caused because `black` places space around it, but `flake8` raises `E203` and complains about there being whitespace before the `:`.  This is discussed in [PEP8 Pet peeves](https://peps.python.org/pep-0008/#pet-peeves) which with specific reference to slicing states...

> However, in a slice the colon acts like a binary operator, and should have equal amounts on either side (treating it as the operator with the lowest priority). In an extended slice, both colons must have the same amount of spacing applied. Exception: when a slice parameter is omitted, the space is omitted:

Others have been caught by this before (see [Black #280](https://github.com/psf/black/issues/280).

There are a few options...

## Refactor `get_out_path()`

One possible solution is to refactor `get_out_path()` and a first try suggested by @ns-rse was the following...

```
def get_out_path(
    image_path: Path = None, base_dir: Path = None, output_dir: Path = None
) -> Path:
    """Adds the image path relative to the base directory to the output directory.

    Parameters
    ----------
    image_path: Path
        The path of the current image.
    base_dir: Path
        Directory to recursively search for files, if not specified the current directory is scanned.
    output_dir: Path
        The output directory specified in the configuration file.

    Returns
    -------
    Path
        The output path that mirrors the input path structure.
    """
    try:
        return output_dir / "/".join([i for i in image_path.parent.parts if i not in base_dir.parts])
    except TypeError:
        LOGGER.error("A string form of a Path has been passed to 'get_out_path()'")
        raise
```

This also needs the `.parent` removing from the call to `get_out_path()` made in `run_topostats.py` on line 163.

Apparently this fails tests if the path has a repeated directory name in it (e.g. `/tmp/something/tmp/images/` as it removes both `tmp`).

## Disable `E203`

We could disable `E203` either on the `black` side or `flake8` side via configuration of either package in `pyproject.toml`.

ns-rse:
Closes #337 

Handles taking the relative path of each image found within `base_dir` and creates the directory (minus any filename) under the `output_dir`.

The key is using the `.relative_to` method, but careful attention is required in case the `output_dir` is _not_ within the `base_dir` and is somewhere completely different.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -37,7 +37,7 @@ def test_get_out_path() -> None:
     output_dir = Path("output/here")
     out_path = get_out_path(image_path, base_dir, output_dir)
     assert isinstance(out_path, Path)
-    assert out_path == Path("output/here/path/test.spm")
+    assert out_path == Path("output/here/path/")
 
 
 def test_update_config(caplog) -> None:
--- a/topostats/run_topostats.py
+++ b/topostats/run_topostats.py
@@ -168,7 +168,7 @@ def process_scan(
         grain_out_path = _output_dir
     else:
         filter_out_path = Path(_output_dir) / image_path.stem / "filters"
-        grain_out_path = Path(_output_dir).parent / image_path.stem / "grains"
+        grain_out_path = Path(_output_dir) / image_path.stem / "grains"
         filter_out_path.mkdir(exist_ok=True, parents=True)
         Path.mkdir(_output_dir / image_path.stem / "grains" / "upper", parents=True, exist_ok=True)
         Path.mkdir(_output_dir / image_path.stem / "grains" / "lower", parents=True, exist_ok=True)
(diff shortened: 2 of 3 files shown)


SylviaWhittle (approved):
Works as intended 👍

PR #339, opened 2022-10-24
#137 file types/asd

### LINKED ISSUE #304: Support  ASD file format

MaxGamill-Sheffield:
Closes #304 

Due to the `.asd` filetype containing image frames, a few wider changes have been implemented along with the inclusion of asd file processing:
1. File input has been moved outside of `process_scan` and into `__main__()`.
2. File input class now takes a list of file path(s) and has been renamed `LoadScans`.
3. File inputs now produce a dictionary in the name of <LoadScans_obj>.img_dic containing keys of "images", "img_paths" and "px_2_nms". Values are a list of np.ndarray's, Path's and float's where each index corresponds to a matching image, path and px_2_nm value.
5. Process scan now takes an image, path and px_to_nm value as a zipped tuple, still allowing for multi-core processing

Other than this, a few tests have been modified and asd tests added apart from the px_to_nm.

Still need to think about how the px_2_nm function tests for all filetypes should be carried out - would it be best to have a fixture of the loaded scan of each to pass through to their appropriate functions?


MaxGamill-Sheffield:
I should add that due to the issues with `libasd` building on M1 macs, I have commented out all parts referencing `libasd` and any of its used functions. 

For now, if someone wants to process these types images, they will need to build `libasd` themselves (or pip install it) and then uncomment lines in `io.py` (and also in conftest and test_io.py for testing)

topostats/io.py
```diff
@@ -7,6 +7,7 @@
 
 from pySPM.Bruker import Bruker
 from afmformats import load_data
+#import libasd  # uncomment and import libasd to use asd formats
```
ns-rse, line 10: Need to uncomment.

ns-rse (changes requested):
Quite a few of the `asd` functions and tests are commented out.

Made a few comments and realise that the suggestion to use nested dictionaries would require a fair bit of work but I think the clarity it should bring may be worthwhile as I'm mindful that with sample Notebooks in the wings some people may want to tinker with the classes.

tests/test_io.py
```diff
@@ -72,3 +80,39 @@ def test_load_scan_load_jpk() -> None:
...
     """Test extraction of data from loaded JPK image."""
     assert True
+
+
+# uncomment below line and import libasd to test asd formats
+@pytest.mark.parametrize(
+    "load_scan_object, suffix, length, frame_no, image_shape, image_sum, filename, pixel_to_nm_scaling",
+    [
+        ("load_scan", ".spm", 1, 0, (1024, 1024), 30695369.188316286, "minicircle", 0.4940029296875),
+        #("load_scan_asd", ".asd", 64, 40, (256, 256), 5958870.556640625, "minicircles_frame_40", 1.953125),
```
ns-rse, line 90: Will need uncommenting.

topostats/io.py
```diff
@@ -149,16 +159,100 @@ def _load_jpk(self) -> None:
...
+                self.image, self.pixel_to_nm_scaling = self.load_jpk()
+            if self.suffix == ".ibw":
+                self.image, self.pixel_to_nm_scaling = self.load_ibw()
+                self.add_to_dic(self.image, self.img_path.with_name(self.filename), self.pixel_to_nm_scaling)
+            if self.suffix == ".asd":
+                self.img_path = str(self.img_path)
+                self.image, self.pixel_to_nm_scaling = self.load_asd()
+                for i, frame in enumerate(self.image):
+                    pathname = list(Path(self.img_path).parts)
+                    pathname[-1] = self.filename + "_frame_" + str(i)
```
ns-rse, line 250:
```suggestion
                    pathname[-1] = self.filename + f"_frame_{str(i)}"
```

topostats/io.py
```diff
@@ -149,16 +159,100 @@ def _load_jpk(self) -> None:
...
+    #def _asd_px_to_nm_scaling(self, scan: Union[libasd.Data2ch_v0, libasd.Data2ch_v1]) -> float:
+    #    """Extracts the pixel to nanometre scaling value from an .asd file object.
+
+    #    Parameters
+    #    ----------
+        scan: Union[libasd.Data2ch_v0, libasd.Data2ch_v1]
+    #        The object resulting from loading the filepath via the libasd library.
+    #    Returns
+    #    -------
+    #    px_to_m_scaling: float
```
ns-rse, line 214: Typo here and other places in this function `px_to_m_scaling` > `px_to_nm_scaling`.

However, to be consistent across all other code I think its better to stick with `pixel_to_nm_scaling` as used in `Filters` / `Grains` / `GrainStats` classes). 

They're all the same thing and should therefore have the same name, it makes it easier when using keywords and avoids the possibility that some may think they have different meanings/values.

topostats/io.py
```diff
@@ -149,16 +159,100 @@ def _load_jpk(self) -> None:
...
+    #    """
+    #    try:
+    #        scan = libasd.read_asd(self.img_path)
+    #        print(type(scan))
+    #        LOGGER.info(f"[{self.filename}] : Loaded image from : {self.img_path}")
+    #    except FileNotFoundError:
+    #        LOGGER.info(f"[{self.filename}] File not found : {self.img_path}")
+    #    try:
+    #        # libasd docs seem like there is only 2 channels i.e. help(scan)
+    #        # libasd.Header_v0 uses "type" and _v1 uses "kind"
```
ns-rse, line 179: Is it possible to determine which file version is being used?

If so it would perhaps be sensible to convert this to an `if: ... else:` with a view to possible changes in the future if/when a `_v2` comes along.

tests/conftest.py
```diff
@@ -265,22 +265,29 @@ def small_array_filters(small_array: np.ndarray, load_scan: LoadScan, filter_con
...
+    """Instantiate a LoadScans object."""
+    scan_loader = LoadScans([RESOURCES / "minicircle.spm"], **loading_config)
     return scan_loader
 
 
 @pytest.fixture
-def load_scan_data(load_scan: LoadScan) -> LoadScan:
-    """Instantiate a LoadScan object."""
+def load_scan_data(load_scan: LoadScans) -> LoadScans:
+    """Instantiate a LoadScans object."""
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Yep my bad, this is from an cp of the above test structure and for an older test before parameterising the `test_load_scan_get_data` test.

I have now fully removed this as a fixture as it is also not used - good spot.

topostats/io.py
```diff
@@ -149,16 +159,100 @@ def _load_jpk(self) -> None:
             LOGGER.error(f"[{self.filename}] : {exception}")
         return jpk
 
+    # uncomment below func and import libasd to use asd formats
+    #def load_asd(self) -> tuple:
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: This shouldn't fail as the asd tests shouldn't be run.

This PR is basically here for the switch-up in how files are loaded and put into process scan. The features of loading asd formats are there but should be used with caution as the the libasd library isn't in setup.cfg because it the binaries aren't built on MacOS and so fails on some OS's.

As such I have commented out all asd references.

tests/conftest.py
```diff
@@ -265,22 +265,29 @@ def small_array_filters(small_array: np.ndarray, load_scan: LoadScan, filter_con
...
     load_scan.get_data()
     return load_scan
 
 
+#@pytest.fixture
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Yes, the feature and tests are implemented but due to the issue raised by @SylviaWhittle with installing the libasd library, they have been commented out so that a user who knows that they are doing can uncomment and use the feature (or until we hear back from libasd),  but the file loading has been changed.

topostats/io.py
MaxGamill-Sheffield: Yep, I've added looking for the header types to the file

topostats/io.py
MaxGamill-Sheffield: It was in anticipation of going back to the SI units after this task was finished but you're right. Try and keep the things on one branch the same.

topostats/io.py
MaxGamill-Sheffield: I've tidied this whole bit up

topostats/io.py
```diff
@@ -149,16 +159,100 @@ def _load_jpk(self) -> None:
...
+            if self.suffix == ".asd":
+                self.img_path = str(self.img_path)
+                self.image, self.pixel_to_nm_scaling = self.load_asd()
+                for i, frame in enumerate(self.image):
+                    pathname = list(Path(self.img_path).parts)
+                    pathname[-1] = self.filename + "_frame_" + str(i)
+                    pathname = Path(*pathname)
+                    self.add_to_dic(frame, pathname, self.pixel_to_nm_scaling)
+
+    def add_to_dic(self, image, img_path, px_2_nm):
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: So I'd like to avoid adding `f"{frame}"` in this function due to not all filetypes containing frames and also unpacking the dict in `Filters`, so that the functions can be as easy to use as possible.

The dictionary has now been implemented with docstrings

topostats/io.py
```diff
@@ -149,16 +159,100 @@ def _load_jpk(self) -> None:
...
+        "Adds the image, image path and pixel to nanometre scaling value to the img_dic dictionary"
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Ah yep seems I forgot about the docstring - I've added this with the dictionary change

topostats/io.py
```diff
@@ -193,23 +201,128 @@ def _ibw_pixel_to_nm_scaling(self, scan) -> float:
...
-            self.image, self.pixel_to_nm_scaling = self.load_jpk()
-        if self.suffix == ".ibw":
-            self.image, self.pixel_to_nm_scaling = self.load_ibw()
+        """Method to extract image, filepath and pixel to nm scaling value, and append these to the
+        img_dic object.
+        """
+        for img_path in self.img_paths:
+            self.img_path = img_path
+            self.filename = img_path.stem
+            self.suffix = img_path.suffix
```
ns-rse, line 292: As with above comments about `self.suffix` because this is now within the `get_data` method and could possible change if there are files of mixed type within a given directory I think that here you can within this function do away with `self.suffix` and just use...

```suggestion
            suffix = img_path.suffix
```

(although all the `if suffix == "..."` also need updating (see above comment).

ns-rse (changes requested):
Few comments about `self.img_path` v `self.img_paths` in `load.spm()` and the fact that `self.suffix` can probably be done away with and just use `suffix` which is extracted within `get_data()` and could potentially change if there are files of mixed type in an input directory.

topostats/io.py
```diff
@@ -83,17 +84,24 @@ def __init__(
...
+        """Extract image and pixel to nm scaling from the Bruker .spm file.
+
+        Returns
+        -------
+        tuple(np.ndarray, float)
+            A tuple containing the image and its pixel to nanometre scaling value.
+        """
         LOGGER.info(f"Loading image from : {self.img_path}")
         try:
             scan = Bruker(self.img_path)
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: I agree, will change

topostats/io.py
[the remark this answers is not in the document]
MaxGamill-Sheffield: First off, this is only going into dev because I thought that's where we are putting everything. Please feel free to push to PyPI when @SylviaWhittle's flattening is complete, this is not an essential requirement for the PyPI release.


So, to update you on this, I've taken your feedback and found a workaround for the running and tests. I've added try and excepts around the `libasd` import, functions and tests. These now make the code run as expected on non-libasd files, but throws the following warning if the module cannot be found (as an inpromptu solution), and does not halt the processing of the non-libasd files:
```
[Wed, 26 Oct 2022 14:46:44] [WARNING ] [topostats] libasd module not installed. Ignore if '.asd' files are not being processed. 
 Otherwise install via 'pip install .[libasd]' or if on an M1/2 Mac, follow libasd installation guidelines for 'build yourself' here: https://github.com/ToruNiina/libasd
```

libasd files will also throw this warning along with the ```NameError: name 'libasd' is not defined``` error when processed, thus stopping their processing until the module is imported.

As you can see, install instructions for the libasd files are found in the error message (and if you'd like to suggest somewhere else please let me know), with a pip installable `pip install .[libasd]` option for those without M1/M2 Macs, and the libasd instructions there if they do.

Tests have been handled in a similar way where `ModuleNotFoundError`'s are asserted true, and otherwise `AssertionError`'s are raised (if any). 

My ask of you, @ns-rse here, is if there is a way to also to `pip install .[libasd]` in the GitHub tests as these all use non-M1/2 MacOS's. This would solve the issue you present and run these tests in the GitHub workflows (and pass locally, although possibly not properly run depending on wether libasd is installed).

I'm also not sure of the possible complications this would induce to pushing to PyPI but I'm all ears :)

topostats/io.py
```diff
@@ -83,17 +84,24 @@ def __init__(
...
-        self.file_path = self.img_path
-        self.filename = self.img_path.stem
-        self.suffix = self.img_path.suffix
+        self.filename = None
+        self.suffix = None
         self.image = None
         self.pixel_to_nm_scaling = None
+        self.img_dic = {"images": [], "img_paths": [], "px_2_nms": []}
 
     def load_spm(self) -> tuple:
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: I believe this is outdates and now set to an empty dict `{}` so I can easily add to it

topostats/io.py
MaxGamill-Sheffield: It was also being used in the error messages I've just realised but I've changed it to collect it from `self.img_path.suffix` instead so `self.suffix `can become `suffix`

MaxGamill-Sheffield:
@ns-rse I'll put my comment to one of your change requests here too so it can be more easily seen:

Please feel free to push to PyPI when @SylviaWhittle's flattening is complete, this is not an essential requirement for the PyPI release.

So, to update you on this, I've taken your feedback and found a workaround for the running and tests so all can be included in this PR. I've added try and excepts around the libasd import, functions and tests. These now make the code run as expected on non-libasd files, but throws the following warning if the module cannot be found (as an inpromptu solution), and does not halt the processing of the non-libasd files:

[Wed, 26 Oct 2022 14:46:44] [WARNING ] [topostats] libasd module not installed. Ignore if '.asd' files are not being processed. 
 Otherwise install via 'pip install .[libasd]' or if on an M1/2 Mac, follow libasd installation guidelines for 'build yourself' here: https://github.com/ToruNiina/libasd
libasd files will also throw this warning along with the NameError: name 'libasd' is not defined error when processed, thus stopping their processing until the module is imported.

As you can see, install instructions for the libasd files are found in the error message (and if you'd like to suggest somewhere else please let me know), with a pip installable pip install .[libasd] option for those without M1/M2 Macs, and the libasd instructions there if they do.

Tests have been handled in a similar way where ModuleNotFoundError's are asserted true, and otherwise AssertionError's are raised (if any).

My ask of you, @ns-rse here, is if there is a way to also to `pip install .[libasd]` in the GitHub tests as these all use non-M1/2 MacOS's. This would solve the issue you present and run these tests in the GitHub workflows (and pass locally, although possibly not properly run depending on wether libasd is installed).

I'm also not sure of the possible complications this would induce to pushing to PyPI but I'm all ears :)

tests/test_io.py
```diff
@@ -39,6 +41,43 @@ def test_load_scan_spm(load_scan: LoadScan) -> None:
...
+        load_scan_asd.filename = load_scan_asd.img_paths[0].stem
+        images, px_to_nm_scaling = load_scan_asd.load_asd()
+        image = images[40]  # some frames are just black so pick a middle one
+        assert len(images) == 64
+        assert isinstance(image, np.ndarray)
+        assert image.shape == (256, 256)
+        assert image.sum() == 5958870.556640625
+        assert isinstance(px_to_nm_scaling, float)
+        assert px_to_nm_scaling == 1.953125
+    except AssertionError:
```
ns-rse, line 58: There is an explicit approach to testing for exceptions with PyTest.

```
    with pytest.rauses(AssertionError):
        <action that raises the error>
```

It often requires splitting out assertion tests to their own function (at least that is the approach I take).

I'm not sure the `try: ...  except:` is necessary here though since `libasd` is installed in the GitHub Workflow and both locally on yours and others developments environments.

tests/test_io.py
```diff
@@ -47,77 +86,50 @@ def test_load_scan_spm(load_scan: LoadScan) -> None:
...
-def test_load_scan_load_jpk() -> None:
-    """Test loading of JPK image."""
-    assert True
-
-
-def test_load_scan_extract_jpk() -> None:
-    """Test extraction of data from loaded JPK image."""
-    assert True
+    # temporary try statement until libasd has correct binaries
+    try:
```
ns-rse, line 121: As above I don't think the `try: ... except: ...` is required as it should pass everywhere people are running.

ns-rse (changes requested):
Great work as always @MaxGamill-Sheffield 

I've a general comment on the approach taken to `libasd` handling will require some restructuring of configuration and documentation.

Until we have a reliable method of providing binaries for `libasd` we have to help people get things setup.

As such I think this should be done _before_ people get to the stage of trying to analyse their files and instead during the installation stage, that way once they've installed TopoStats and if on ARM OSX systems explicitly cloned, compiled and installed `libasd` then the processing of files should "Just Work(tm)".

To which end the installation instructions should have a message along the lines of.

> **WARNING** 
> If you are using OSX on a system with an ARM processor (Apple Mac M1 or M2 computers), you will encounter an error message about `libasd` not being available for your system as at present there is no binary available.
>
> ```
> ERROR: Could not find a version that satisfies the requirement something (from versions: none)
> ERROR: No matching distribution found for libasd
> ```
>
> Please follow the instructions to [build libasd yourself](https://github.com/ToruNiina/libasd#building-by-yourself). You will have to install [cmake](https://cmake.org/install/) first if you do not already have it (binaries for cmake are also available via [brew](https://formulae.brew.sh/formula/cmake)). If you encounter issues doing this please do not hesitate to ask questions in the [Discussions](https://github.com/AFM-SPM/TopoStats/discussions) on GitHub.

I don't think its worth trying to explain the instructions here when they are detailed on the package page although it may be as simple as...

```
brew install cmake
git clone --recursive https://github.com/ToruNiina/libasd.git
pip install --user ./libasd
```

...but then that assumes all users will have [brew](https://brew.sh/) installed (I'd hazard a guess most won't).

This then allows you to do away with the extra requires for `libasd` and put it in the main requirements and the GitHub Workflow can have the additional line removed.

It would be sensible to leave the error messages in place on the off (likely?) chance that people somehow get that far and haven't installed `libasd` although the wording could be adjusted to point them to the installation instructions of TopoStats again (that way we keep instructions in one place and don't have to update them in the error message).

Whilst this requires the user to do a bit more work up-front I think its useful as it catches the problem earlier.

Another major advantage is that it doesn't inconvenience users who _aren't_ using ARM OSX systems, they should find that `libasd` installs  from PyPI pulling the dependencies in and they don't need to think about having to do an additional `pip install topostats.[libasd]`  which now I come to think about it I don't think is even possible and so we almost _have_ to take this approach for the time being if we are to publish to PyPI.

topostats/io.py
```diff
@@ -6,15 +6,20 @@
...
 from ruamel.yaml.main import round_trip_load as yaml_load, round_trip_dump as yaml_dump
 from topostats.logs.logs import LOGGER_NAME
 
 LOGGER = logging.getLogger(LOGGER_NAME)
 
+try:
+    import libasd
+except ModuleNotFoundError:
+    LOGGER.warning(
+        "libasd module not installed. Ignore if '.asd' files are not being processed. \n Otherwise install via 'pip install .[libasd]' or if on an M1/2 Mac, follow libasd installation guidelines for 'build yourself' here: https://github.com/ToruNiina/libasd"
```
ns-rse, line 20: Need to change the `pip install .[libasd]` since that is for installing from a locally cloned repository. Can also link directly to the build instructions of libasd

```suggestion
        "libasd module not installed. Ignore if '.asd' files are not being processed. \n Otherwise install via 'pip install topostats[libasd]' or if on an M1/2 Mac, follow libasd installation guidelines for 'build yourself' here: https://github.com/ToruNiina/libasd#building-by-yourself"
```

tests/test_io.py
MaxGamill-Sheffield: From what I understand from here https://docs.pytest.org/en/7.1.x/how-to/assert.html, it seems as though `pytest.raises()` passes the test if the error is raised. However, this is not what I want.

Here I am not testing that the produced outcome of the function is to raise and exception/error. Instead these tests should be ignored when `libasd` is not installed, and run when it is. So the try's here check and run that.

Although, you have raised a good point that maybe we can include tests for errors e.g. when the channel is wrong to check the error messages.

MaxGamill-Sheffield:
Seems like there were some bits missing/re-written in the merge so hopefully I caught them all. Will re-check after tests complete/fail

tests/test_io.py
MaxGamill-Sheffield: outdated with libasd import now available

tests/test_io.py
MaxGamill-Sheffield: outdated with libasd import now available

topostats/io.py
MaxGamill-Sheffield: removed due to libasd import now available

ns-rse (changes requested):
Looking at this again, will investigate the `libasd` issue on M$-Win.

Minor suggestion `dic > dict` in a function name.

ns-rse:
We still have failures on Windows because the binaries on PyPI are problematic (the don't include the `.so` compiled C++ code required to run properly).

This is in some manner related to the version bump and I've created a Pull Request to address this on `libasd` (see [here](https://github.com/ToruNiina/libasd/pull/13)). Hopefully once merged this will update PyPI with `libasd-1.5.6` which will have valid Windows wheels and our test suite will run correctly.

ns-rse:
Superseded by #675

PR #341, opened 2022-10-24
Tidy up exceptions in get_out_path(); parametrize get_out_path test & exceptions

ns-rse:
Thanks to @MaxGamill-Sheffield eagle eye on spotting the `TypeError` exception would never get raised I've added parameterized tests for `get_out_path()` which highlighted that if `image_path` is relative rather than absolute (unlikely to occur but never know if the code is being used interactively in a Notebook) then an `AttributeError` _is_ raised.

I also learnt that `Path()` methods automatically convert string arguments to `Path()` objects themselves and you can have `"some_dir" / Path("another_dir")` previously I'd only ever done it `Path("another_dir") / "some_dir"`.


ns-rse:
The tests are failing because `pytest-7.2.0` removed `py` as a dependency and `pytest-regtest-1.5.0` requires `py`.  This has been addressed in a few other Pull Requests that are out for review and shouldn't block merging.

SylviaWhittle (approved):

PR #342, opened 2022-10-24
Regression tests for refactoring Plotting.py

ns-rse:
This introduces a series of regression tests for the plots produced by @Jean-Du `Plotting.py`.

I've moved this to `topostats/plotting.py` and had to add a few lines here and there to make them work with tests but its mostly unchanged.

The tests provide a reference of the plots that are required so that as refactoring proceeds I know I'm not changing the required output by accident.  If changes are required to the plots they will need the tests adjusting.

Currently `topostats/plotting.py` fails linting and as do a number of the files for DNA Tracing. I've configured `flake8` and `black` to ignore these and at the same time addressed a bunch of linting errors across a host of files that were being reported. A few have been explicitly disabled on the offending lines (although I know at least one is addressed by #338 so I may get a merge conflict).


ns-rse:
For some reason I had to recreate the target images as `pytest-mpl` was saying there was a difference that exceeded the threshold.

I can't see it myself...

## Baseline

image

## Result

image

## Failed Diff

image


:shrug:

ns-rse:
Couldn't get one skeleton plot test to work (succeeds locally just not here). I'll resolve that when I'm next working on refactoring DNATracing.

ns-rse:
I was thinking about the structure of `get_out_path()` and realised there is an alternative approach of checking for image suffixes first and then using `try:...except:`...

```python
    # If image_path has a suffix we need to strip the filename
    if image_path.suffix:
        # If image_path is relative and doesn't include base_dir then a ValueError is raised in which case we just want
        # to append the image_path to the output_dir
        try:
            return output_dir / image_path.parent.relative_to(base_dir)
        except ValueError:
            return output_dir / image_path.parent
        except AttributeError:
            LOGGER.error("A string form of a Path has been passed to 'get_out_path()' for image_path")
            raise
    else:
        # Ditto if we have just a plain directory path without a suffix
        try:
            return output_dir / image_path.relative_to(base_dir)
        except ValueError:
            return output_dir / image_path
        except AttributeError:
            LOGGER.error("A string form of a Path has been passed to 'get_out_path()' for image_path")
            raise

```

The above works fine and passes the tests too but is more lines of code and has duplication (the `except AttributeError:` section) so I've stuck with the logic used originally of `try:...except:` capturing non-relative paths first and then checking for file suffixes.

tests/test_plotting.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+# Not testing, path is contingent on system and changes with each test
+# def test_pathman(tmpdir) -> None:
+#     """Regression test for pathman()."""
+#     plotname = pathman(str(tmpdir))
+#     assert plotname == "/tmp/"
+
+
+def test_labelunitconversion() -> None:
```
SylviaWhittle, line 49: Is this something to be completed later, or is it functional currently?

ns-rse: Well spotted, just a place holder for now.

PR #346, opened 2022-10-26
Add Python 3.11.0 to tests workflow matrix

### LINKED ISSUE #345: Support Python 3.11 in CI/CD pipeline
[Python 3.11 was released 2022-10-24](https://www.python.org/downloads/release/python-3110/) and it would be prudent to include in the CI/CD pipeline.

File to change is `.github/workflows/tests.yaml` and its line 15 where '3.11' needs adding.

```yaml
        python-version: ['3.8', '3.9', '3.10', '3.11']
```

ns-rse:
Closes #345

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ['3.8', '3.9', '3.10']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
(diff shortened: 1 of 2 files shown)


bobturneruk:
Something up with `pytest-regtest`?

ns-rse:
> Something up with `pytest-regtest`?

Yes, @MaxGamill-Sheffield and I have just been discussing this, seems `py` (which is in maintenance mode) isn't being pulled in as a dependency. Just trying explicitly installing it.

I saw this yesterday a few times and then the problem went away of its own accord after restarting the tests a few times. :shrug:

ns-rse:
Looks like `h5py` package has issues building under Python 3.11.0 going to close this for now and try in a few weeks when all packages are available under PyPI against 3.11.0.

PR #348, opened 2022-10-31
Added contingency for spm px_2_nm

MaxGamill-Sheffield:
I realised that when working with some files/data, the `pySPM.Bruker().pxs` gave a value of [[0.0, "um"],[0.0, "um"]].

This was because the metadata was lost/not included and thus the value for the px_2_nm scaling was 0, causing issues throughout TopoStats.

Where gwyddion defaults to a pixel size of 1m, I have implemented a quick fix which similarly defaults the pixel size to 1nm if this field is 0.


ns-rse (approved):
Good catch and a simple solution

PR #368, opened 2022-11-18
311 | Adding to documentation pending release; tidying config

ns-rse:
As we look to release to PyPI I've made some revisions to the documentation and tidied up a few things here and there.

* Sets the `base_dir` in `topostats/default_config.yaml` to be the current directory `run_topostats` is invoked from.
* Documentation gets package version automatically
* Installing now covers PyPI and GitHub cloning
* Introduction page tidied up a bit.
* Links to `default_config.yaml` under Usage corrected to `main` branch.
* Config pyproject.toml twekaed
* Some more linting with Black across files.


docs/installation.md
```diff
@@ -29,32 +29,71 @@ You are now ready to clone and install TopoStats.
...
+selecting a destination to clone to.
+
+
+#### Installing TopoStats from the Cloned Repository
+
+Once cloned you will have to open a Terminal and navigate to the directory you cloned and _after_ activating your
+virtual environment install TopoStats with the following.
+
 ``` bash
+cd /path/to/where/topostats/was/cloned/TopoStats
```
SylviaWhittle, line 85: I have found it very hard to indicate to users where to insert things and not just copy / paste the code line. I know of no perfect way to do it and have no solid suggestions here, only that perhaps users will not be able to tell that they should keep the `/TopoStats` at the end?

SylviaWhittle (approved):
Minor comments, feel free to ignore / push back 😃

docs/installation.md
ns-rse: A perennial challenge, can't think of a way around it either.

PR #387, opened 2022-12-03
Maxgamill sheffield/183 dnatracing refactor

MaxGamill-Sheffield:
This PR centres around refactoring `dnatracing.py` upto and including the `disrodered_traces` function. This means this PR centers around organising the skeletonisation code from `tracingfuncs`, removing the unused code, adding docstrings and putting it into `skeletonize.py`. 

The regtests remain unaltered and pass so things are working as expected, but no additional tests have been made for Joe's skeletonisation code as I'm unsure what some of them do, there is a lot, and the class is meant to act like Zhangs method (but produces different results). - If anyone can offer assistance here, that would be very helpful!

I also altered some of the utility functions in `dnatracing.py` to be more understandable. However, I do have an issue with a bit of the code - dilation requires each grain to have their own binary image (to prevent close grains touching), but this also dramatically slows processing times and workflow as the whole image can't be skeletonised/pruned at once (which all the skeletonisation methods can do).


topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        np.ndarray
+            The mask array reduce to a single pixel thickness
+
+        Notes
+        -----
+        This method is based on Zhang's method but produces different results
+        (less branches but slightly less accurate).
+        """
+        # image height bits don't seem to be used but are there?
+        return joeSkeleonize(image, mask).doSkeletonising()
```
ns-rse, line 169: Typo here and in the class definition...

```suggestion
        return joeSkeletonize(image, mask).doSkeletonising()
```

MaxGamill-Sheffield:
Added tests for dnatracing functions (still none for Joe's skeleton funcs)

topostats/tracing/dnatracing.py
```diff
@@ -157,65 +159,41 @@ def _crop_array(array: np.ndarray, bounding_box: Tuple) -> np.ndarray:
...
+        Parameters
+        ----------
+        traces: dict
+            A dictionary of the molecule_number and points within the skeleton.
+        """
 
         self.num_circular = 0
         self.num_linear = 0
 
         for dna_num in sorted(traces.keys()):
```
ns-rse, line 190: As a general comment I'm not a fan of looping over dictionary/lists within a method.

A function/method should do one thing. If it then needs to be applied to a dictionary or list kt can be.

To which end the `for dna_num in sorted(traces.keys())` should be moved out of this method and it can be made a `@staticmethod` that takes a single `np.ndarray` and line 93 in `trace_dna()` moved into the preceeding loop.

The same is true of all of the methods outside of that loop.

topostats/tracing/dnatracing.py
```diff
@@ -79,12 +79,18 @@ def __init__(
...
-        self.gaussian_filter()
+        self.grains = self.binary_img_to_dict(self.grains_orig)
+        # What purpose does binary dilation serve here? Name suggests some sort of smoothing around the edges and
+        # the resulting array is used as a mask during the skeletonising process.
+        self.dilate_grains()
+        for grain_num, grain in self.grains.items():
+            skeleton = getSkeleton(self.gauss_image, grain).get_skeleton(self.skeletonisation_method)
+            pruned_skeleton = pruneSkeleton(self.gauss_image, skeleton).prune_skeleton(self.pruning_method)
+            self.skeleton_dict[grain_num] = pruned_skeleton #if pruned_skeleton[pruned_skeleton==1].size < 10 else pass
         self.get_disordered_trace()
```
ns-rse, line 90: This and the subsequent calls should be refactored to work with a single molecule/grain and moved into the `for:` loop on line 86.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+            List of [x, y] coordinate positions
+
+        Returns
+        -------
+        bool
+            Returns T/F depending if the surrounding points have met the criteria
+            of the binary thin a, b returncount, csharp and dsharp checks below.
+        """
+
+        self.p7, self.p8, self.p9, self.p6, self.p2, self.p5, self.p4, self.p3 = self.getLocalPixelsBinary(
```
ns-rse, line 280: Attributes should be defined within `__init__()` otherwise you see an error when linting [`W0201` `attribute-defined-outside-init`](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/attribute-defined-outside-init.html).

As there are a number of points here you also likely to run into [`R0902 too many instance attributes`](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/too-many-instance-attributes.html). To minimise the number of attributes you can initialise under `__init__` a dictionary to hold the values.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        )
+        # Add in generic code here to protect high points from being deleted
+        return (
+            self._binaryThinCheck_a()
+            and self._binaryThinCheck_b_returncount() == 1
+            and self._binaryThinCheck_csharp()
+            and self._binaryThinCheck_dsharp()
+        )
+
+    def _binaryThinCheck_a(self):
```
ns-rse, line 291: As we're trying to adhere to PEP8 [Class names](https://pep8.org/#class-names) should use CamelCase whilst [function names](https://pep8.org/#function-names) should be lower case (and typically snake_case).

Pylint warns us of this issue ([`C0103` - `invalid name`](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html).

Further "`ThinCheck_a`" doesn't explain what the check is very well. Would be useful to have a descriptive name, perhaps something like `_check_point_on_edge_of_grain()` and describe why the sum of these points being between 2 and 6 qualifies as such.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        Condition A protects the endpoints (which will be > 2)
+
+        Returns
+        -------
+        bool
+            if point lies on edge of graph and isn't an endpoint.
+        """
+        return 2 <= self.p2 + self.p3 + self.p4 + self.p5 + self.p6 + self.p7 + self.p8 + self.p9 <= 6
+
+    def _binaryThinCheck_b_returncount(self):
```
ns-rse, line 302: As we're trying to adhere to PEP8 [Class names](https://pep8.org/#class-names) should use CamelCase whilst [function names](https://pep8.org/#function-names) should be lower case (and typically snake_case).

Pylint warns us of this issue ([`C0103` - `invalid name`](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/invalid-name.html).

Further "`_b_`" isn't really informative about what the condition is, if we could name this more descriptively that would be informative.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+                [self.p6, self.p7] == [0, 1],
+                [self.p7, self.p8] == [0, 1],
+                [self.p8, self.p9] == [0, 1],
+                [self.p9, self.p2] == [0, 1],
+            ]
+        )
+
+        return count
+
+    def _binaryThinCheck_c(self):
```
ns-rse, line 319: Can we describe this check in any greater detail and why its used?

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        """Check if p2, p4 or p6 is 0 - seems very specific
+
+        Returns
+        -------
+        bool
+            if p2, p4 or p6 is 0.
+        """
+        return self.p2 * self.p4 * self.p6 == 0
+
+    def _binaryThinCheck_d(self):
```
ns-rse, line 329: Ditto why this check?

Also this and some of the others might be combined. The products being calculated will be true if any one of the values is `0` (since `0 * anything = 0` and so a simpler check might be to ensure `p2` / `p4` / `p6` / `p8` are all non-zero as the tests are as they stand not mutually exclusive.

A better understanding of why this is required or considered a useful check would help, maybe @JoeBeton could help, or perhaps a close reading of the cited paper.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        -------
+        Callable
+            Returns the function appropriate for the required skeletonizing method.
+        """
+        if method == "joe":
+            return self._prune_joe(self.image, self.skeleton)
+        raise ValueError(method)
+
+    @staticmethod
+    def _prune_joe(image: np.ndarray, skeleton: np.ndarray) -> np.ndarray:
```
ns-rse, line 492: I think a more descriptive name here would be appropriate. `joe` doesn't mean much to someone coming to this with fresh eyes.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+
+        Returns
+        -------
+        np.ndarray
+            The skeleton with spurious branching artefacts removed.
+        """
+        return joePrune(image, skeleton).prune_all_skeletons()
+
+
+class joePrune:
```
ns-rse, line 510: I think a more descriptive name here would be appropriate. `joe` doesn't mean much to someone coming to this with fresh eyes.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        to make assessments about the overall shape/structure of traces"""
+
+        remaining_coordinates = np.argwhere(self.mask).tolist()
+
+        for x, y in remaining_coordinates:
+            self.p7, self.p8, self.p9, self.p6, self.p2, self.p5, self.p4, self.p3 = self.getLocalPixelsBinary(
+                self.mask, x, y
+            )
+
+            # Checks for case 1 pixels
```
ns-rse, line 376: I suspect this is copied verbatim but we could perhaps describe what `case 1` and `case 2` are from the Zhang paper.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        """Assess if local area has 4-connectivity.
+
+        Returns
+        -------
+        bool
+            T/F if any neighbours of the 4-connections have a near pixel.
+        """
+        return 1 in (self.p2 * self.p4, self.p4 * self.p6, self.p6 * self.p8, self.p8 * self.p2)
+
+    def _binaryFinalThinCheck_b(self):
```
ns-rse, line 393:
```suggestion
    def _binaryFinalThinCheck_b(self) -> bool:
```

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+        """Assess if local area 4-connectivity is connected to multiple branches.
+
+        Returns
+        -------
+        bool
+            T/F if any neighbours of the 4-connections have a near pixel.
```
ns-rse, line 399:
```suggestion
            Logical indicator of whether any neighbours of the 4-connections have a near pixel.
```

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+            A y coordinate within the binary map.
+
+        Returns
+        -------
+        np.ndarray
+            A flattened 8-long array describing the values in the binary map
+            around the x,y point
+        """
+        local_pixels = binary_map[x - 1 : x + 2, y - 1 : y + 2].flatten()
+        # [[p7, p8, p9],    [[0,1,2],
```
ns-rse, line 429: These lines can be removed if no longer needed.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,641 @@
...
+
+            # The branches are typically short so if a branch is longer than
+            #  0.15 * total points, its assumed to be part of the real data
+            max_branch_length = int(len(coordinates) * 0.15)
+
+            # first check to find all the end coordinates in the trace
+            potential_branch_ends = self._findBranchEnds(coordinates)
+
+            # Now check if its a branch - and if it is delete it
+            for x_b, y_b in potential_branch_ends:
```
ns-rse, line 581: Presumably `x_b` and `y_b` are co-ordinates of the branch. We can avoid `invalid-name` linting errors by instead using `x_branch` and `y_branch`.

topostats/tracing/tracingfuncs.py
```diff
@@ -816,6 +227,9 @@ def loopedLinearTrace():
 class genTracingFuncs:
```
ns-rse, line 227: Don't really need this to be a class if its only got a series `@staticmethod` it can just be a series of functions.

Might perhaps be better to refactor as...

```
class genTracingFuncs():
    """Class for generating tracing functions."""
    def __init__(self, binary_map: np.ndarray, x: int, y:int) -> None:
    """Initialise the class."""
    self.binary_map = binary_map
    self.x = x
    self.y = y

    ...
```

tests/tracing/test_dnatracing.py
```diff
@@ -8,105 +8,85 @@
...
-                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-                [0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0],
-            ]
-        ),
+DILTATE_TARGETS = {
+    1: np.asarray(
```
ns-rse, line 14: As this is a dictionary could we use more informative keys to describe what the arrays are? Otherwise could just use a list.

topostats/default_config.yaml
```diff
@@ -33,14 +33,16 @@ grains:
   # Thresholding by area
   absolute_smallest_grain_size: 100
   absolute_area_threshold:
-    upper: [ 500, 800 ] # above surface [Low, High] in nm^2 (also takes null)
+    upper: [ 500, null ] # above surface [Low, High] in nm^2 (also takes null)
```
ns-rse, line 36: Surprised this doesn't cause problems with some of the `minicircle` tests. I've certainly been bitten by having the wrong values here when testing my code.

I think generally we need to decide on a fixed set of values for the `default_config.yaml` and stick with them. If tests require them to be different values then the values in the config dictionaries can be changed in the tests.

I did create an issue for discussing and recording what the default values should be but can't find it right now.

ns-rse: Just to update I found the issue I was thinking of, its #302.

topostats/tracing/dnatracing.py
MaxGamill-Sheffield: My aim with this PR was only to refactor the dnatracing and skeletonisation methods necessary to continue with the knots and cats work which Alice is presenting soon. Hence why I've only added the docstring here and not modified/refactored any of the code.

I'd expect this to be done in a future PR where the traces are refactored as I just don't have the time right now. This can then be merged alongside the skeleton refactor (this) and the curvature refactor (the one you're currently working on) to refactor all of dnatracing bit by bit.

topostats/tracing/dnatracing.py
MaxGamill-Sheffield: I'm actually hoping to remove the dependancy for the single molecule dictionary and skeletonise the whole grain image at once, hence why I've left this like this. (all of the skeletonisation funcs can work do skeletonise a whole image of multiple skeletons).

This is due to the dilate function which means that grains may overlap so it is necessary to have each grain as a separate image and iterate over the single grain images. However, the dilation may close interior holes which can make the skeletons not follow their intended path.

What do you think @ns-rse?

topostats/tracing/skeletonize.py
MaxGamill-Sheffield: It's a description to show what each of the positions: p2, p3... etc mean. I've moved this into the docstring as I think it helps with understanding

topostats/tracing/skeletonize.py
MaxGamill-Sheffield: These are described in `final_skeletonisation_iteration`

topostats/default_config.yaml
MaxGamill-Sheffield: Yep this is my bad. There was no issue because the larger of the thresholds while it does pick up a large grain, it resides on the edge so is removed. So, more of a "lucky" mistake

tests/tracing/test_dnatracing.py
MaxGamill-Sheffield: This just mimics how the arrays are fed from the dilate function. It also makes it easier to compare what we expect from grain 1, 2, 3 etc from the GRAINS array in the conftest. i.e. dnatracing_dilate.grains is a dict of grain_number, grain_array, which we can directly compare to the grain number and grain array from the DILATE_TRAGETS

topostats/tracing/tracingfuncs.py
MaxGamill-Sheffield: I'd first say that I like how the class groups these functions and thus enhances readability. I also don't think your suggested refactor would work as some of the functions take points etc.

This might be one for when this entire part (the rest of dnatracing) has been refactored

MaxGamill-Sheffield:
@ns-rse just a little update to address some of your comments:

1. This PR is just a refactoring of the skeletonisation methods and dnatracing workflow upto and including this point. This means that while I may have added some docstrings, I will not be doing the refactoring of the entirety of dnatracing (for now). This is to help me focus on the knots and cats work which requires this.
 
2. I'm also not entirely sure what each of the checks looks for, and what puzzles me more is that these checks and the "joe" method in general derive from the Zhang paper but produces different results from the scikit image Zhang method. hence why I've left the method as "joe". I'm definitely game to change this but I'm not sure what to.

3. Alice and I were hoping that you would undertake the task of legacy tests after refactoring Jean's curvature stuff. Or if @SylviaWhittle has the time (although I believe she's taking time off soon).

In the meantime, I'll make a fork of this branch so I can continue working on the knots and cats stuff but I hope that other than the: tests, "joe" method and descriptions of some funcs in the "joe" method, there shouldn't be anything fundamentally wrong with this in terms of the way the pipeline works. Please let me know if there are

ns-rse:
@MaxGamill-Sheffield Thanks for getting back.

Based on what you've described I don't think this is ready to merge into `main` then as we should be avoiding leaving [technical debt](https://www.atlassian.com/agile/software-development/technical-debt) in the main branch. Typically it takes a long time to go back and address this and in the mean time features get built on top and the task becomes bigger. If the API in terms of the skeletons that are returned is stable then refactoring can wait and working off of a fork for the cats and knots is more appropriate.

It is concerning that the existing method of skeletonisation that implements the Zhang paper (that Joe implemented) doesn't give the same results and it would I think be worth understanding why this is (the scikit-image implementation has a [suite of tests](https://github.com/scikit-image/scikit-image/blob/a91f4f8bab92c4eede48e4949805c748f436c4de/skimage/morphology/tests/test_skeletonize.py) in place).

topostats/tracing/dnatracing.py
ns-rse: Not sure, that confuses me as you first say a single image with multiple skeletons can be processed but then that dilation can cause grains to overlap and so they then need iterating over individually.

`Grains` returns individual grains (as a dictionary from memory). These grains are at this stage the single unit of analysis for tracing (each is traced on its own) and any subsequent analysis rather than the image as a whole. Thus having `DNATracing` class handle a single grain/molecule seems most appropriate. It could then be inherited by another class (`TraceAll` for want of a better name) that handles all grains from an image and basically just loops over the dictionary returned by `Grains`.

MaxGamill-Sheffield: Sorry for the confusion. Yep, currently all molecules are skeletonised, pruned and traced separately, but I've made it so all the skeletonisation and pruning functions can work on the whole image, where I think this works best and quickest.

I could then move the subsequent functions within the for loop to do the tracing if you'd like? But for now due to the dilation, I have to keep the skeletonisation and pruning in the for loop.

MaxGamill-Sheffield:
> @MaxGamill-Sheffield Thanks for getting back.
> 
> Based on what you've described I don't think this is ready to merge into `main` then as we should be avoiding leaving [technical debt](https://www.atlassian.com/agile/software-development/technical-debt) in the main branch. Typically it takes a long time to go back and address this and in the mean time features get built on top and the task becomes bigger. If the API in terms of the skeletons that are returned is stable then refactoring can wait and working off of a fork for the cats and knots is more appropriate.

Thanks for checking this out @ns-rse,
This draft PR is kind of how I pictured the process going forward without an "untested dev branch". I.e. the features are finished but lacking tests to be handed over for completion. The draft PR acts as a hand-off sign, where I can still make the feature amendments necessary to reduce the test creators workload and avoid the branch being forgotten about.

topostats/tracing/skeletonize.py
```diff
@@ -2,78 +2,651 @@
...
+        Condition A protects the endpoints (which will be > 2)
+
+        Returns
+        -------
+        bool
+            if point lies on edge of graph and isn't an endpoint.
+        """
+        return 2 <= self.p2 + self.p3 + self.p4 + self.p5 + self.p6 + self.p7 + self.p8 + self.p9 <= 6
+
+    def _binary_thin_check_b_returncount(self) -> bool:
```
ns-rse, line 311: I think this will return an `int` in the range of 1 to 8 rather than a `bool`.

MaxGamill-Sheffield, line 311:
```suggestion
    def _binary_thin_check_b_returncount(self) -> int:
```

MaxGamill-Sheffield:
Closing PR until tests in #412 completed

PR #390, opened 2022-12-06
Sylvia whittle/remove scars

### LINKED ISSUE #154: Add grain removal based on height (identify and isolate scars)
Removing anomalously high-up grains would be useful as it would remove artefacts such as scars and blobs that are not wanted and would mess with the scaling of the data.

A system that detects grains that are 2 standard deviations from the mean grain height and removes them.

### LINKED ISSUE #202: Remove scarring

Some images have unwanted artifacts known as "scarring" on them. Being able to remove these during processing would provide cleaner images for subsequent analysis. These are typically always straight lines which aids in their identification for removal.

Removing of scarring artifacts during image processing. There are tools available in [scikit-image](https://scikit-image.org). Specifically the [Straight line Hough transform](https://scikit-image.org/docs/stable/auto_examples/edges/plot_line_hough_transform.html)

> The Hough transform in its simplest form is a method to detect straight lines [1](https://scikit-image.org/docs/stable/auto_examples/edges/plot_line_hough_transform.html#id3).

Alternatives are available in OpenCV (see [StackOverflow Thread](https://stackoverflow.com/questions/33541551/hough-lines-in-opencv-python)).

N/A

SylviaWhittle:
Closes #202
Closes #154 

This pull request adds the long awaited scar removal feature to TopoStats. Configuration for the scar removal has been added in the config file, and good/useful defaults have been set for the parameters. 

The scar removal is done by the new `Scars` class. Much like with the `Filters` class, its instantiation takes an image and the required parameters, then one can call the `remove_scars()` class method on the instance. `remove_scars()` is the only method intended to be used externally to the class, though it is written in a way that you could call other class methods if you really wanted to. 

There is a config option to turn off the scar removal if wanted.

This scar removal has been deemed to be of good enough quality, and it has been tested on a reasonable sample of scarred images, however with more usage, bugs may still be found. 

An example of what the scar removal in action:

With scars:
image

Scars removed:
image


MaxGamill-Sheffield (changes requested):
Awesome work! Results look great!

The one thing most of my comments are about is whether this should be a separate class in `filters.py`, just because this is going to be used in the filters workflow and it seems a unusual to have a separate config block for this where it could be incorporated into the filters block and maybe a scar_removal subsection.

tests/conftest.py
```diff
@@ -243,6 +249,22 @@ def random_filters(test_filters_random_with_mask: Filters) -> Filters:
...
+    """Scars object using a synthetic image"""
+    scars = Scars(
+        img=synthetic_scars_image,
+        filename="",
+        run=True,
+        removal_iterations=2,
+        threshold_low=1.5,
+        threshold_high=1.8,
+        max_scar_width=2,
+        min_scar_length=1,
```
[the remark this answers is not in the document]
SylviaWhittle: For the `max_scar_width` and `min_scar_length`, yes. I think it's easier to debug and see what is going on if the script is allowed to count scars that are only 1 pixel wide, and also only two pixels tall. Otherwise there would be a lot more combinations of pixel patterns that I would need to make (because of the number of permutations increasing). 

For the other two variables, these are just what work as a discriminator between the low value pixels and the high value pixels that I placed in the image. 

All of this is to say that I created the image and the tests before trying it on real data, so once I found an algorithm that worked on the tests, I had to update it for the real data. I think that having differing values in the tests here is okay since the base functionality does not change when using different values.

topostats/filters.py
```diff
@@ -290,7 +291,7 @@ def gaussian_filter(self, image: np.ndarray, **kwargs) -> np.array:
             **kwargs,
         )
 
-    def filter_image(self) -> None:
+    def filter_image(self, scars_config: dict) -> None:
```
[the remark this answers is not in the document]
SylviaWhittle: When setting out writing the `Scars` class, I considered this. I decided against it because I thought it would make the `filters.py` file even more massive, and far more difficult to develop / comprehend. 

I also decided against inheritance because this also would increase complexity. Also doing it this way gives us a little freedom to use the `Scars` class in a more modular way in the future if needs be - we might want to run scar removal independently of `filters`, which would be made easier with the modular way of implementing it - though perhaps I'm overdoing the future-proofing?

I agree that the method for passing the `scars` config seems a little odd on the surface, however the reason for including the `scars` input there, is that the `Scars` class needs to be instantiated during the filtering process, and passing multiple sets of `kwargs` is inadvisable in `init()` methods. 

What do you think? 😄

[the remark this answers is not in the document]
SylviaWhittle: @ns-rse what do you think about this?

tests/conftest.py
```diff
@@ -130,6 +130,12 @@ def small_mask() -> np.ndarray:
     return RNG.uniform(low=0, high=1, size=SMALL_ARRAY_SIZE) > 0.5
 
 
+@pytest.fixture
+def synthetic_scars_image() -> np.array:
+    """Small synthetic image for testing scar removal."""
+    return np.load(RESOURCES / "synthetic_scar_image.npy")
```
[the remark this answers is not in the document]
SylviaWhittle: It's quite small - shape of 20x20 IIRC. Perhaps a little too big though to write in the code? Up to you!

tests/test_scars.py
```diff
@@ -0,0 +1,17 @@
...
+BASE_DIR = Path.cwd()
+RESOURCES = BASE_DIR / "tests" / "resources"
+
+
+def test_remove_scars(synthetic_scar_removal: Scars):
+    """Test removal of scars"""
+
+    scars_removed = synthetic_scar_removal.remove_scars()
+
+    target = np.load(RESOURCES / "synthetic_scar_removal.npy")
```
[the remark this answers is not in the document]
SylviaWhittle: As above, 20x20 IIRC. It was also a bit of an experiment into storing output images as `.npy` rather than images with axes.

topostats/scars.py
```diff
@@ -0,0 +1,470 @@
...
+        -------
+        None
+        """
+
+        # Spread scars that have close to threshold edge-points
+        for row in range(marked.shape[0]):
+            # Spread right
+            for col in range(1, marked.shape[1]):
+                if marked[row, col] >= threshold_low and marked[row, col - 1] >= threshold_high:
+                    marked[row, col] = threshold_high
```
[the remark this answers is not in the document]
SylviaWhittle: This part of the code removes scar pixels that are not high enough to break the `threshold_high` barrier, *and* not next to a pixel that breaks the `threshold_high` barrier - essentially allowing scars to bleed a little to the sides, but only if they are definitely scars. Does this answer your question?

topostats/filters.py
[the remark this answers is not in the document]
ns-rse: I think keeping `Scars` as its own class is eminently sensible.

One approach of using inheritance (and I haven't used it much at all) might be to have this pattern, passing the arguments to Scars appropriately using `super()`.

```
class Filters(Scars):
    def __init__(...):
        super().__init__(...)
```

That is `Filters` inherits `Scars` because you want filtering to have the functionality to remove scars.

I don't think this would necessarily require moving configuration sections around as you could simply combine the two dictionaries and pass them in when instantiating `Filters()`.

```
filter_config.update(scars_config)
```

topostats/scars.py
```diff
@@ -0,0 +1,470 @@
...
+
+            # Spread left
+            for col in range(marked.shape[1] - 1, 0):
+                if marked[row, col - 1] >= threshold_low and marked[row, col] >= threshold_high:
+                    marked[row, col - 1] = threshold_high
+
+    def remove_too_short_scars(self, marked: np.ndarray, threshold_high: float, min_scar_length: int) -> None:
```
ns-rse, line 241: Pedantic but you can ditch the `_too` from this...

```suggestion
    def remove_short_scars(self, marked: np.ndarray, threshold_high: float, min_scar_length: int) -> None:
```

topostats/scars.py
```diff
@@ -0,0 +1,470 @@
...
+                min_scar_value = min(min_scar_value, img[row + k, col])
+                max_border_value = max(img[row, col], img[row + k + 1, col])
+                # Check if scar ended
+                if min_scar_value - max_border_value > threshold_low * stddev:
+                    while k:
+                        val = (img[row + k, col] - max_border_value) / stddev
+                        marked[row + k, col] = val
+                        k -= 1
+
+    def mark_if_negative_scar(
```
ns-rse, line 135: I'm surprised this and other functions don't raise a PyLint [no-self-use](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-self-use.html) error. :thinking:

tests/test_scars.py
```diff
@@ -0,0 +1,17 @@
...
+from pathlib import Path
+import numpy as np
+
+from topostats.scars import Scars
+
+BASE_DIR = Path.cwd()
+RESOURCES = BASE_DIR / "tests" / "resources"
+
+
+def test_remove_scars(synthetic_scar_removal: Scars):
```
ns-rse, line 11: This is akin to a regression test since it's testing the overall result.

Unit-tests test the individual methods for their action under different conditions and it would be useful to have unit tests for the different methods in place.

topostats/filters.py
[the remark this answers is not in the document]
SylviaWhittle: Thank you for that comprehensive analysis @ns-rse. Given that it *might* be possible that we want to have values derived from the filtering process affect parameters in scar removal, the current solution, while weird might be the simplest and least likely to cause problems for future development. 

@MaxGamill-Sheffield what are your thoughts about this?

topostats/scars.py
```diff
@@ -0,0 +1,470 @@
...
+        row_col: tuple,
+        stddev: float,
+        img: np.ndarray,
+        marked: np.ndarray,
+        threshold_low: float,
+        max_scar_width: int,
+    ) -> None:
+        """
+        Determine if the points below and including the pixel at the specified row and column are an positive scar.
+        If they are, mark them in the marked 2d np.ndarray. Note that this only detects positive scars.
```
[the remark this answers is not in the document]
SylviaWhittle: Positive scars are scars that "point" in the positive direction, ie they lay above the data. Negative scars are scars that are lower than the data. Got any ideas for a less confusing term than positive / negative? Perhaps upper / lower?

tests/conftest.py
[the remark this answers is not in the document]
ns-rse: As I've commented on the #387 I think this is the correct way of doing things and we need to stop continually changing values in `topostats/default_config.yaml` whenever a new feature or refactor is undertaken.

If tests require specific values change those in the dictionary for the specific test but leave `default_config.yaml` as is. To which end I think #302 could probably be reopened for discussion as there hasn't been much and certainly not a consensus on what should be used (but that is likely the major problem with experimental conditions introducing variation that precludes having a one-size-fits-all configuration).

tests/_regtest_outputs/test_run_topostats.test_process_scan_lower.out
```diff
@@ -1,3 +1,3 @@
...
 Molecule Number                                                                                                                                                                                                                                                                                                                                                                
-0                312.321826  135.710289    2.047651    6.802569     4.468766       4.524109   -0.812277    -0.43474      -0.573088    -0.578251 -31.751035  54.908751           106.156919                 5.459192                 13.041404               71.195532      2.388889     lower  13.079421   5.459192        46.737667     False              5.96906  minicircle
+0                312.332803  135.688334    2.028629    6.781691     4.495498       4.558832   -0.811737   -0.436235      -0.572944    -0.579688 -31.829952  54.908751           106.156919                 5.459192                 13.041404               71.195532      2.388889     lower  13.079421   5.459192        47.836563     False              5.96906  minicircle
```
ns-rse, line 3: Do we know that these changes are because of a scar in `minicricles.spm`?

topostats/scars.py
```diff
@@ -0,0 +1,470 @@
...
+        self.img = img
+        self.filename = filename
+        self.run = run
+        self.removal_iterations = removal_iterations
+        self.threshold_low = threshold_low
+        self.threshold_high = threshold_high
+        self.max_scar_width = max_scar_width
+        self.min_scar_length = min_scar_length
+
+    def mark_if_positive_scar(
```
ns-rse, line 74: A unit test would have an artificially created positive scar in isolation passed into this method and check to see if it is marked as a positive scar.

In addition a negative scar would also be passed in and the result tested to ensure its not marked as a positive scar.

It should be possible to write one test and paramterize the test.

topostats/scars.py
```diff
@@ -0,0 +1,470 @@
...
+                        stddev=stddev,
+                        img=img,
+                        marked=marked,
+                        threshold_low=threshold_low,
+                        max_scar_width=max_scar_width,
+                    )
+
+        elif direction == "negative":
+            # Find negative scars
+            for row in range(img.shape[0] - 1):
```
ns-rse, line 359: You could perhaps reduce code duplication by having the conditional `if: ... elif:` within the loop as you are looping over the same range of `row`/`col` regardless of whether the `direction` is `positive` or `negative.

```
for row in range(img.shape[0] - 1):
    for col in range(img.shape[1]):
        if direction == "positive:
            self.mark_if_positive_scar(...)
        elif direction == "negative":
            self.mark_if_negative_scar(...)
        else:
            raise ValueError(f"Invalid direction : {direction}")
```

topostats/scars.py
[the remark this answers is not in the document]
ns-rse: I'd suggest not using `upper` / `lower` as they are already used in thresholding and there is then scope for conflation.

Documenting and explaining the nomenclature clearly in docstrings as well as under `docs/workflow.md` (or child documents thereof) would be useful and noting their meaning in Notebooks that are under development would be clear I think.

topostats/filters.py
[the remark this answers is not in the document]
MaxGamill-Sheffield: I agree that it's easier to understand and use but disagree about it being used outside of filters. Filters is essentially our pre-processing stage to end up with a suitable image we can then perform analysis on. Although, called filters, I believe that scar removal is included in this pre-processing pipeline.

With the modularisation of the filters class, I think that once the Filters object is initialised, code-competent users should be able to build their own pipelines using the functions within it (flatten then align rows vs align rows then flatten etc), in which I still think scars should belong. This might also be helped with an interactive GUI that we are looking into in collaboration with comp sci, where people could drag-and-drop modules (in this case our funcs) to build pipelines (possibly very far into the future however).

For those reasons, I like the idea of class inheritance but I can understand this is unfamiliar territory and may take a while to implement and we might want to avoid it, but this would help to remove the additional config input which I do think should be part of the filters config. Can talk about this on Tuesday.

tests/conftest.py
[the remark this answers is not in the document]
MaxGamill-Sheffield: More just to reduce the number of files in the repo as if we had one for all 138 tests that a lot of files

tests/test_scars.py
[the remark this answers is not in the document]
MaxGamill-Sheffield: I agree I like it better, especially for the instances where the data has changed and not the plotting params

topostats/filters.py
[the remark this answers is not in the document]
SylviaWhittle: @ns-rse @MaxGamill-Sheffield should I refactor this so that the Scars class is a child of Filters?

topostats/scars.py
SylviaWhittle: Made `mark_if_negative_scar` and `mark_if_positive_scar` static.

topostats/filters.py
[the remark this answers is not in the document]
MaxGamill-Sheffield: I think it would make sense as then we wouldn't have to pass it separate params and can include it in the filters workflow but I understand that it could slow down development, unless this isn't too difficult of a task and doesn't require much refactoring but I wouldn't know as I'm also not familiar with child classes.

If it does, you could bring it up with Alice in tomorrow's meeting?

tests/_regtest_outputs/test_run_topostats.test_process_scan_lower.out
SylviaWhittle: I believe they are. I have added a plot to show the marked scars mask, and there are a few scars for `minicircle.spm` that are inconsequential with regards to the grains that are detected, however will affect quantities such as average image value and the flattening by a **tiny** amount, which seems to affect each grain a little. 

image

I think the spatial and geometric quantities such as the centres and the radii change because the edge pixels of the grains that are just above / below the threshold are shifted by functions in the second round of flattening, and are moved across the threshold for being detected as a grain. I am currently trying to confirm this with some plots of which pixels in the grains change.

topostats/filters.py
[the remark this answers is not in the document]
SylviaWhittle: I'll give it a go!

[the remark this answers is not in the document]
ns-rse: My current, poorly informed thinking, is as follows...

1. "filtering"  is the overall first pre-processing step (its actually many steps).
2. Its implemented in `Filters()` and currently has methods for `align_rows()` , `median_flatten()` etc.
3. We wish to add additional steps, i.e. scar removal.
4. That has been implemented in the `Scars()` class.
5. This additional functionality should be available _within_ the `Filters()` class.
6. There may be additional pre-processing steps that get implemented that should be available under `Filters()` (e.g. a class for `BlobRemoval()`).
7. `Filters()` should then inherit those, then when people wish to have control over the steps in which they process things they only need to instantiate `Filters()` and the functionality of `Scars()` / `BlobRemoval()` / `SomethingElseFunky()` are available in that instantiated object as they are all.

This could be done via inheritance but we're not dealing with representations of objects that have similar properties that inherit properties from a parent as in many examples of inheritance. 

To this end must apologise as I feel I may have wasted time and key-strokes by using the wrong term/design pattern as what I should have suggested is a design pattern known as composition as `Filters()` would be a composite of its existing methods, `Scars()`, `BlobRemoval()`, `SomethingElseFunky()` and anything else that is required.

An exposition of this is given in [Inheritance and Composition: A Python OOP Guide – Real Python](https://realpython.com/inheritance-composition-python/)

Some (crude) pseudo-code might be...

```
class Filters()
    def __init__(self, ...):
        ... <other options>
        self.scars = None
        self.blob_removal = None

    ... <all existing Filters methods>

    def filter_image(self) -> None:
        self.images["initial_median_flatten"] = self.median_flatten(self.images["pixels"], mask=None)
        ... <other filter steps required>
        if self.scars is not None:
            self.images["scars_removed"] = self.scars.remove_scars()
        if self.blob_removal is not None:
            self.images["blobs_removed"] = self.blob_removal.remove_blobs()
        ... <remaining filter steps after scar/blob removal> 

filters = Filters(**filter_config)
if scar_config["run"]:
    scar_config.pop("run")
    filters.scars = Scars(**scar_config)
if blob_config["run"]:
    blob_config.pop("run")
    filters.blob_removal = BlobRemoval(**blob_config)
# Make it so!
filters.filter_image()
```

tests/test_scars.py
SylviaWhittle: Added more tests 👍

topostats/scars.py
SylviaWhittle: Unit tests added 👍

topostats/scars.py
SylviaWhittle: Done 👍

topostats/scars.py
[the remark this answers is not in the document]
SylviaWhittle: Made it a little clearer 👍

SylviaWhittle:
I have removed the `Scars` class from `scars.py`, and now it is just a module with a set of functions, one externally-accessible function, `remove_scars`, and several internal functions, not intended for external use.

tests/conftest.py
```diff
@@ -242,6 +254,21 @@ def random_filters(test_filters_random_with_mask: Filters) -> Filters:
     return test_filters_random_with_mask
 
 
+@pytest.fixture
+def synthetic_scar_removal_args(synthetic_scars_image: np.ndarray) -> dict:
```
ns-rse, line 258: These are in essence configuration options and on line 31 the `default_config.yaml` is loaded to a fixture. Subsequent to that chunks of the configuration are subset out to their own fixtures (see lines 44-110) and changes made to any of the options where tests run with different parameters.


I think it would therefore be consistent to follow this convention and have on line 113 something like...

```
@pytest.fixture
def scarring_config(default_config: Dict) -> Dict:
    """Configurations for scar removal"""
    config = default_config["remove_scars"] 
    config["threshold_low"] = 1.5
    config["threshold_high"] = 1.8
    config["max_scar_width"] = 2
    config["min_scar_length"] = 1
    config.pop("run")
    return config
```

topostats/default_config.yaml
```diff
@@ -18,6 +18,13 @@ filter:
     upper: 1.0
   gaussian_size: 1.0121397464510862 # in px
   gaussian_mode: nearest
+remove_scars:
```
ns-rse, line 21: Discussion really, should this remain as a separate set of options or should it be nested within `filter:`?

If the later then this chunk needs indenting to...

```
  remove_scars:
    run: true
    removal_iterations: 2
    threshold_low: 0.250
    threshold_high: 0.666
    max_scar_width: 4
    min_scar_length: 16
```

topostats/run_topostats.py
```diff
@@ -129,6 +129,7 @@ def process_scan(
     img_path_px2nm: Dict[str, Union[np.ndarray, Path, float]],
     base_dir: Union[str, Path],
     filter_config: dict,
+    scars_config: dict,
```
ns-rse, line 132: If the `scars_config` section is going to remain as an independent top-level dictionary in `default_config.yaml` then this will be required.

But there is the option of moving the whole `remove_scars` configurations and nesting it under the `filter:` section.  If this is going to be done then you wouldn't need to pass `scars_config` in as a separate argument to `process_scan` as it is nested within `filter_config`.

topostats/run_topostats.py
```diff
@@ -427,6 +430,7 @@ def main():
         process_scan,
         base_dir=config["base_dir"],
         filter_config=config["filter"],
+        scars_config=config["remove_scars"],
```
ns-rse, line 433: Same as above...

If the `scars_config` section is going to remain as an independent top-level dictionary in `default_config.yaml` then this will be required.

But there is the option of moving the whole `remove_scars` configurations and nesting it under the `filter:` section.  If this is going to be done then you wouldn't need to pass `scars_config` in as a separate argument to `process_scan` as it is nested within `filter_config`.

topostats/filters.py
```diff
@@ -296,7 +297,7 @@ def gaussian_filter(self, image: np.ndarray, **kwargs) -> np.array:
...
```
ns-rse, line 300: If the `scars_config` section is going to remain as an independent top-level dictionary in `default_config.yaml` then this will be required.

But there is the option of moving the whole `remove_scars` configurations and nesting it under the `filter:` section.  If this is going to be done then you wouldn't need to pass `scars_config` in as a separate argument to `process_scan` as it is nested within `filter_config`.

topostats/scars.py
```diff
@@ -0,0 +1,445 @@
...
+        Note that length here is horizontal, as scars are thin, horizontal features.
+
+    Returns
+    -------
+    self.img
+        The original 2-D image with scars removed, unless the config has run set to False, in which case it
+        will not remove the scars.
+    """
+
+    if run:
```
ns-rse, line 413: The logic for controlling whether to run the filtering should sit a level up within the `Filter.filter_image()` method there's no point making the call to `scars.remove_scar()` if its not going to do anything.

topostats/filters.py
```diff
@@ -315,10 +316,15 @@ def filter_image(self) -> None:
         self.images["initial_tilt_removal"] = self.remove_tilt(self.images["initial_median_flatten"], mask=None)
         self.images["initial_quadratic_removal"] = self.remove_quadratic(self.images["initial_tilt_removal"], mask=None)
 
+        # Remove scars
```
ns-rse, line 319: The logic for controlling whether to run the filtering should sit here within the `Filter.filter_image()` method as there's no point making the call to `scars.remove_scar()` if its not going to do anything.  

```suggestion
        # Remove scars
        if scar_config["run"]:
            scar_config.pop("run")
```

topostats/filters.py
ns-rse, line 300: If `scars_config` is moved and nested under `filter_config` in `default_config` then there will be no need to pass this as an option here, rather the `def __init__(self)` will need updating to take the argument, which will be a keyword to `**filter_config` and as I've hopefully correctly suggested in the below comment the logic of whether to run scarring is then used before making the call to the scarring method itself.

topostats/scars.py
```diff
@@ -0,0 +1,445 @@
...
+                break
+            min_scar_value = min(min_scar_value, img[row + k, col])
+            max_border_value = max(img[row, col], img[row + k + 1, col])
+            # Check if scar ended
+            if min_scar_value - max_border_value > threshold_low * stddev:
+                while k:
+                    val = (img[row + k, col] - max_border_value) / stddev
+                    marked[row + k, col] = val
+                    k -= 1
+
```
ns-rse, line 73: Could add a`LOGGER.debug()` statement to indicate a scarring has been completed.

```suggestion
            LOGGER.debug("Scar removed")
```

topostats/scars.py
```diff
@@ -0,0 +1,445 @@
...
+    # If sharp enough rise
+    min_scar_value = img[row + 1, col]
+    max_border_value = img[row, col]
+    if min_scar_value - max_border_value > threshold_low * stddev:
+        # Possibly in a scar
+        for k in range(1, max_scar_width + 1):
+            if row + k + 1 >= img.shape[0]:
+                # Bottom of image, break
+                break
```
ns-rse, line 64: Could add a `LOGGER.debug()` statement to indicate a scarring has been completed.

```suggestion
                LOGGER.debug("Bottom of image.")
                break
```

topostats/scars.py
```diff
@@ -0,0 +1,445 @@
...
+    col = row_col[1]
+
+    # If sharp enough dip
+    min_scar_value = img[row + 1, col]
+    max_border_value = img[row, col]
+    if min_scar_value - max_border_value < -threshold_low * stddev:
+        # Possibly in a scar
+        for k in range(1, max_scar_width + 1):
+            if row + k + 1 >= img.shape[0]:
+                # Bottom of image, break
```
ns-rse, line 126: Similar to above, `LOGGER.debug()` statements might be useful here and below.

topostats/scars.py
```diff
@@ -0,0 +1,445 @@
...
+    scar_mask: np.ndarray
+        A boolean image of pixels that determine which values are flagged as scars and therefore should
+        be interpolated over in the original data image.
+
+    Returns
+    -------
+    None
+    """
+
+    for row in range(img.shape[0]):
```
ns-rse, line 334: Could iterate directly over arrays with...
```suggestion
for row, col in np.index(img.shape):
```

tests/test_scars.py
```diff
@@ -0,0 +1,632 @@
...
+                    [3],
+                    [3],
+                    [3],
+                    [0],
+                ]
+            ),
+            2,
+            3,
+        ),
+        # Unsuccesssful scar segment: negative scar rather than positive
```
ns-rse, line 50: :+1: Really useful for understanding what each set of parameters are doing, thanks. :smile:

tests/test_scars.py
```diff
@@ -0,0 +1,632 @@
...
+
+    target_img = np.load(RESOURCES / "test_scars_synthetic_remove_scars.npy")
+    target_mask = np.load(RESOURCES / "test_scars_synthetic_remove_scars_mask.npy")
+
+    np.testing.assert_array_equal(scars_removed, target_img)
+    np.testing.assert_array_equal(scar_mask, target_mask)
+
+
+@pytest.mark.parametrize(
+    "potential_positive_scar, target, threshold_low, max_scar_width",
```
ns-rse, line 25: Purely style but I think its more logical to have the parameters first and then `target` at the end.

Same would apply to other parameterised tests.

ns-rse (changes requested):
Great work @SylviaWhittle saves the overhead of instantiating a class where most methods were `@staticmethod` and keeps things simple.

Most of my comments pertain to where configuration is being defined and the logic to control the running of scar removal.

The configuration doesn't _have_ to be nested, but it makes things a bit simpler in my view as we wouldn't have to subset those out from the configuration dictionary and pass it around, they're a key/value within `config_filter`. This removes a few lines of code from `run_topostats.py` (keep code bloat down is good in my view).

topostats/scars.py
SylviaWhittle: This condition triggers `image_width * max_scar_width)` times per image (estimate about 3000-10000). Would this not slow down the processing? IIRC, printing values slows down scripts a lot, I assume logging does too?

SylviaWhittle: I tried it, and it does not slow it down given that we use `INFO` as our logging state, and even with `DEBUG` as our logging state, it doesn't slow the process down dramatically. A concern I have though, is that when a developer is debugging TopoStats generally, and sets the logger state to `DEBUG`, they will get thousands of lines of `"Bottom of image"` appearing, which might make the process of debugging harder than if they just add debug lines as and when they need? I'm inexperienced with regards to best debugging practice, so do tell me if I'm wrong about this.

topostats/default_config.yaml
SylviaWhittle: Nested 👍

tests/test_scars.py
SylviaWhittle: Done 👍

tests/conftest.py
SylviaWhittle: Done 👍

topostats/filters.py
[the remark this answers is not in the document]
SylviaWhittle: Change obsolete due to `scars.py` not being OOP now.

ns-rse (commented):
Not sure if I've understood this correctly but I think the logic here can be simplified/clearer.

topostats/scars.py
```diff
@@ -333,32 +333,31 @@ def _remove_marked_scars(img: np.ndarray, scar_mask: np.ndarray) -> None:
...
-                    img[row + k - 1, col] = interp_val
-                    scar_mask[row + k - 1, col] = 0.0
-                    k -= 1
-                    LOGGER.debug("Scar removed")
+    for row, col in np.ndindex(img.shape):
+        if scar_mask[row, col] == 1.0:
+            # Scar detected
+            # Find out how tall it is
+            scar_width = 1
+            while True:
```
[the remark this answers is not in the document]
SylviaWhittle: @ns-rse I thought that in making the code a bit more verbose with the loop like that, it would make it more clear, sorry. I have made a change that is based off of your suggestion. I had to edit it to make it work but I think it's in the same spirit?

Let me know if this change is the kind of thing you were intending.

ns-rse (commented):
One minor question about `pylint`.

topostats/filters.py
```diff
@@ -35,6 +36,7 @@ def __init__(
         threshold_absolute: dict = None,
         gaussian_size: float = None,
         gaussian_mode: str = "nearest",
+        remove_scars: dict = None,
```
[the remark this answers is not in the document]
SylviaWhittle: @ns-rse It passes for me 🤔 

```
❯ git checkout SylviaWhittle/remove_scars
Already on 'SylviaWhittle/remove_scars'
Your branch is up to date with 'origin/SylviaWhittle/remove_scars'.
❯ pylint topostats/filters.py

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

Are we sure that it's the `dict = None` line? When I asked you about the issue, it was in relation to `plottingfuncs.py` and not `scars.py`, and it has a few other dictionary arguments?

My pylint setup is behaving very funky though. It flags some imports (that **definitely** work when running the code) as unable to be imported. Eg:

```
❯ pylint tests/test_scars.py
************* Module test_scars
tests/test_scars.py:6:0: E0401: Unable to import 'topostats' (import-error)

------------------------------------------------------------------
Your code has been rated at 8.65/10 (previous run: 8.65/10, -0.00)
```

So maybe it's something broken with my pylint?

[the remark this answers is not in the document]
ns-rse: Ah, my confusion about which file you'd asked me about, sorry.

I've seen that sort of error before and found it confusing too. It might be a [VSCode/venv/pylint](https://stackoverflow.com/questions/51095449/pylint-false-positive-e0401-import-errors-in-vscode-while-using-venv) conflict that is raising it.

ns-rse (approved):
Sorry for not getting back to this sooner.

Looks good to go.

PR #399, opened 2022-12-09
Ns rse/curvature

ns-rse:
This is the start of my refactoring of `JeanDu/168-curvature-merge` (see #365).

It is by no means complete and doesn't currently get used in any processing (but this is by design as it makes the Pull Request easier to review).

I've focused on the `find_curvature()` method that @Jean-Du has written and have done the following...

* Made a separate sub-module (`dnacurvature`) and class (`Curvature`).
* Split the steps of calculating derivatives and local curvature into individual methods.
* Made these methods work for a single molecule/grain rather than looping over a dictionary or list of molecules. Methods/functions should do a single thing, if they need reusing then call them from a loop.
* Rather than appending three circular molecules I instead take a certain number of points from the end and append to the start and, vice versa, take a certain number of molecules from the start and append to the end. How many to take is configurable and determined by the `np.gradient()` option `edge_order` (see [numpy.gradient](https://numpy.org/doc/stable/reference/generated/numpy.gradient.html) for more details).
* Introduced a series of tests that calculate gradient for circles, ellipses and parabolas. These are parameterised to test with different shapes and `edge_order`. For readability these are split over individual tests for each shape, probably could have had it all as one large set of `pytest.mark.parametrize()` but would have been harder to follow (**NB** - Fixtures defined in `conftest.py` wouldn't work here because the `points` to create are passed in and there are different functions generating the different shapes).

Outstanding things to refactor are...

* [ ] Refactor revised skeletonising method implemented on branch `JeanDu/168-curvature-merge` in `dnatracing.py`. It may be wise to wait for #387 to be completed before adding this.
* [ ] Refactor plotting functions added to `dnatracing.py` to `plottingfuncs.py`. Should work with a single molecule.
* [ ] Save length and curvature statistics on a per grain basis from `topostats.run_topostats.main()`, these will be attributes of the class of type `pd.DataFrame()` and will therefore carry the `to_csv()` method to simplify writing.


Jean-Du:
Hi, Neil, thanks a lot for this! Where is the curvature stored and is there any output for curvatue yet?

ns-rse:
Each test is parameterized with its expected output using a `pytest.parameterize` fixture that defines the variables that are input to the test and those that are output.

There are three tests which I lifted from your original code, one for circles, one for ellipses and one for parabolas.

The fixture for circles defines the following parameters

```
"points, radius, edge_order, expected_first_derivative, expected_second_derivative, expected_local_curvature",
```

That is the number of points, the radius, the edge_order, the expected_first_derivative, expected_second_derivative and the expected_local_curvature.

There then follows sets of values where the number of points, radius or edge_order varies and the expected values are given.

The same follows for the ellipses and parabola.

This is how we can write tests without having any real-life data, because for example we would expect to have all local-curvature values equal for a circle, and given these are small samples we can, if needs by validate the values by hand.

ns-rse:
After discussion with @Jean-Du a concern was that the curvature for closed loops (circles or ellipses) was dependent on where within the shape the calculation started. I've now added tests to check this by using `np.roll()` to "roll/shift" the arrays for circles and ellipses around and repeat the calculations. The resulting first and second derivatives and local curvature are compared to what is expected **but** this requires rolling the resulting arrays backwards the opposite number of spaces so they can be compared directly to the expected values.

As a final check the curvature from the original shapes is compared to the backward-rolled curvature.

Pleasingly the curvature is the same regardless of what point the calculation starts, and so we can now be confident that this will be true when it comes to actual grains. :tada:

tests/tracing/test_dnacurvature.py
```diff
@@ -0,0 +1,941 @@
...
+    np.testing.assert_almost_equal(np.var(curvature.local_curvature), expected_variance)
+
+
+@pytest.mark.parametrize(
+    "points, radius, shift, edge_order, expected_variance, expected_first_derivative, expected_second_derivative, expected_local_curvature",
+    [
+        (
+            10,
+            10,
+            5,  # Shift array by 3 points
```
Jean-Du, line 297: Why does this say 5 in the code but 3 in the comment?

tests/tracing/test_dnacurvature.py
```diff
@@ -0,0 +1,941 @@
...
+            np.asarray(
+                [5.0, 0.1765308, 0.04620573, 0.04620573, 0.1765308, 5.0, 0.1765308, 0.04620573, 0.04620573, 0.1765308]
+            ),
+        ),
+        (
+            10,
+            5.0,
+            1.0,
+            0.0,
+            3,  # Shift array by 5 points
```
Jean-Du, line 652: The same typo as before

Jean-Du, line 652: Same issue as before; I'd have changed it myself but I am not sure if you'd want to change the code or the comment

tests/tracing/test_dnacurvature.py
```diff
@@ -0,0 +1,941 @@
...
+            np.asarray(
+                [5.0, 0.1765308, 0.04620573, 0.04620573, 0.1765308, 5.0, 0.1765308, 0.04620573, 0.04620573, 0.1765308]
+            ),
+        ),
+        (
+            10,
+            5.0,
+            1.0,
+            0.0,
+            3,  # Shift array by 7 points
```
Jean-Du, line 691: Same issue as above

Jean-Du (changes requested):
Thanks a lot for your work! Everything looks good, apart from a minor issue with `test_dnacurvature.py` where some shifts aren't implemented to the same number as described in the comments.

tests/tracing/test_dnacurvature.py
ns-rse: Thanks :+1: corrected.

tests/tracing/test_dnacurvature.py
ns-rse: Thanks :+1: corrected.

Jean-Du (approved):
All looks good!

PR #398, opened 2022-12-08
Adds option to generate config file; adds to documentation and clarifies

ns-rse:
Closes #393 #395 #396

* Adds `run_topostats --create-config-file` option to generate `config.yaml` in place.
* Documents usage as well as adding basics on command line interface
* Removes link to ReadTheDocs, in preference of working towards documentation that builds against versions and is hosted on GitHub Pages
* Tidys up linting of files that were touched as part of this commit.
* Fixes errors in .pre-commit-config.yaml so that black, flake and pylint all run.

PR #400, opened 2022-12-09
384 | Refactors validation and introduces test for validate_config()

### LINKED ISSUE #384: Refactor Validation
Currently validation has grown from #302 such that I've written individual functions for each schema that is being validated. This is poor design. Instead there should be a single validation function and `Schemas` should be passed in for validation.

ns-rse:
Closes #384

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,20 +18,20 @@ repos:
       - id: black
         types: [python]
         additional_dependencies: ['click==8.0.4']
-        args: ["--extend-exclude", "topostats/plotting.py", "topostats", "tests", "*.ipynb"]
+        args: ["--extend-exclude", "topostats/plotting.py"]
 
   - repo: https://github.com/pycqa/flake8.git
     rev: 5.0.4
     hooks:
       - id: flake8
-        args: ["--extend-exclude=topostats/_version.py", "topostats", "tests", "*.ipynb"]
+        args: ["--extend-exclude=topostats/_version.py"]
         additional_dependencies: [flake8-print]
         types: [python]
   - repo: local
     hooks:
       - id: pylint
-        args: ["--rcfile=.pylintrc", "--ignore-patterns=*_version.py", "topostats", "tests"]
+        args: ["--rcfile=.pylintrc"]
         name: Pylint
-        entry: python -m pylint.__main__
+        entry: python -m pylint
         language: system
         files: \.py$
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,7 +20,8 @@ fail-on=
 fail-under=10.0
 
 # Files or directories to be skipped. They should be base names, not paths.
-ignore=CVS
+ignore=CVS,
+       _version.py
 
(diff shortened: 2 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -18,20 +18,20 @@ repos:
...
       - id: flake8
-        args: ["--extend-exclude=topostats/_version.py", "topostats", "tests", "*.ipynb"]
+        args: ["--extend-exclude=topostats/_version.py"]
         additional_dependencies: [flake8-print]
         types: [python]
   - repo: local
     hooks:
       - id: pylint
-        args: ["--rcfile=.pylintrc", "--ignore-patterns=*_version.py", "topostats", "tests"]
+        args: ["--rcfile=.pylintrc"]
```
SylviaWhittle, line 33: Does this mean that `.ipynb` notebook files will be checked with pylint ?

.pylintrc
```diff
@@ -172,8 +173,7 @@ spelling-store-unknown-words=no
 [MISCELLANEOUS]
 
 # List of note tags to take in consideration, separated by a comma.
-notes=FIXME,
-      XXX,
+notes=XXX,
```
SylviaWhittle, line 176: Just wondering, why remove the FIXME from the notes that pylint warns about?

topostats/run_topostats.py
```diff
@@ -291,7 +290,7 @@ def process_scan(
                     if key in ["grain_image", "grain_mask", "grain_mask_image"]
                 }
                 grainstats = {}
-                for direction in grains.directions.keys():
+                for direction, _ in grains.directions.items():
```
SylviaWhittle, line 293: I assume there is a reason for doing so, but why iterate over `items()` instead of `keys()` if only the key is to be used?

SylviaWhittle (approved):
Looks good. A couple of minor comments / questions. 

I ensured that the config validation works and that the `test_validation.py` picks up errors that break functionality.

Thank you!

.pylintrc
ns-rse: I've removed this as there are a few `FIXME` notes around in the codebase that do need addressing. They should really be written up into issues, but would likely get overlooked as specific tasks as other work takes priority but by leaving them in place they act as an _aide-mémoire_ when myself or anyone else is looking at/working on that section of code.

topostats/run_topostats.py
ns-rse: Yep, its [pylint consider-using-dict-items C0206](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/consider-using-dict-items.html).

.pre-commit-config.yaml
ns-rse: Ah, good spot. I didn't copy over the necessary change to `setup.cfg` from #398 which excludes `ipynb` from `pylint`, once that is merged it will sort things it.

It was because #398 was awaiting approval and I wanted to get this quick improvement to validation out for PR that I copied the changes over, need to be more attentive next time.

PR #405, opened 2022-12-15
Splining for linear molecules

Jean-Du:
Introduced splining for linear molecules. In addition, the method `get_disordered_trace` has been renamed to `get_disordered_traces` for consistency with other methods.


topostats/tracing/dnatracing.py
```diff
@@ -401,11 +402,7 @@ def get_splined_traces(self):
...
 
-        step_size = int(7e-9 / (self.pixel_size))  # 3 nm step size
-        interp_step = int(1e-10 / self.pixel_size)
-        # Lets see if we just got with the pixel_to_nm_scaling
-        # step_size = self.pixel_size
-        # interp_step = self.pixel_size
+        step_size_px = int(self.step_size_m / (self.pixel_size))
 
         # FIXME : Iterate over self.fitted_traces directly use either self.fitted_traces.values() or self.fitted_trace.items()
         for dna_num in sorted(self.fitted_traces.keys()):
```
ns-rse, line 408: I would recommend iterating over `self.fitted_traces.items()` using as otherwise I think `pylint` might complain about this, but realise this is outside the scope of your additions (but I started writing this before clocking which method this was under).

```suggestion
        for dna_num, _ in sorted(self.fitted_traces.items()):
```

topostats/tracing/dnatracing.py
```diff
@@ -493,69 +488,88 @@ def get_splined_traces(self):
...
-                #        out = interp.splev(np.linspace(0,1,nbr*step_size), tck)
-                #        splined_trace = np.column_stack((out[0], out[1]))
-                #        self.splined_traces[dna_num] = splined_trace
-                #    except ValueError: #if the trace is really messed up just delete it
-                #        self.mol_is_circular.pop(dna_num)
-                #        self.disordered_trace.pop(dna_num)
-                #        self.grains.pop(dna_num)
-                #        self.ordered_traces.pop(dna_num)
 
             else:
```
ns-rse, line 496: As per above making this handling of linear spines its own method is going to be a better design.

topostats/tracing/dnatracing.py
```diff
@@ -493,69 +488,88 @@ def get_splined_traces(self):
...
-                                          range(i, len(self.fitted_traces[dna_num][:, 0]), step_size)])
-                    y_sampled = np.array([self.fitted_traces[dna_num][:, 1][j] for j in
-                                          range(i, len(self.fitted_traces[dna_num][:, 1]), step_size)])
-
-                    interp_f = interp.interp1d(x_sampled, y_sampled, kind='cubic', assume_sorted=False)
+                for i in range(step_size_px):
+                    x_sampled = np.array(
+                        [
+                            self.fitted_traces[dna_num][:, 0][j]
+                            for j in range(i, len(self.fitted_traces[dna_num][:, 0]), step_size_px)
```
ns-rse, line 504: What is the difference between `fitted_traces` and `ordered_traces` that means they need separate handling?

(I need to check the logic elsewhere to see what is being done to `ordered_traces`.

ns-rse (changes requested):
First pass on review, I think splitting splinning into two methods, one for linear and one for circular and having each method take only a single molecule will simplify the design (and make it easier to incorporate into the refactored DNA tracing in the long run).

Whilst the regression tests have been updated there are no unit-tests for the splinning method that has been modified and we should have unit tests for each step in place.

Splitting methods into linear and circular as suggestedf above will also make it easier to write unit-tests for each.

topostats/tracing/dnatracing.py
```diff
@@ -423,34 +420,32 @@ def get_splined_traces(self):
...
 
-                # if nbr/step_size > 4: #the degree of spline fit is 3 so there cannot be less than 3 points in splined trace
+            if self.mol_is_circular[dna_num]:
 
-                # ev_array = np.linspace(0, 1, nbr * step_size)
-                ev_array = np.linspace(0, 1, int(nbr * step_size))
+                ev_array = np.linspace(0, 1, nbr * step_size_px)
 
-                for i in range(step_size):
+                for i in range(step_size_px):
```
ns-rse, line 433: The advantage of having a method that works with a single molecule is that you then don't need to have all the complicated logic here to extract the arrays.

topostats/tracing/dnatracing.py
```diff
@@ -493,69 +488,88 @@ def get_splined_traces(self):
                 if not self.splining_success:
```
ns-rse, line 488: Further I think the naming convention could be clearer.

It requires a fraction more thought to think what the result of

```
if not splinng_success:
```

...is (for me at least!).

It would be clearer to have...

```
splining_failure = False
```

...defined at the start of the method and then within the `except ValueError:`...

```
except ValueError:
    ...
    splining_failure True
...

if splining_failure:
    continue
```

That said I'm not sure you even need this `if not self.splining_success:` or `if splining_failure:` at all since the errors are captured by virtue of the `try: ... except: ...` statements and since there is nothing beyond `continue` in the `if:` statement then its kind of redundant.

topostats/tracing/dnatracing.py
```diff
@@ -493,69 +488,88 @@ def get_splined_traces(self):
...
+                        )
 
-                    # tck = interp.splrep(x_sampled, y_sampled, quiet = 0)
-                    # out = interp.splev(np.linspace(start_x,end_x, nbr*step_size), tck)
-                    splined_trace = np.column_stack((x_new, y_new))
+                        try:
+                            tck, u = interp.splprep([x, y], s=5, per=0, quiet=1)
+                            out = interp.splev(np.linspace(0, 1, nbr * step_size_px), tck)
+                            splined_trace = np.column_stack((out[0], out[1]))
+                        except ValueError:  # sometimes even the ordered_traces are too bugged out so just delete these traces
```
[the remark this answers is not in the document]
Jean-Du: Hi, Neil, thanks a lot for your review! I'll have a look at your suggestions and address the ones that are quick to fix. As for spliting the splining into linear and circular, I am not sure if this is necessary, because the process for splining linear and circular molecules are almost the same, with the only difference being the parameters used when calling `interp.splprep`; I don't think this change in numbers should merit a separate method?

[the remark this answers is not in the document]
Jean-Du: Would it be a better idea for me to merge the linear and circular splining into one part, with the if statement only applied before the `interp.splprep()` instead of the entire code block?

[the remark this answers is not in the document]
Jean-Du: Also, do we have unit tests for the splining of circular molecules? If so, I should be able to modify them to be compatible with both circular and linear molecules

topostats/tracing/dnatracing.py
Jean-Du: The idea is that, the splinining is based on `fitted_traces`, but if that gives errors, the splining is done to `ordered_traces` instead

ns-rse:
I've looked closer at my suggestion of a method for `linear` and a method for `circular` splining and actually I think you just need a single method to which you can pass arguments, if I split out the code within `if self.mol_is_circular[dna_num]:` into its own file and the code under `else:` into its own file then I can use `diff` to compare the code line-by-line to each other and the only differences are....

```
❱ diff circular.py linear.py 
17c17
<                         tck, u = interp.splprep([x_sampled, y_sampled], s=2, per=2, quiet=1, k=3)
---
>                         tck, u = interp.splprep([x_sampled, y_sampled], s=5, per=0, quiet=1, k=3)
37c37
<                             tck, u = interp.splprep([x, y], s=2, per=2, quiet=1)
---
>                             tck, u = interp.splprep([x, y], s=5, per=0, quiet=1)
56,63d55
< 
<                 if not self.splining_success:
<                     continue
< 
<                 spline_average = np.divide(spline_running_total, [step_size_px, step_size_px])
<                 del spline_running_total
<                 spline_average = np.delete(spline_average, -1, 0)
<                 self.splined_traces[dna_num] = spline_average
```

...and so the arguments for `s` and `per` are the main things that differ between the two sets of code, and then if its circular it then does some additional steps.

So splitting out the step that extracts the x and y arrays will reduce duplicated code too (you have the same steps twice under `if:` and twice under `else:`).

topostats/tracing/dnatracing.py
ns-rse: What errors might arise and why? If we can understand those its surely useful. Is there a reason why `fitted_traces` is preferred if it works? Is there any reason not to just use `ordered_traces` all the time?

Jean-Du: @ns-rse  This was from Joe's code so I am not sure what errors might arise either. `fitted_traces` are better than `ordered_traces` because they take the height of the pixels into account, instead of treating it as binary data. For a detailed explanation, see Section 2.1.5.4 of this paper: https://www.sciencedirect.com/science/article/pii/S1046202321000207

topostats/tracing/dnatracing.py
Jean-Du: If I recall correctly, this is for sampling the coordinate array so that the splining can be made smoother, and this mechanism should still be in place even with a method for a single molecule

topostats/tracing/dnatracing.py
```diff
@@ -493,69 +488,88 @@ def get_splined_traces(self):
...
+                                self.ordered_traces[dna_num][:, 1][j]
+                                for j in range(i, len(self.ordered_traces[dna_num][:, 1]), step_size_px)
+                            ]
+                        )
 
-                    # tck = interp.splrep(x_sampled, y_sampled, quiet = 0)
-                    # out = interp.splev(np.linspace(start_x,end_x, nbr*step_size), tck)
-                    splined_trace = np.column_stack((x_new, y_new))
+                        try:
+                            tck, u = interp.splprep([x, y], s=5, per=0, quiet=1)
```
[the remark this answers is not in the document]
Jean-Du: I tried removing the `u` but it seems to cause problems, so I replaced it with `_`

topostats/tracing/dnatracing.py
```diff
@@ -459,23 +454,23 @@ def get_splined_traces(self):
...
+                            tck, u = interp.splprep([x, y], s=2, per=2, quiet=1)
+                            out = interp.splev(np.linspace(0, 1, nbr * step_size_px), tck)
                             splined_trace = np.column_stack((out[0], out[1]))
                         except ValueError:  # sometimes even the ordered_traces are too bugged out so just delete these traces
                             self.mol_is_circular.pop(dna_num)
-                            self.disordered_trace.pop(dna_num)
+                            self.disordered_traces.pop(dna_num)
                             self.grains.pop(dna_num)
                             self.ordered_traces.pop(dna_num)
                             self.splining_success = False
```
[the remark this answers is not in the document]
Jean-Du: Fixed

Jean-Du:
Closing this PR as @ns-rse is planning to refactor `dnatracing.py` and use code from this branch instead.

PR #407, opened 2022-12-16
Maxgamill sheffield/406 broken skeleton images

MaxGamill-Sheffield:
This PR:
- streamlines plottingfuncs by getting the colourmap in `__init__`
- adds option to provide a mask colourmap in config
- adds dpi option (this will be set within the skeleton images and not in the config)
- updates validation with new colourmaps


MaxGamill-Sheffield:
This appears to have taken most of the skeleton stuff with it - closing for now

PR #408, opened 2022-12-16
Molecule # in all_statistics.csv; correcting plotting validation

### LINKED ISSUE #403: Add molecule number in the all_statistics.csv spreadsheet
Introduce a column of molecule number in the all_statistics.csv spreadsheet and make sure it is consistent with the molecule number in the cropped out images of all the grains, so that the user can look up the statistics for each individual grain.

**Additional context**

ns-rse:
Closes #403

`all_statistics.csv` now includes the molecule number as the first column.

Also in testing it highlighted that the recent merge of #400 hadn't correctly validated the plotting configuration so corrects that too.

### DIFF
--- a/topostats/run_topostats.py
+++ b/topostats/run_topostats.py
@@ -387,7 +387,7 @@ def main():
     # Load plotting_dictionary and validate
     plotting_dictionary = pkg_resources.open_text(__package__, "plotting_dictionary.yaml")
     config["plotting"]["plot_dict"] = yaml.safe_load(plotting_dictionary.read())
-    validate_plotting(config["plotting"]["plot_dict"])
+    validate_config(config["plotting"]["plot_dict"], schema=PLOTTING_SCHEMA, config_type="YAML configuration file")
 
     # FIXME : Make this a function and from topostats.utils import update_plot_dict and write tests
     # Update the config["plotting"]["plot_dict"] with plotting options
@@ -450,7 +450,7 @@ def main():
                 pbar.update()
     results = pd.concat(results.values())
     results.reset_index()
-    results.to_csv(config["output_dir"] / "all_statistics.csv", index=False)
+    results.to_csv(config["output_dir"] / "all_statistics.csv", index=True)
     LOGGER.info(
         (
             f"All statistics combined for {len(img_files)} images(s) are "


Jean-Du (changes requested):
The addition of molecule number is useful and should be merged, but this doesn't quite resolve #403 because the mismatch between the molecule number in the cropped out images and the molecule number in the spreadsheet still exists. For example, in an image with 10 grains numbered 0-9, if molecule 3 is removed in dnatracing.py, the resulting spreadsheet becomes 0-8, instead of 0-2 and 4-9.

Would it be possible to merge this without closing #403 ?

ns-rse:
Sorry @Jean-Du I didn't understand that level of detail from the description of #403.

Some questions...

* Why would/is `DNAtracing()` removing molecules?
* How do you know if a molecule has been removed? Is it by comparison to the images grainstats output?

I won't merge this if its not solving the problem, thats not how software development is meant to work. It just introduces a job to do further down the line. Now I have a refined/clearer specification I will go back and solve the problem in light of additional information.

Jean-Du:
Hi @ns-rse , sorry about not making the issue clearer in #403 .
`dnatracing.py` removes molecules whose skeleton contains too few pixels to be traced through the method `purge_obvious_crap`. I knew a molecule has been removed by comparing the number of cropped images produced in the `grains` folder with the number of molecules in the grains statistics spreadsheet. There is one more molecule in the former compared to the latter. I will supply an example file via slack since the data can't be published.

ns-rse:
> dnatracing.py removes molecules whose skeleton contains too few pixels to be traced through the method purge_obvious_crap. I knew a molecule has been removed by comparing the number of cropped images produced in the grains folder with the number of molecules in the grains statistics spreadsheet. There is one more molecule in the former compared to the latter. I will supply an example file via slack since the data can't be published.

This makes we wonder then, how often will a grain be detected by `Grains()` and the statistics produced by `GrainStats()` be required if there are no `DNATracing()` statistics for it?

I ask because if the answer is "_never_" then the functionality of removing small grains performed by the `purge_obvious_crap()` (horrible name, not at all descriptive) should be moved to the `Grains()` stage because then the need to include empty data points for `DNATracing()` stats when merging with `GrainStats()` output disappears. Molecule numbers can and will be included, but they will always align because for every grain we will _always_ have both Grain _and_ Tracing statistics.

I'm trying to get a better understanding of what is required and address the issue as far back in the processing pipeline rather than patching/fixing things further down the line because of inappropriate workflows/decisions in the past.

Jean-Du:
@ns-rse 
> I ask because if the answer is "never" then the functionality of removing small grains performed by the purge_obvious_crap() (horrible name, not at all descriptive) should be moved to the Grains() stage because then the need to include empty data points for DNATracing() stats when merging with GrainStats() output disappears.
A potential issue with that would be, `purge_obvious_crap` (I didn't name this but I agree it needs to be renamed to something else) requires the skeletonisation of the molecule, which isn't something we want to do to all data analysed by topostats. If the user is analaysing data of proteins, for example, as opposed to DNA, the grains aren't meant to be skeletonised or traced, and the entire `dnatracing.py` needs to be turned off, including the method `purge_obvious_crap`.
I can't give a definite answer to your question, but I am more inclined to say it's not 'never', since I have encountered situations where the tracing works badly on some molecules but the tracing-independent statistics are still useful. However I also recognise that, in this situation, the user can simply turning `dnatracing.py` off.

ns-rse:
Will deal with this as part of a major overhaul/refactor of DNATracing.

Issue #367, opened 2022-11-17
Colour maps slightly wrong


SylviaWhittle:
**Describe the bug**
@derollins pointed out to me that the colour maps for new Topo are still slightly off. See attached images for comparison (left - old topo, right - new topo). This hinders work on projects that utilised the old TopoStats. Eddie also believes that the old colour scheme is better, so if the current scheme could be adjusted to match the previous one, that would be great. 

<img width="587" alt="image" src="https://user-images.githubusercontent.com/86117496/202489820-ea123a5a-ce32-470a-bb1f-adc4a29b5c1a.png">

**(Not required if using Docker) Your computer configuration (please complete the following information):**
- All computer configurations
- All Python versions

MaxGamill-Sheffield:
I think I know how to fix the nanoscope colour gradient thing, or at least find the correct gradient values. When someone has time, please could someone get these via:
1. Open Gwyddion
2. Edit > Colour gradients
3. Click the Nanoscope option
4. Click “create new item based on selected one”
5. Click the new one
6. Click “edit selected item”

Now for the tricky bit:
1. Let me know if the arrows have the same spacing (this will be trickier if not)
2. Click each of the triangles and let me know the RGB values

MaxGamill-Sheffield:
Closed as resolved in #409

ns-rse:
Sorry for not linking these two, had dropped out of visibility having been merged. One more closed issue closer to a new release. :tada:

PR #409, opened 2022-12-18
Maxgamill sheffield/406 plot improvements

### LINKED ISSUE #406: Skeletons appear broken when plotting
**Describe the bug**
Single pixel thick skeletons appear broken on the produced plots, this seems to be due to the reduced dpi.

**To Reproduce**
This has been made to break-up the knots and cats branch. I have obtained skeletons from dnatracing, and used the skeleton as a mask for the `data2` input to `plottingfuncs.Images`

**Expected behavior**
Skeletons should be visible:
image

**Output**
Skeletons seem broken:
image

MaxGamill-Sheffield:
Fixes #406
This PR adds:
- dpi and mask_cmap into plottingfuncs (the dpi setting will be applied on the plottingdict level and not be an adjustable parameter in the config - mainly because this is useful for skeletons and not much else)
- gets the cmaps in `init` instead of in the code for readability
- adds `gwyddion` and `blu` colourmaps
- adds mask_cmap to default config
- Adds these to validation
- Adds single test using small array and mask for the 'blu' colourmap

Example with `blu` "mask_cmap" selected:
minicircle_upper_processed_masked


SylviaWhittle:
Thanks for this @MaxGamill-Sheffield, the blue mask is a lot more clear. Also I really like the structural changes you have made, should increase readability and reduce errors in the future 👍 

I have a couple small questions about this PR before I review it:

- You mentioned that you have added a single test using the small array format for the blu colormap - I can't see any edits to testing files, have I missed it / have you intentionally left it out after writing the PR?

- Could you explain how the dpi works? I can't see in the code where it is actually set. At least not from searching for `dpi` in the codebase. I thought you meant that it would be set in `plotting_dictionary.yaml` where you mention that it would be set on the plotting dict level but I must have been mistaken.

MaxGamill-Sheffield:
@SylviaWhittle Yep I didn't push the commits whoops, pushed now.

And `dpi` is basically the matplotlib plot resolution factor. It's not been included anywhere in the plotting dict yet but for things like skeletons images, it needs to be increased (see #406) so will be set for these and similar plots.

MaxGamill-Sheffield:
Seems to receives numpy errors within the igor binary wave import (no import changes have been made in these edits. Thus I'm going to leave it for now:
```
from igor import binarywave
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/igor/binarywave.py:110: in <module>
    1:_numpy.complex, # NT_CMPLX, makes number complex.
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/numpy/__init__.py:284: in __getattr__
    raise AttributeError("module {!r} has no attribute "
E   AttributeError: module 'numpy' has no attribute 'complex'
```

My numpy version: `1.23.5`

topostats/theme.py
```diff
@@ -78,3 +82,21 @@ def nanoscope():
 
         colormap = LinearSegmentedColormap.from_list("nanoscope", vals, N=256)
         return colormap
+
+    @staticmethod
+    def gwyddion():
```
ns-rse, line 87: I'm naive, but how does this differ qualitatively from `nanoscope`?

#367 documents that the colourmaps are slightly off, if this corrects that why is a new colourmap being introduced rather than correcting `nanoscope`?

topostats/default_config.yaml
```diff
@@ -48,5 +48,6 @@ plotting:
   zrange: [null, null]  # low and high height range for core images (can take [null, null])
   colorbar: true  # Options : true, false
   axes: true # Options : true, false (due to off being a bool when parsed)
-  cmap: nanoscope # Options : nanoscope, afmhot
+  cmap: nanoscope # Options : nanoscope, afmhot, gwyddion
+  mask_cmap: blu # Options : blu, jet_r and any in matplotlib
```
ns-rse, line 52: I might have missed it but would it be worth adding a `dpi` option to the `default_config.yaml`?

Otherwise users would have to load all the classes and modules into an interactive Python session, process their images and then instantiate the `Images()` class with the desired `dpi` option.

```suggestion
  mask_cmap: blu # Options : blu, jet_r and any in matplotlib
  dpi: figure # Options figure or int (integer) representing the dots per inch
```

topostats/validation.py
```diff
@@ -212,8 +212,10 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
             "cmap": Or(
                 "afmhot",
                 "nanoscope",
-                error="Invalid value in config for 'plotting.cmap', valid values are 'afmhot' or 'nanoscope",
+                "gwyddion",
+                error="Invalid value in config for 'plotting.cmap', valid values are 'afmhot', 'nanoscope' or 'gwyddion'",
             ),
+            "mask_cmap": str,
```
ns-rse, line 218: If adding a `dpi` option to `default_config.yaml` validation here would be required.

```suggestion
            "mask_cmap": str,
            "dpi": Or("figure", lambda n: 1 > 0, error="Invalid value for 'dpi', valid values are 'figure' or an int value > 0"),
```

topostats/theme.py
MaxGamill-Sheffield: No, worries, thanks for asking :) 

Here I was just testing if the colourmaps could easily be created from the values inside the Gwyddion software. Gywddion.net (or `gwyddion` here) is one such widely used and industry standard colormap that when I saw it work I thought we should just keep it in incase anyone has a preference for it. 

nanoscope is the just our labs preferred publication colourmap but there are others, such as this one.

This led me to add to #367 with the steps for someone to replicate the the colourmap without simulating an spm image which I'm unsure how easy that is. If you want I could try and obtain those values and fix that issue in this PR?

ns-rse (changes requested):
Excellent work as always @MaxGamill-Sheffield 

I think it may be necessary to add an option to `default_config.yaml` so that users of the CLI (`run_topostats`) can specify the `dpi` value and see the results. I hope the suggestions are useful.

topostats/default_config.yaml
MaxGamill-Sheffield: So I thought it be best not to, in order to keep the config less complex, and instead add it into the plotting dictionary as this really only applies to the skeletons. what are your thoughts?

ns-rse:
The builds are failing because of an issue with `igor` (which hasn't been updated for quite some time).

Looking at the [log](https://github.com/AFM-SPM/TopoStats/actions/runs/3725997173/jobs/6319307732#step:5:17) the issue is with line 110 of `igor/binarywave.py` where the type `numpy.complex` is attempted to be used.

The pipeline installed [Numpy 1.24.0](https://github.com/AFM-SPM/TopoStats/actions/runs/3725997173/jobs/6319307732#step:4:23) and the documentation for [1.24.0](https://numpy.org/doc/1.24/reference/arrays.scalars.html#complex-floating-point-types) does not list such a type.

I've checked this between Numpy 1.23.4 and 1.24.0...

## Numpy 1.23.4

```python
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '1.23.4'

In [3]: np.complex
<ipython-input-2-087dad13a955>:1: DeprecationWarning: `np.complex` is a deprecated alias for the builtin `complex`. To silence this warning, use `complex` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.complex128` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  np.complex
Out[3]: complex
```

## Numpy 1.24.0

```python
In [1]: import numpy as np

In [2]: np.__version__
Out[2]: '1.24.0'

In [3]: np.complex
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 np.complex

File ~/.virtualenvs/topostats/lib/python3.10/site-packages/numpy/__init__.py:284, in __getattr__(attr)
    281     from .testing import Tester
    282     return Tester
--> 284 raise AttributeError("module {!r} has no attribute "
    285                      "{!r}".format(__name__, attr))

AttributeError: module 'numpy' has no attribute 'complex'

```

v1.24.0 was [released 2022-12-19](https://github.com/numpy/numpy/releases/tag/v1.24.0) and the release notes state

> The deprecation for the aliases np.object, np.bool, np.float, np.complex, np.str, and np.int is expired (introduces NumPy 1.20). Some of these will now give a FutureWarning in addition to raising an error since they will be mapped to the NumPy scalars in the future.

...and so the behaviour is expected. It links to the Pull Request ([22607](https://github.com/numpy/numpy/pull/22607)) that closed the deprecation and that in turn links to a related issue discussing deprecation of the `np.bool` alias (see [#22021](https://github.com/numpy/numpy/issues/22021)). 

For now I would pin the dependency on Numpy v1.23.4 and open an issue with the [igor](http://git.tremily.us/?p=igor.git) although their repository appears to be self-hosted (of sorts) and doesn't appear to have the ability to report issues.  It may be a case of having to clone and fix this (quick testing reveals `np.complex_` is valid under Numpy 1.24.0 but whether that is the correct solution I can not say at this stage).

Ideally I would like to avoid pinning versions of packages though, currently we have `scikit-image-0.19.3` pinned because of a breaking issue but I'm keeping an eye out periodically for that to be fixed and then we can unpin.  Pinning versions introduces a lot of headaches in package dependencies.

ns-rse:
After discussing in todays meeting my suggestion of `dpi` as an option in `default_config.yaml` should perhaps be shifted to `plotting_dictionary.yaml` as the `dpi` value is only pretinent to the images that have been skeletonised.

MaxGamill-Sheffield:
With Eddies help with the nanoscope file, I now have the modified the nanoscope colourmap to address #367 producing the resulting change:
Old colourmap:
minicircle_processed copy

New colourmap:
minicircle_processed

SylviaWhittle:
That colour map looks great!

MaxGamill-Sheffield (commented):
Looks good thanks for the test image @SylviaWhittle

topostats/theme.py
SylviaWhittle: I think keeping PRs small and modularising issues is best, so probably keep it to its own issue and PR? Not an expert here though.

SylviaWhittle (approved):
This looks good to me and works as intended as far as I can tell. Maybe wait until @ns-rse reviews it too since I had a bit of a hand in editing a few lines and so it feels a bit conflicting.

topostats/plottingfuncs.py
```diff
@@ -93,6 +94,8 @@ def __init__(
             Format to save the image as.
         histogram_log_axis: bool
             Optionally use a logarithmic y axis for the histogram plots.
+        histogram_binis: int
```
ns-rse, line 97: Typo
```suggestion
        histogram_bins: int
```

topostats/plottingfuncs.py
```diff
@@ -223,17 +236,20 @@ def save_figure(self):
                     format=self.save_format,
                     bbox_inches="tight",
                     pad_inches=0,
+                    dpi=self.dpi,
                 )
             else:
-                plt.savefig((self.output_dir / f"{self.filename}.{self.save_format}"), format=self.save_format)
+                plt.savefig(
+                    (self.output_dir / f"{self.filename}.{self.save_format}"), format=self.save_format, dpi=self.dpi
```
ns-rse, line 243: I realise this isn't a change introduced with this PR but [savefig()](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html) infers the `format` from the file-extension and as these are identical the code can be reduced to...

```suggestion
                    (self.output_dir / f"{self.filename}.{self.save_format}"), dpi=self.dpi
```

ns-rse (changes requested):
One typo in a docstring and one minor reduction in code suggested, but otherwise good to do.

tests/test_plottingfuncs.py
```diff
@@ -196,3 +196,21 @@ def test_plot_and_save_non_square_bounding_box(
...
+
+@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
+def test_mask_cmap(plotting_config: dict, tmp_path: Path) -> None:
+    """Test the plotting of a mask with a different colourmap (blu)."""
+    plotting_config['mask_cmap'] = 'blu'
+    fig, _ = Images(
+        data=array,
+        output_dir=tmp_path,
+        filename="colour.png",
+        data2=mask,
```
ns-rse, line 213: Test is failing because `data2` which wasn't a very informative option name was changed to `masked_array`. Wouldn't have shown as a merge conflict as this is a new test.

```suggestion
        masked_array=mask,
```

MaxGamill-Sheffield: I've changed this because it now doesn't have to be a binary masked array as the colourmap can take maps and not just single colours, and I thought that it might confuse people. 

I see now that I've kind of confused myself so will change all back to masked array (as technically the array can be anything)

ns-rse (approved):
Cool, thanks, ready to merge :+1:

MaxGamill-Sheffield:
Forgot to add: closes #367

PR #413, opened 2022-12-21
img_name from methods > self.filename; disabling flake8/pylint checks

ns-rse:
Very minor changes I noticed in `topostats/filters.py` with regards to logging with the `self.filename` rather than `img_name`.  Also disables a few `pylint` errors although they suggest some refactoring _may_ be useful (low priority though).

Also includes a change to `setup.cfg` to pin the Numpy version because a deprecation of methods causes problems with the `igor` library (see [comment](https://github.com/AFM-SPM/TopoStats/pull/409#issuecomment-1359184427)).

To pull this change over use [git cherry-pick](https://www.atlassian.com/git/tutorials/cherry-pick) after this commit has been merged...

```
git checkout main && git pull
git checkout <your-branch>
git cherry-pick 4ca0e43
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ zip_safe = False
 install_requires =
   igor
   matplotlib
-  numpy
+  numpy==1.23.4
   pandas
   pySPM
   pyyaml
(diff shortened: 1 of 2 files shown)


SylviaWhittle (approved):
Good spot, thank you for cleaning this up.

ns-rse:
Not a problem at all.

PR #416, opened 2022-12-23
Regression test for run_topostats entry point

### LINKED ISSUE #308: Regression test to ensure default configuration works
On the back of introducing validation of input (see #299 and #303) and the proposal to remove `config/example.yaml` and only work with `topostats/default_config.yaml` it has been suggested by @SylviaWhittle (see [comment](https://github.com/AFM-SPM/TopoStats/pull/303#issuecomment-1274334113) and [clarification](https://github.com/AFM-SPM/TopoStats/pull/303#issuecomment-1274385686)) that we should have a regression test to ensure `run_topostats` works without being explicitly provided a configuration file, that is that `topostats/default_config.yaml` is loaded, passes validation and runs to completion.

ns-rse:
Closes #308

Introduces a couple of extra tests...

* One that checks the help is displayed when `run_topostats [-h|--help]` is run.
* One that checks `run_topostats` runs completely using the `topostats/default_config.yaml`.

I couldn't work out how to get to run using the package resources because if no arguments are added then because the call to `run_topostats` is made within `pytest` the `sys.argv` that `pytest` was invoked with are instead passed and so this errors. Instead we explicitly specify the configuration file with the `--config topostats/default_config.yaml`.

For a better explanation of this situation see [this thread](https://stackoverflow.com/a/55260580/1444043).

I also removed `minicircle.spm` from the root of the repository, there is a copy under `tests/resources/minicircle.spm` anyway.

**EDIT** : Tests started failing so I made some necessary commits to debug the CI/CD, turns out the default number of cores which was `4` was the source of the failure as some of the runners have only `2` cores. I've therefore reduced this in the `default_config.yaml` and noted that it could safely be increased in the documents.


SylviaWhittle:
Thanks for this @ns-rse.

I am experiencing a problem when trying to run the new test though. When attempting to run the test, the test hangs. All other tests run successfully. 

It uses 24GB of memory before my system tells me to force shut down VSCode due to the memory cap being reached. I do have a huge amount of data in my topostats folder, could this be the issue? Is it trying to process it all? I don't know where to go with troubleshooting this, as tests don't print diagnostic information as far as I can tell. 

image

Do you have any advice as to what I should do to diagnose the issue?

ns-rse:
Hi @SylviaWhittle 

Thanks for testing this and apologies I forgot to add that it requires very few `.spm` in the repositories directory. The reason for this is because `default_config.yaml` has `base_dir: ./` and so it searches the current directory for all of the specified files (by default `.spm`). We could modify the test to use a specified directory but then that's not testing the `default_config.yaml`.

If you're able to move all other `.spm` you work on out of the project directory that would I think be the preferred solution. Sorry if that messes up your workflow.

SylviaWhittle (approved):
I like the completion line that you've added!

PR #425, opened 2023-01-09
Fix histogram bins 'maximum size exceeded' crash by specifying bin size

SylviaWhittle:
This bulky PR fixes an issue where files would very rarely be unable to produce histogram plots due to the automatic bin size calculator internal to `matplotlib` crashing (caused by an error thrown by `np.linspace` : `Maximum Size Exceeded`). The solution here is simply to set the bin size as a constant 200. This seems to be the best size balance between horizontal resolution and being able to see individual columns.


ns-rse (changes requested):
Thanks for picking up and addressing this issue @SylviaWhittle .

I'm not aware of the context and how frequently these plots are required but I'm not a fan of hard-coding variables within functions. It means should users wish to change them they have a hard time doing so.

To which end I think it would be preferable to have `bins` as an option in the `__init__` of the `Images` class and set `self.bins` so users can specify/modify this value should they need to.

In turn `default_config.yaml` will need...

```
plotting:
  ...
  bins: 200
```

..adding and the `DEFAULT_CONFIG_SCHEMA` defined in `topostats/validation.py` will need a check made that the value passed is non-zero (see the current `lambda` functions for checking `absolute_thresholds` are above or below `0` which I've just put a PR out to remove).

Finally, if you think this is an error or bug in Matplotlib its well worth reporting it up-stream so it can be fixed.

SylviaWhittle:
> Thanks for picking up and addressing this issue @SylviaWhittle .
> 
> I'm not aware of the context and how frequently these plots are required but I'm not a fan of hard-coding variables within functions. It means should users wish to change them they have a hard time doing so.
> 
> To which end I think it would be preferable to have `bins` as an option in the `__init__` of the `Images` class and set `self.bins` so users can specify/modify this value should they need to.
> 
> In turn `default_config.yaml` will need...
> 
> ```
> plotting:
>   ...
>   bins: 200
> ```
> 
> ..adding (it should carry through neatly as a `**kwarg`).
> 
> The `DEFAULT_CONFIG_SCHEMA` defined in `topostats/validation.py` will need a check made that the value passed is non-zero (see the current `lambda` functions for checking `absolute_thresholds` are above or below `0` which I've just put a PR out to remove).
> 
> Finally, if you think this is an error or bug in Matplotlib its well worth reporting it up-stream so it can be fixed.

Excellent point about the hardcoded variables, I'll do that.

I'm not sure if this is a bug with the histogram plotting, or if it is just a sensible limitation... It's very hard to tell / diagnose specifically since it's a bit of a black box. All I have been able to determine is the error that is thrown, and that it does not happen for manually defined bin numbers.

SylviaWhittle:
Added config option for `histogram_bins` instead of a hard-coded value.

ns-rse (approved):
Excellent, thanks @SylviaWhittle and well remembered on updating `docs/configuration.md` :+1: Looks good to go.

Issue #242, opened 2022-08-02
Google Colab Notebooks for running analyses


ns-rse:
**Is your feature request related to a problem? Please describe.**
It would be useful for users who wish to understand and tinker with the steps of processing to have example Jupyter Notebooks, running in Google Colab, to go through worked examples of processing images.

**Describe the solution you'd like**

Initially a simple Jupyter Notebook that takes the user through the steps of processing `minicircle.spm`.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

ns-rse:
Some work is on-going on this see `notebooks/` under branch [ns-rse/242a-notebooks branch](https://github.com/AFM-SPM/TopoStats/tree/ns-rse/242a-notebooks).

It depends on the workflow which will change as refactoring of DNA Tracing and other plotting functions progresses, although the early stages of Filtering, Grains, GrainStats are fairly stable and unlikely to change much.

PR #426, opened 2023-01-10
Relaxes validation of absolute_threshold

### LINKED ISSUE #423: Absolute values : validation failing
**Describe the bug**
As reported by @MaxGamill-Sheffield, the validation is failing for valid values of absolute threshold values. 

**To Reproduce**

Run topostats with a variety of absolute thresholds, there will be some / all that are invalid incorrectly. 

**Suggestions** 

@MaxGamill-Sheffield suggests that once this has been amended, we could also do zeroing steps so that the absolute thresholds are not just relative to the mean pixel value for the image.

ns-rse:
Closes #423

I understand the problem but do not feel this is an optimal solution given the situation that these absolute thresholds are image specific and the aim to use TopoStats to batch process images.

There are two threads here...

1) The need to have sensible values against which validation is performed as noted for discussion in #302.

2) The challenge of trying to write software and have defaults that cover all scenarios (or say ~80%).

Given the threshold can change on a per-image basis I do not think having user specified `absolute_threshold` for batch processing is sensible. Its highly likely some in a given batch will fail, users then split their batch of images into smaller batches with custom configs and process these. This seems convoluted and against the aim of batch processing.

Users can always load data into Notebooks and use TopoStats interactively to process on a per image basis (prototype notebooks are covered by issue #242, see also `ns-rse/242a-notebooks` branch for example). For batch processing the thresholds should be derived from the data itself using the `otsu` or `std_dev` method.

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -11,6 +11,7 @@ TEST_SCHEMA = Schema(
         "a": Path,
         "b": Or("aa", "bb", error="Invalid value in config, valid values are 'aa' or 'bb"),
         "positive_integer": lambda n: 0 < n,
+        "absolute_threshold": Or(int, float, error=("Invalid value in config should be type int or float")),
     }
 )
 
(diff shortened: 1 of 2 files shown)


SylviaWhittle:
Good points @ns-rse. Do you think @MaxGamill-Sheffield's suggestion of zeroing mid-processing would help here, since this might enable standardisation between images from the same sample? If I'm right on this, then batch processing images from the same sample would perhaps be consistent.

SylviaWhittle:
So during flattening, when the absolute threshold is being applied, we have currently:
image

However the absolute thresholds produce incorrect thresholds for grains, so if we zero the data between every step, we would have...

image

ns-rse:
Entirely open to discussion on this, although perhaps a separate issue/group meeting would be appropriate and I'd want to be informed/guided by data from a wide range of samples of different types and quality.

Rescaling or might help, but my intuition/guess is that variance will differ across images so even if each image is centered on zero the distributions will vary widely and so taking an absolute threshold would still be problematic. Assuming sizes follow a Gaussian distribution (after removal of noise via Otsu Thresholding) then using the mean -/+ N * std_dev would seem most appropriate (thanks to the Central Limit Theorem, ~95% of data points are within 1.96 standard deviations, ~99% within ~3 standard deviations etc.).

As this is a more substantial consideration such changes would I think be outside the scope of the Bug fixing milestone.

ns-rse:
Thanks @SylviaWhittle 

I had thought about the possibility of users specifying `upper` < `lower` and had a brief search of how to test this in the Schema framework but didn't find anything and given we may wish to revise this in light of the above comments opted not to spend any more time investigating.

PR #429, opened 2023-01-10
Corrects path output construction in get_out_path

### LINKED ISSUE #428: Images saved to wrong place & Processed dir error
**Describe the bug**
No matter where the output_dir is set, all images go into a `./Processed` folder that is made instead of `./<output_dir>/Processed`. This is an issue because the "`older_statistics.csv` file does want to go into the latter, throwing an error when this dir doesn't exist.

**To Reproduce**
Run topostats (main branch)
See "Processed" directory in `.`
Error thrown due to missing 

**Expected behavior**
Processed folder should be in "./<output_dir>"

**Output**
Something along the lines of :
`OSError: "Processed"  directory not found`

**Additional context**
FIX:
Remove the `.parent` on line 173 -> ` _output_dir = get_out_path(image_path, base_dir, output_dir) / "Processed"`

ns-rse:
Hopefully closes #428

Probably worth checking this branch with some of the instances where it has failed first though.

Some minor linting errors are also corrected (but not all resolved as one has been disabled).

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -144,9 +144,6 @@ def test_load_scan_jpk(load_scan_jpk: LoadScans) -> None:
     load_scan_jpk.img_path = load_scan_jpk.img_paths[0]
     load_scan_jpk.filename = load_scan_jpk.img_paths[0].stem
     image, px_to_nm_scaling = load_scan_jpk.load_jpk()
-    print(image.shape)
-    print(image.sum())
-    print(px_to_nm_scaling)
     assert isinstance(image, np.ndarray)
     assert image.shape == (256, 256)
     assert image.sum() == 286598232.9308627
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -5,7 +5,7 @@ from unittest import TestCase
 import numpy as np
 import pytest
 
-from topostats.io import read_yaml, LoadScans
+from topostats.io import read_yaml, find_images, get_out_path, LoadScans
 
 BASE_DIR = Path.cwd()
 RESOURCES = BASE_DIR / "tests" / "resources"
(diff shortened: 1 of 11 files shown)


SylviaWhittle:
(Oops - I posted a comment on #428 that was meant for here)

I tried this PR out on my system with a variety of different folder setups, and it seems to do as @ns-rse says. It does not fix the double periods issue (though this PR isn't intending to AFAIK - I checked just in case and to clarify), so that still needs a PR after this one.

Here is an example of the file structure, could someone verify if this is intended? Looks good to me.

image


Oh and a visual demonstration of the difference between the folder structures (new vs old):

image

ns-rse:
:thinking: I'm not sure this is correct either because my intention has always been to retain the directory structure under which images are found should be retained under `output`. Why? because if there are two files with the same name but in different directory levels they shouldn't get over-written.

With

```
base_dir = Path("./data/folder")
output_dir = Path("output")
image_path1 = base_dir / "minicircles/folder_a/folder_b/minicircles1.png.spm"
image_path2 = base_dir / "minicircles/folder_a/folder_b/minicircles2.png.spm"
image_path3 = base_dir / "minicircles/folder_a/minicircles1.png.spm"
```

Then the output should be (and I'm ditching `Processed` for the reason I've explained above that I think it is redundant since `output` _is_ the "processed" directory)...

```
get_out_path(image_path1, base_dir, output_dir)
PosixPath("output/minicircles/folder_a/folder_b/minicircles1.png/")

get_out_path(image_path2, base_dir, output_dir)
PosixPath("output/minicircles/folder_a/folder_b/minicircles2.png/")

get_out_path(image_path3, base_dir, output_dir)
PosixPath("output/minicircles/folder_a/minicircles1.png/")
```

As the code stands in this PR it gives...

```
get_out_path(image_path1, base_dir, output_dir)
PosixPath('output/minicircles1.png')

get_out_path(image_path2, base_dir, output_dir)
PosixPath('output/minicircles2.png')

get_out_path(image_path3, base_dir, output_dir)
PosixPath('output/minicircles1.png')
```

...which is problematic as we have image_path3 output over-writing image_path1 output.

I'll correct this tomorrow.

If no one has any objects to removing `Processed` I'll remove that too (I think it would have the added bonus that if `output_dir:` is set to the `base_dir` then it would result in a directory adjacent to each input file with the results once the above issue is corrected).

SylviaWhittle:
Thanks for the clarification @ns-rse. I like the suggestion of removing `Processed`, however I will ask the experimentalists tomorrow in-person (there's a group meeting).

MaxGamill-Sheffield:
Hi @ns-rse thanks for sorting this, I'll take a look at this PR after my meeting finishes at 5ish.

I just wanted to check in that @SylviaWhittle has passed on the need to keep the "Processed" directory. If not, this is due to the fact that if nested files are being processed, the "Processed" folder reduces the need to sift through all the processed images that would otherwise be in this directory, which also contains the nested data.

i.e.:
```
data - img1.spm
     - img2.spm
     - nested_data - img3.spm
```
would produce:
```
output - img1_processed
       - img1_masked
       - img2_processed
       - img2_masked
       - nested_data - img3_processed
                     - img3_masked
```

I hope that makes sense :)

tests/test_io.py
```diff
@@ -29,6 +29,92 @@ def test_read_yaml() -> None:
...
+        # Relative path, nested (two deep) under base_dir, with file_suffix
+        (
+            Path("/some/random/path"),
+            Path("images/today/test.spm"),
+            Path("output/here"),
+            Path("output/here/images/today/test"),
+        ),
+        # Relative path, nested under base_dir, no file suffix
+        (Path("/some/random/path"), Path("images/"), Path("output/here"), Path("output/here/images/")),
+        # Absolute path, nested under base_dir, output not nested under base_dir, with file_suffix
```
SylviaWhittle, line 82: Here it says `output not nested under base_dir`, what does this mean in contrast to the other tests?

SylviaWhittle: Ahah I think I get it. Including `/` or not switches between forcing the output to be just what the output dir is set as (with the `folder_name/image_name` directory added), or appending the input file structure on top of the output structure.

tests/test_utils.py
```diff
@@ -139,7 +76,10 @@ def test_folder_grainstats(tmp_path: Path, minicircle_tracestats: pd.DataFrame)
     """Test a folder-wide grainstats file is made"""
     input_path = tmp_path / "minicircle"
     minicircle_tracestats["Basename"] = input_path / "subfolder"
-    true_out_path = tmp_path / "subfolder" / "Processed"
-    Path.mkdir(true_out_path, parents=True)
+    out_path = tmp_path / "subfolder" / "processed"
+    Path.mkdir(out_path, parents=True)
+    print(f"###### out_path   : {out_path}")
```
SylviaWhittle: Print statements left in.

tests/test_io.py
ns-rse: Correct :smiley:

ns-rse:
Thanks @MaxGamill-Sheffield we had protracted discussion in `#experimental` yesterday and I think I've understood the desired structure and sussed it out.

PR #434, opened 2023-01-13
Corrects image paths, lints code, adds command line options and more

### LINKED ISSUE #362: Clear completion message
Starting to use Topostats it is hard to know if the programme has finished running successfully or it has just crashed since the final line currently looks the same as all the rest but just says it had written a config file. 

It would be great to have a clear (friendly!) message that lets you know that the programme has finished running without crashing. This could also highlight significant errors if there were any.

### LINKED ISSUE #391: Incorrect naming of output files when input files have multiple `.`s in them.
**Describe the bug**

While the issue of the crashes when processing files with multiple `.`s in the file name was resolved in #271, the strings following the second `.` is lost. This second string is important for distinguishing different iterations / versions of similar scans. 

Currently, if one processes two images with the same name before the second `.`, but different after the second `.`, the second processed image will be saved with the same name as the first and so overwrite the data of the first.

Example here:

File names that differ only in post-second `.` text:
image

Output data:
image

**To Reproduce**

Run two images with the same name before the second `.`, but differing after the second `.`. Notice that the second image overwrites the output for the first image.

**Expected behavior**
The output images should retain the entire names, but have the second `.` removed. Perhaps like this:

in:
```
data.01.spm
```

out:
```
data01.png
```

 - MacOS
 - topostats version: 2.0.0
 - Python version: 3.10.

### LINKED ISSUE #428: Images saved to wrong place & Processed dir error

### LINKED ISSUE #430: FileExistsError caused by file names containing periods
**Describe the bug**
When running multiple files with periods in their names, the part after the period is discarded when creating directories named after the files, so if the part before the period is identical, the following error is triggered: 
`FileExistsError: [Errno 17] Cannot create a file when that file already exists: 'Processed\\[name of the file before the period]\\grains\\lower'`

**To Reproduce**

- Command used to run topostats
```
run_topostats --config config.yaml
```

- Config file (you can attach this or paste it below)
```yaml
base_dir: ./
output_dir: ./output
warnings: ignore
cores: 4
quiet: false
file_ext: .spm
loading:
  channel: Height
filter:
  threshold_method: std_dev # Options : std_dev (default), otsu, absolute
  threshold_std_dev:
    lower: 10.0
    upper: 1.0
  threshold_absolute:
    lower: -1.0
    upper: 1.0
  gaussian_size: 1.0121397464510862 # in px
  gaussian_mode: nearest
grains:
  # Thresholding by height
  threshold_method: std_dev # Options : std_dev, otsu, absolute
  threshold_std_dev:
    lower: 1.0
    upper: 1.0
  threshold_absolute:
    lower: -1.0
    upper: 1.0
  # Thresholding by area
  absolute_smallest_grain_size: 100
  absolute_area_threshold:
    upper: [ 300,800 ] # above surface [Low, High] in nm^2 (also takes null)
    lower: [ null, null ] # below surface [Low, High] in nm^2 (also takes null)
grainstats:
  cropped_size: 60.0 # Length (in nm) of square cropped images (can take -1 for grain-sized box)
  save_cropped_grains: true # Options : true, false
dnatracing:
plotting:
  save_format: png # Options : see https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html
  image_set: all  # Options : all, core
  zrange: [0, 4]  # low and high height range for core images (can take [null, null])
  cmap: nanoscope # Options : nanoscope, afmhot
  histogram_log_axis: false # Options : true, false
```
**Expected behavior**
The entire file name should be used to create those directories, instead of just the part before the period. For example, when processing a directory with `filename.01.spm` and `filename.02.spm`, directories `filename.01` and `filename.02` should be generated, instead of `filename`, and the error can then be avoided.

**Output**

If applicable please include the output error, this can be a copy and paste of the output (preferable) or a screenshot.
Output files can be attached to this bug report.

 - OS: windows 10
 - topostats version: 5bbf692
 - Python version: 3.9.12
 - Optionally, your installed packages: best way to get this is to paste the results of typing `pip freeze`.

**Additional context**

### LINKED ISSUE #433: Exclude plotting configuration from config.yaml written after processing
**Describe the bug**

`run_topostats` writes the configuration file used at the end of processing. This includes...

```yaml
plotting:
  ...
  plot_dict:
    ...
```

...which are all the options for plotting the images and are appended to `topostats/default_config.yaml` from `topostats/plotting_dictionary.yaml`.

This means if a user attempts to modify and use the `<output_dir>/config.yaml` that a run generates the validation of the configuration fails because `config` and `plotting_dict` have their own individual Schemas and are validated independently.

The resolution would be to drop `plot_dict` from `config`.

In `run_topostats.py` there is the following section at the end of `main()`...

```python
    # Write config to file
    LOGGER.info(f"Writing configuration to : {config['output_dir']}/config.yaml")
    write_yaml(config, output_dir=config["output_dir"])
```

The `plot_dict` section can be removed using...

```python
    # Write config to file
    LOGGER.info(f"Writing configuration to : {config['output_dir']}/config.yaml")
    config["plotting"].pop("plot_dict")
    write_yaml(config, output_dir=config["output_dir"])
```

ns-rse:
Closes #362 
Closes #391 
Closes #428 
Closes #430 
Closes #433 

Pull request and code changes grew after initially making #429 (see for useful comments/feedback) but in summary...

## Double periods

`LoadScans()` class gets the filename using `Path.stem` method. However `processed_scan()` in `run_topostats.py` also took `Path.stem` so if filenames had two (or more) periods in them the filename ended up truncated and the last two sections after periods were excised (e.g. `mini.circle.spm` > `mini`). This is resolved by using `Path.name` in `processed_scan()` now instead.

There were also some problems with the `utils.get_out_path()` function which have been corrected but aren't easily visible because of having moved the function from `utils.py` to `io.py`.

## Processed Directory

* Retain a `processed` directory for _each_ level of nesting of files.
* Changed the default `image_set: core` (was `all`) so that minimal output is generated, decreasing run-time, two images are now output for each image under `processed`, one from the `z_threshed` array (aka "Height Thresholded") and one from `mask_overlay` (aka "Height Thresholded with Mask"). Images of individual grains are written to `processed/grains/<direction>/`. A file for all statistics in a given folder is written to `processed/folder_grainstats.csv` (and actually includes DNA Tracing statistics too!).

## Command line argument to run with cores

Added a simple command line argument (`-j` / `--cores`) to over-ride the number of cores used in processing (reduced default number in `default_config.yaml` because of issue with GitHub runners, this means its easy when running/testing locally to over-ride this).

`-j` because `-c` is already taken for `--config_file` and `-j` is the option the `gcc` compiler takes for the number of cores to use.

## Restructured Code

Having lots of things in a `utils` module is poor practice, it means we've not thought about code and what its doing well enough (I'm guilty of this as I created the module!). To which end I have moved some functions and their tests around.

## Linting

Resolved some linting errors, disabled some others across files that I touched. Also renamed `data2` option to the `Images()` class to `masked_array` which is more informative.

## Completion message

Added a completion message that summarises the images found, how many successfully processed (which means that some statistics were calculated for them). It includes the input directory, output directory along with key URLs, how to report bugs and a request to cite the software if used.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -113,7 +113,7 @@ def get_out_path(
 
 
 def find_images(base_dir: Union[str, Path] = None, file_ext: str = ".spm") -> List:
-    """Scan the specified directory for images with the given file extension.
+    """Recursively scan the specified directory for images with the given file extension.
 
     Parameters
     ----------
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -2,7 +2,7 @@
 import logging
 from datetime import datetime
 from pathlib import Path
-from typing import Union, Dict
+from typing import Dict, List, Union
 import numpy as np
 
 import pySPM
@@ -73,6 +73,62 @@ def write_yaml(
             f.write(yaml_dump(config_yaml))
         except YAMLError as exception:
             LOGGER.error(exception)
+
+
+def get_out_path(
+    image_path: Union[str, Path] = None, base_dir: Union[str, Path] = None, output_dir: Union[str, Path] = None
+) -> Path:
+    """Adds the image path relative to the base directory to the output directory.
+
+    Parameters
+    ----------
+    image_path: Path
+        The path of the current image.
+    base_dir: Path
+        Directory to recursively search for files.
+    output_dir: Path
+        The output directory specified in the configuration file.
+
+    Returns
+    -------
+    Path
+        The output path that mirrors the input path structure.
+    """
+    # If image_path is relative and doesn't include base_dir then a ValueError is raised, in which
+    # case we just want to append the image_path to the output_dir
+    try:
+        # Remove the filename if there is a suffix, not always the case as
+        # get_out_path is called from folder_grainstats()
+        if image_path.suffix:
+            return output_dir / image_path.relative_to(base_dir).parent / image_path.stem
+        return output_dir / image_path.relative_to(base_dir)
+    except ValueError:
+        if image_path.suffix:
+            return output_dir / image_path.parent / image_path.stem
+        return Path(str(output_dir) + "/" + str(image_path))
+    # AttributeError is raised if image_path is a string (since it isn't a Path() object with a .suffix)
+    except AttributeError:
+        LOGGER.error("A string form of a Path has been passed to 'get_out_path()' for image_path")
+        raise
+
+
+def find_images(base_dir: Union[str, Path] = None, file_ext: str = ".spm") -> List:
+    """Scan the specified directory for images with the given file extension.
+
+    Parameters
+    ----------
+    base_dir: Union[str, Path]
+        Directory to recursively search for files, if not specified the current directory is scanned.
+    file_ext: str
+        File extension to search for.
+
+    Returns
+    -------
+    List
+        List of files found with the extension in the given directory.
+    """
+    base_dir = Path("./") if base_dir is None else Path(base_dir)
+    return list(base_dir.glob("**/*" + file_ext))
 
 
 # pylint: disable=too-many-instance-attributes
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -478,7 +478,7 @@ def minicircle_grain_mask(minicircle_grain_threshold_otsu: Grains) -> Grains:
     minicircle_grain_threshold_otsu.directions["upper"] = {}
     minicircle_grain_threshold_otsu.directions["upper"]["mask_grains"] = _get_mask(
         image=minicircle_grain_threshold_otsu.image,
-        threshold=minicircle_grain_threshold_otsu.thresholds["upper"],
+        thresh=minicircle_grain_threshold_otsu.thresholds["upper"],
         threshold_direction="upper",
         img_name=minicircle_grain_threshold_otsu.filename,
     )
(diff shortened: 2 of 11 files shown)


topostats/io.py
```diff
@@ -75,6 +75,62 @@ def write_yaml(
...
+            return output_dir / image_path.parent / image_path.stem
+        return Path(str(output_dir) + "/" + str(image_path))
+    # AttributeError is raised if image_path is a string (since it isn't a Path() object with a .suffix)
+    except AttributeError:
+        LOGGER.error("A string form of a Path has been passed to 'get_out_path()' for image_path")
+        raise
+
+
+def find_images(base_dir: Union[str, Path] = None, file_ext: str = ".spm") -> List:
+    """Scan the specified directory for images with the given file extension.
```
SylviaWhittle, line 116: Very minor point but should we specify that this is recursive?

SylviaWhittle (changes requested):
Excellent, thank you @ns-rse.

Just one very nitpicky thing in the docstring of one function.

SylviaWhittle (approved):
Excellent. Thank you @ns-rse.

Issue #268, opened 2022-08-23
Convert absolute_smallest_grain_size in pixels to SI units


ns-rse:
Currently the `absolute_smallest_grain_size` set in the `grains` class for excluding noise is in pixels. It would be preferable to have this in nanometers.

ns-rse:
See also #243

bobturneruk:
This is to remove small bits of crap in the image, not digital artefacts.

MaxGamill-Sheffield:
Discussed with Alice, moving away from Milestone 1 until file formats implemented and a pixel-wise unit can possibly be extracted

MaxGamill-Sheffield:
Seems like this is in `nm` at least:
```
remove_small_objects(
                image,
                min_size=(self.minimum_grain_size * self.pixel_to_nm_scaling),
                **kwargs,
            )
```

This function comes from https://scikit-image.org/docs/stable/api/skimage.morphology.html#skimage.morphology.remove_small_objects and inputs an integer size so it may only need an int wrapper.

PR #437, opened 2023-01-16
Sylvia whittle/268 si area thresholding

SylviaWhittle:
Closes https://github.com/AFM-SPM/TopoStats/issues/268

This PR makes a few changes to area thresholding of grains, primarily to the previously named absolute_smallest_grain_size.

Rename `absolute_smallest_grain_size` to `smallest_grain_size` to hopefully avoid any confusion between this and absolute_area_threshold.
Fix the application of the scaling factor `pixel_to_nm_scaling` which was being applied incorrectly to area values.
Update the comments, including adding units in the docstrings to try to avoid confusion of the units of the area values at any stage.
The tests have had to be updated. The changes are accounted for and make sense (to me at least). Two extra grains are found in `test_run_topostats::process_scan_lower` (and `process_scan_both`), this is because of a minor shift in the `smallest_grain_size` value used in noise_removal compared to other functions during the conversion to nanometres that could not be avoided (due to different functions being differently wrong in their usage of the scaling factor)


topostats/default_config.yaml
```diff
@@ -38,9 +38,9 @@ grains:
     upper: 1.0
   direction: upper # Options: upper, lower, both (defines whether to look for grains above or below thresholds or both)
   # Thresholding by area
-  absolute_smallest_grain_size: 100
+  smallest_grain_size: 24 # Size in nm^2 of tiny grians/blobs (noise) to remove.
```
ns-rse, line 41: Might be useful to include the units in the variable name...

```suggestion
  smallest_grain_size_nm2: 24 # Size in nm^2 of tiny grians/blobs (noise) to remove.
```

topostats/grains.py
```diff
@@ -158,23 +158,40 @@ def remove_noise(self, image: np.ndarray) -> np.ndarray:
...
+            f"[{self.filename}] : Removing noise (< {self.smallest_grain_size} nm^2 \
+{self.smallest_grain_size / (self.pixel_to_nm_scaling**2):.2f} px^2)"
+        )
+        return remove_small_objects(
+            image, min_size=self.smallest_grain_size / (self.pixel_to_nm_scaling**2), **kwargs
+        )
 
     def remove_small_objects(self, image: np.array, **kwargs):
-        """Remove small objects."""
+        """Remove small objects from the input image. Threshold determined by the minimum_grain_size variable of the
```
ns-rse, line 172: Thanks for bulking out missing doc strings :+1:

topostats/grains.py
```diff
@@ -158,23 +158,40 @@ def remove_noise(self, image: np.ndarray) -> np.ndarray:
...
             small_objects_removed = remove_small_objects(
                 image,
-                min_size=(self.minimum_grain_size * self.pixel_to_nm_scaling),
+                min_size=self.minimum_grain_size,  # minimum_grain_size is in pixels squared
                 **kwargs,
             )
             LOGGER.info(
-                f"[{self.filename}] : Removed small objects (< {self.minimum_grain_size * self.pixel_to_nm_scaling})"
+                f"[{self.filename}] : Removed small objects (< \
+{self.minimum_grain_size} px^2 {self.minimum_grain_size / (self.pixel_to_nm_scaling)**2} nm^2)"
```
ns-rse, line 194: Minor formatting suggestion, not essential...
```suggestion
{self.minimum_grain_size} px^2 / {self.minimum_grain_size / (self.pixel_to_nm_scaling)**2} nm^2)"
```

topostats/validation.py
```diff
@@ -133,7 +133,7 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
         },
         "grains": {
             "run": Or(True, False, error="Invalid value in config for grains.run, valid values are 'True' or 'False'"),
-            "absolute_smallest_grain_size": int,
+            "smallest_grain_size": int,
```
ns-rse, line 136: Would need updating if variable name changed, could perhaps check that its positive

```suggestion
            "smallest_grain_size_nm": lambda n: n > 0.0,
```

topostats/grains.py
```diff
@@ -41,7 +41,7 @@ def __init__(
             "lower": [None, None],
         },
         direction: str = None,
-        absolute_smallest_grain_size: float = None,
+        smallest_grain_size: float = None,
```
ns-rse, line 44: If changing name to reflect units...

```suggestion
        smallest_grain_size_nm: float = None,
```

Docstring should be updated too but can't comment on unchanged lines.

topostats/grains.py
```diff
@@ -76,7 +76,7 @@ def __init__(
         self.absolute_area_threshold = absolute_area_threshold
         # Only detect grains for the desired direction
         self.direction = [direction] if direction != "both" else ["upper", "lower"]
-        self.absolute_smallest_grain_size = absolute_smallest_grain_size
+        self.smallest_grain_size = smallest_grain_size
```
ns-rse, line 79:
```suggestion
        self.smallest_grain_size_nm = smallest_grain_size_nm
```

topostats/grains.py
```diff
@@ -145,8 +145,8 @@ def calc_minimum_grain_size(self, image: np.ndarray) -> float:
         else:
             self.minimum_grain_size = -1
 
-    def remove_noise(self, image: np.ndarray) -> np.ndarray:
-        """Removes noise which are objects smaller than the 'absolute_smallest_grain_size'.
+    def remove_noise(self, image: np.ndarray, **kwargs) -> np.ndarray:
+        """Removes noise which are objects smaller than the 'smallest_grain_size'.
```
ns-rse, line 149:
```suggestion
        """Removes noise which are objects smaller than the 'smallest_grain_size_nm'.
```

topostats/grains.py
```diff
@@ -158,23 +158,40 @@ def remove_noise(self, image: np.ndarray) -> np.ndarray:
         Returns
         -------
         np.ndarray
-            2D Numpy array of image with objects < absolute_smallest_grain_size removed.
+            2D Numpy array of image with objects < smallest_grain_size removed.
```
ns-rse, line 161:
```suggestion
            2D Numpy array of image with objects < smallest_grain_size_nm removed.
```

topostats/grains.py
```diff
@@ -289,10 +313,12 @@ def find_grains(self):
                 self.directions[direction]["tidied_border"] = self.tidy_border(
                     self.directions[direction]["labelled_regions_01"]
                 )
+                LOGGER.info(f"[{self.filename}] : Removing noise ({direction})")
                 self.directions[direction]["removed_noise"] = self.area_thresholding(
                     self.directions[direction]["tidied_border"],
-                    [self.absolute_smallest_grain_size * self.pixel_to_nm_scaling, None],
+                    [self.smallest_grain_size, None],
```
ns-rse, line 319:
```suggestion
                    [self.smallest_grain_size_nm, None],
```

topostats/grains.py
```diff
@@ -202,10 +220,16 @@ def area_thresholding(self, image: np.ndarray, area_thresholds: list):
             upper = image.size * self.pixel_to_nm_scaling**2
         if lower is None:
             lower = 0
+        # Get array of grain numbers (discounting zero)
         uniq = np.delete(np.unique(image), 0)
         grain_count = 0
-        for grain_no in uniq:
+        LOGGER.info(
+            f"[{self.filename}] : Area thresholding grains | Thresholds: L:{lower / self.pixel_to_nm_scaling**2:.2f},\
```
ns-rse, line 227: You can split long log messages by just closing the string then having another quoted string straight after (see `validation.py` for examples). Saves having to escape carriage returns and text remains readable.

ns-rse (changes requested):
Some quite minor suggestions to clarify that units are in `_nm` otherwise good to go.

topostats/grains.py
```diff
@@ -158,23 +158,40 @@ def remove_noise(self, image: np.ndarray) -> np.ndarray:
...
-        return remove_small_objects(image, min_size=self.absolute_smallest_grain_size)
+        LOGGER.info(
+            f"[{self.filename}] : Removing noise (< {self.smallest_grain_size_nm2} nm^2"
+            "{self.smallest_grain_size_nm2 / (self.pixel_to_nm_scaling**2):.2f} px^2)"
+        )
+        return remove_small_objects(
+            image, min_size=self.smallest_grain_size_nm2 / (self.pixel_to_nm_scaling**2), **kwargs
+        )
 
     def remove_small_objects(self, image: np.array, **kwargs):
```
ns-rse, line 171: I hadn't noticed this before but there is potential here for conflict since we have imported `remove_small_objects` from skimage and this redefines that name, albeit as a method to the `Grains` class.

I noticed it reading the function and thinking "_Why is this function using recursion?_" then checking what had been imported.

Sorry to be a pain but I think we could perhaps avoid this by instead...

```
from skimage import morphology
```

and then calling `morphology.remove_small_objects` on line 187 and on line 128 using `morphology.label()`. This is the suggestion I read about in [this article](http://blog.nicholdav.info/four-tips-structuring-research-python/)

ns-rse (changes requested):
Thanks for making all the variables even clearer @SylviaWhittle 

Sorry to be a pain but I noticed one issue with one of the method names I'd not clocked previously and have suggested a way round this.

topostats/grains.py
SylviaWhittle: I agree! I noticed this while writing this PR and forgot to address it.

ns-rse (approved):
Cool, thanks for sorting that so quickly @SylviaWhittle looks ready to merge.

PR #445, opened 2023-01-23
Sylvia whittle/243 si output stats

### LINKED ISSUE #243: Grainstats and DNA Tracing Stats in SI Units
Currently units are in nm but it might be best to be in SI to make further analyses easier and consistent.

Swap the grainstats file outputs to SI units.

Make this configurable in the config file but this is a bit unnecessary.

**Additional context**
N/A

### LINKED ISSUE #420: Remove extraneous plotting from grainstats.py
`GrainStats()` currently has lines that are used to plot the data, but no image is saved (see lines [158-160](https://github.com/AFM-SPM/TopoStats/blob/main/topostats/grainstats.py#L158)).

These are likely from development/debugging and weren't picked up during review of the PR.

SylviaWhittle:
Closes #243 

This PR attempts to accurately convert the units of both output grain stats and dnatracing stats in SI standard scientific form, as was requested by #243, though please do shout if I've missed something here in the MO of #243.

The changes are as follows:

- added a multiplier of `1e-9` to all length measurements in `grainstats.py`.
- added a multiplier of `1e-9**2` to all area measurements in `grainstats.py`.
- added a multiplier of `1e-9**3` to the volume measurement in `grainstats.py`.
- removed the multipliers of `1e+9` from the length measurements in `dnatracing.py`.
- updated the default config larger area threshold from `192` to `300` to process `minicircle.spm` as it did before #437. (Some grains were not marked due to the area thresholding).
- updated tests accordingly. 

Closes #420 too.


topostats/default_config.yaml
```diff
@@ -38,9 +38,9 @@ grains:
     upper: 1.0
   direction: upper # Options: upper, lower, both (defines whether to look for grains above or below thresholds or both)
   # Thresholding by area
-  smallest_grain_size_nm2: 24 # Size in nm^2 of tiny grians/blobs (noise) to remove.
+  smallest_grain_size_nm2: 50 # Size in nm^2 of tiny grians/blobs (noise) to remove.
   absolute_area_threshold:
-    upper: [ 120, 192 ] # above surface [Low, High] in nm^2 (also takes null)
+    upper: [ 300, 3000 ] # above surface [Low, High] in nm^2 (also takes null)
```
ns-rse, line 43: Thanks for correcting this, we really need to pin the values in `default_config.yaml` and leave them alone (see #302 where I was hoping to have such discussion on default values, might be another issue too).

If tests need different values the configuration can be updated after having loaded it to dictionary in the test suite.

topostats/grainstats.py
```diff
@@ -241,28 +241,31 @@ def calculate_stats(self) -> Dict:
...
+                "radius_min": radius_stats["min"] * self.pixel_to_nanometre_scaling * 1e-9,
+                "radius_max": radius_stats["max"] * self.pixel_to_nanometre_scaling * 1e-9,
+                "radius_mean": radius_stats["mean"] * self.pixel_to_nanometre_scaling * 1e-9,
+                "radius_median": radius_stats["median"] * self.pixel_to_nanometre_scaling * 1e-9,
+                "height_min": np.nanmin(masked_grain_image) * 1e-9,
+                "height_max": np.nanmax(masked_grain_image) * 1e-9,
+                "height_median": np.nanmedian(masked_grain_image) * 1e-9,
+                "height_mean": np.nanmean(masked_grain_image) * 1e-9,
+                "volume": np.nansum(masked_grain_image)
+                * self.pixel_to_nanometre_scaling**2
```
ns-rse, line 255: I suspect its `black` at play here but had to pause and think through why these being on their own line  weren't causing errors, its because whatever is between `:` and the next `,` is evaluated and line breaks ignored.

Maybe moving the comment at the end of the line to above `"volume"` might prevent `black` from wrapping it and make it fractionally easier to read.

ns-rse (changes requested):
Some very minor points, one relates to #420 and moving plotting out, perhaps if the `self.pixel_to_nanometer_scaling * 1e-9` is to be done once then modifying the plotting to account for this for the time being until it can be moved out might one way to handle.

Also, and I can't think there ever will be, is there any possibility of the scaling being something other than `1e-9`? I doubt it since that is `m > nm` but would any instruments ever use say `cm` and the scaling need to be done `cm > nm` (in turn requiring `1e-8` if I've understood correctly).

If so making this a configurable option across the config file and functions now would be sensible and avoid having to make changes in the future.

topostats/default_config.yaml
SylviaWhittle: Good point about fixing the values in the test. I do think these values are better defaults though, my mistake for setting the worse ones in the previous PR. 

I don't think that I should be the sole decider on the default value though, I'll try to get some experimentalist input in #302.

topostats/grainstats.py
SylviaWhittle: Still splits it, sorry. If you have any suggestions on how to make it easier to read, I would be eager to try them!

MaxGamill-Sheffield:
Just hopping on this and what @ns-rse just mentioned that the we don't know what units different file formats might take. As a result I think it's wisest to make *everything* SI units (changing `px_2_nm` to `px_2_m`, and move all the rescaling into `io.py` as the files are loaded?

What do people think?

SylviaWhittle:
Just a note, experimentalists (@derollins) say that they would prefer that the plotting output scales remain in nanometres.

SylviaWhittle:
@MaxGamill-Sheffield I agree, I was just experimenting with this as you posted!

ns-rse:
Yep, making things configurable and deriving from the input files seems the most logical and consistent approach that would also mean the least hassle.

MaxGamill-Sheffield:
> Just a note, experimentalists (@derollins) say that they would prefer that the plotting output scales remain in nanometres.

@SylviaWhittle I think I was playing with some sort of "smart scales" at some point, where the image scale plotted was identified from the px_2_m and largest pixel shape. Maybe this could help declutter the config too?

ns-rse:
Sure, if values can be derived systematically from code/file-headers its the way to go rather than users having to specify values or write additional code to deduce them.


One thing to be aware is the issues that can arise when working with floating points, a good exposition of which is from [Julia Evans blog post on the issue](https://jvns.ca/blog/2023/01/13/examples-of-floating-point-problems/).

To which end if integers can be used throughout and conversion only done at the end, as appears to be being done here in `grainstats` and `dnatracing` that may avoid some problems (although `int`  [have their own problems](https://jvns.ca/blog/2023/01/18/examples-of-problems-with-integers/)).

topostats/grainstats.py
```diff
@@ -241,28 +241,31 @@ def calculate_stats(self) -> Dict:
...
-                "height_min": np.nanmin(masked_grain_image),
-                "height_max": np.nanmax(masked_grain_image),
-                "height_median": np.nanmedian(masked_grain_image),
-                "height_mean": np.nanmean(masked_grain_image),
-                "volume": np.nansum(masked_grain_image) * self.pixel_to_nanometre_scaling**2,
-                "area": region.area * self.pixel_to_nanometre_scaling**2,
-                "area_cartesian_bbox": region.area_bbox * self.pixel_to_nanometre_scaling**2,
-                "smallest_bounding_width": smallest_bounding_width * self.pixel_to_nanometre_scaling,
-                "smallest_bounding_length": smallest_bounding_length * self.pixel_to_nanometre_scaling,
+                "centre_x": centre_x * self.pixel_to_nanometre_scaling * 1e-9,
```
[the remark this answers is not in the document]
SylviaWhittle: Moved the multiplication to up-front. 👍 

Also removed one of the extraneous plots, however I think that your suggestion of plotting the grains might require some extra values being added to `grainstats` DataFrame as in order to do the plotting we need various values pertaining to bounding box of the grains, namely `minr, minc, maxr, maxc = region.bbox`. I think the grain mask obtained and used internally to `calculate_grainstats` will need to be passed out too, as we plot sections of that. 

Do you have any suggestions about this?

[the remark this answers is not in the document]
ns-rse: I was thinking that grainstats should either have a property or return a dictionary of all the data that is required to plot grains. This can then optionally be processed (via dictionary comprehension) to use the classes in `plottingfuncs.py` to plot and save grain level images if required.

My reasoning being that plotting is a distinct action from calculating statistics on grains. If not treated as separate plotting should perhaps be done at the grain detection stage rather than under the statistics calculation, from memory the bounding boxes and other things that isolate grains are all derived via scikit-image's morphology.

Let me know what you think and happy to talk this through in person.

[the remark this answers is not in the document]
SylviaWhittle: Thanks! I'll give it a go 👍

tests/test_grainstats_minicircle.py
```diff
@@ -19,7 +19,7 @@ def test_grainstats_regression(regtest, minicircle_grainstats: GrainStats) -> No
     """Regression tests for grainstats."""
     statistics = minicircle_grainstats.calculate_stats()
     print(statistics)
```
ns-rse, line 21: Pedantic but this line can be replaced, its likely a debugging leftover (probably mine!) as its the next line that saves the file using the `pytest-regtest` plugin.

ns-rse: I've removed this and pushed to this branch.

ns-rse (approved):
Sorry for leaving the `print()` statement in, removed that, should be good to go.

ns-rse (commented):
Not too awkward to get the grains saved.

I was thinking though the data is coming from the `labelled_data` which is an artifact of the `grains` stage. To reduce the size of this class and the number of things it does, particularly the `calculate_stats()` method, the processing/extraction of these could perhaps be either their own method (possibly even in `Grains()` class) or a function all of its own.

tests/test_run_topostats.py
```diff
@@ -74,6 +76,82 @@ def test_process_scan_both(regtest, tmp_path, process_scan_config: dict, load_sc
...
+    ],
+)
+def test_save_cropped_grains(
+    tmp_path: Path, process_scan_config: dict, load_scan_data: LoadScans, image_set, expected
+) -> None:
+    """Tests if cropped grains are saved only when image set is 'all' rather than 'core'."""
+    process_scan_config["plotting"]["image_set"] = image_set
+    process_scan_config["plotting"] = update_plotting_config(process_scan_config["plotting"])
+
+    print(f"image set: {process_scan_config['plotting']['image_set']}")
```
ns-rse, line 93: Debugging line hang-over.

tests/test_run_topostats.py
```diff
@@ -74,6 +76,82 @@ def test_process_scan_both(regtest, tmp_path, process_scan_config: dict, load_sc
...
+
+    img_dic = load_scan_data.img_dic
+    _, _ = process_scan(
```
ns-rse, line 96: Brilliant, having this test `process_scan()` is going to make it nice and easy for me to move tests around when I rebase on `main` because as I mentioned I've moved `process_scan()` out to its own sub-module. :+1:

topostats/validation.py
```diff
@@ -579,7 +572,7 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                     "Invalid value in config 'grain_image.image_type', valid values " "are 'binary' or 'non-binary'"
                 ),
             ),
-            "core_set": True,
+            "core_set": False,
```
ns-rse, line 575: Might need to update `docs/configuration.md` to reflect this default value, sorry.

topostats/run_topostats.py
```diff
@@ -388,22 +400,8 @@ def main(args=None):
         config["plotting"]["plot_dict"], schema=PLOTTING_SCHEMA, config_type="YAML plotting configuration file"
     )
 
-    # FIXME : Make this a function and from topostats.utils import update_plot_dict and write tests
```
ns-rse, line 391: Thank you for fixing my long-standing `FIXME`, very much appreciated :smiley_cat:

topostats/utils.py
```diff
@@ -90,6 +90,27 @@ def update_config(config: dict, args: Union[dict, Namespace]) -> Dict:
     return config
 
 
+def update_plotting_config(plotting_config: dict) -> dict:
```
ns-rse, line 93: Minor but do you think it might be possible to write a parametrized test for this new function? Could no doubt use the existing `topostats/plotting_config.yaml` and check values are updated correctly by looking at one or two image configurations, say one that should be updated and one that shouldn't.

topostats/grainstats.py
```diff
@@ -148,23 +147,30 @@ def is_clockwise(p_1: tuple, p_2: tuple, p_3: tuple) -> bool:
         return not np.linalg.det(rotation_matrix) > 0
 
     def calculate_stats(self) -> Dict:
-        """Calculate the stats of grains in the labelled image"""
+        """Calculate the stats of grains in the labelled image.
```
ns-rse, line 150: Thanks for bulking out the docstring here. :+1:

ns-rse (changes requested):
Looks great, although its grown somewhat from the original :wink: 

If you don't mind writing a short test to check the new `update_plotting_config()` function that would be great.

One value has changed in the `default_config.yaml` and we should therefore update `docs/configuration.md` to keep track of these changes.

tests/test_utils.py
```diff
@@ -34,6 +35,16 @@ def test_update_config(caplog) -> None:
     assert updated_config["c"] == "something new"
 
 
+def test_update_plotting_config(process_scan_config: dict) -> None:
```
ns-rse, line 38: I was thinking this might be paramterised along the lines of...


```suggestion
@pytest.mark.parametrize("image_name,core_set,title,zrange",
[("extracted_channel",False,"Raw Height",KeyError()),
 ("z_threshed",True,"Height Threshold",*id001) 
def test_update_plotting_config(process_scan_config: dict, image_name: str, core_set: bool, title:str) -> None:

    process_scan_config["plotting"] = update_plotting_config(process_scan_config["plotting"])
    assert process_scan_config["plotting"]["plot_dict"][image_name]["core_set"] == core_set
    assert process_scan_config["plotting"]["plot_dict"][image_name]["title"] == title

```

Could have more values tested but `z_threshed`  is one of only two variables that should have a value for `zrange` the others all have that `.pop()`'ed off. That said the value `*id001` looks kind of strange :thinking: 

Hope that makes sense, let me know if not.

SylviaWhittle: I believe I have successfully parameterized them now. The test runs successfully on my machine, and it's late so I'm going to cross my fingers and get some sleep.

Let me know if you're unhappy with the way I handled the `KeyError` exception, I didn't know how else I could handle it (using `KeyError()` as an argument didn't seem to work).

Thanks for your speedy feedback, it's really appreciated.

ns-rse (approved):
Its a ballooned a bit from the initial review but all the changes are super useful, especially pulling out the plotting of grains from `GrainStats()` class and fixing my dated `FIXME` for updating the plotting config as a function, thank you. :+1: 

I think this is ready to merge now, I can then rebase my plotting branch on it and hopefully get a PR out later for merging that (I suspect I may have some merge conflicts to resolve though because of the changes I've made!)

PR #449, opened 2023-01-31
Fixes #447 extra column in CSV

### LINKED ISSUE #403: Add molecule number in the all_statistics.csv spreadsheet

### LINKED ISSUE #447: CSV file structure breaks with some images
**Describe the bug**
With some images, the file structure of the `folder_grainstats.csv` file breaks (incorrectly placed headers and commas). 

image

I think this might be something to do with detection of grains but then subsequent removal before grainstats outputs data on them? I apologise for not looking into this further, as I don't have the time currently. I will come back to this to add information.

**To Reproduce**

- Run topostats with any config on the following set of `.spm` files in the following folder. https://drive.google.com/drive/folders/1dflIB-qEGMahmdLoVXwDC_dm9Xksnv1h

Note that the `folder_grainstats.csv` file is incorrectly formatted.

- Config file

```
base_dir: ./by_sample/A4
output_dir: ./by_sample/A4
#base_dir: ./raphael_difficult_input/A3_MgCl2_008
#output_dir: ./raphael_difficult_output/A3_MgCl2_008_tilt_first
#base_dir: ./comparisons/slide_2_blue
#output_dir: ./comparisons/slide_2_blue
warnings: ignore
cores: 6
quiet: false
file_ext: .spm
loading:
  channel: Height
filter:
  threshold_method: std_dev # Options : otsu, std_dev, absolute
  threshold_std_dev:
    lower: 10.0
    upper: 1.3
  threshold_absolute:
    lower: -1.0
    upper: 1.0
#  gaussian_size: 1.0121397464510862 # in px
  gaussian_size: 0.5
  gaussian_mode: nearest
  remove_scars:
    run: true
    removal_iterations: 2
    threshold_low: 0.250
    threshold_high: 0.666
    max_scar_width: 4
    min_scar_length: 10
grains:
  # Thresholding by height
  threshold_method: std_dev # Options : std_dev, otsu, absolute
  threshold_std_dev:
    lower: 10.0
    upper: 1.3
  threshold_absolute:
    lower: -1.0
    upper: 1.0
  # Thresholding by area
  smallest_grain_size_nm2: 5000
  absolute_area_threshold:
    upper: [ 5000, 10000000 ] # above surface [Low, High] in nm^2 (also takes null)
    lower: [ null, null ] # below surface [Low, High] in nm^2 (also takes null)
grainstats:
  cropped_size: 500 # Length (in nm) of square cropped images (can take -1 for grain-sized box)
  save_cropped_grains: true # Options : true, false
dnatracing:
plotting:
  save_format: png # Options : see https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html
  image_set: core  # Options : all, core
  zrange: [-3, 4]  # low and high height range for core images (can take [null, null])
  cmap: nanoscope # Options : nanoscope, afmhot
  mask_cmap: blu
  histogram_log_axis: false # Options : true, false
  histogram_bins: 200 # Number of bins for histogram plots to use

```

**Expected behavior**
A properly formatted output `csv` file.

 - OS: MacOS
 - topostats version: `main` commit: [c9a6fd9](https://github.com/AFM-SPM/TopoStats/commit/c9a6fd99c311c468e3f5783db78e17d74a140af7)
 - Python version: 3.10

ns-rse:
Closes #447

* Fixes additional column in CSV output when images have no grains detected. This was caused by the empty dummy data frame that is generated when no grains are found not having the index set to `molecule_number` (nee `Molecule Number` see point two below on renaming) and so when concatenated via `pd.concat()` an additional column was introduced.
* Moves `utils.folder_grainstats()` to `io.folder_grainstats()`.
* Consistent case/format for variable names in CSV (all lower snake-case).
* CSVs now include `molecule_number` Closes #403

This will fail tests because it impacts regression tests. I've recent switched to using [Git Worktrees](https://medium.com/ngconf/git-worktrees-in-use-f4e516512feb) which seems to impact my ability to run tests. I'll run these elsewhere and update the tests in a subsequent commit.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -100,7 +100,7 @@ def get_out_path(
     # case we just want to append the image_path to the output_dir
     try:
         # Remove the filename if there is a suffix, not always the case as
-        # get_out_path is called from folder_grainstats()
+        # get_out_path is called from save_folder_grainstats()
         if image_path.suffix:
             return output_dir / image_path.relative_to(base_dir).parent / image_path.stem
         return output_dir / image_path.relative_to(base_dir)
@@ -133,7 +133,9 @@ def find_images(base_dir: Union[str, Path] = None, file_ext: str = ".spm") -> Li
     return list(base_dir.glob("**/*" + file_ext))
 
 
-def folder_grainstats(output_dir: Union[str, Path], base_dir: Union[str, Path], all_stats_df: pd.DataFrame) -> None:
+def save_folder_grainstats(
+    output_dir: Union[str, Path], base_dir: Union[str, Path], all_stats_df: pd.DataFrame
+) -> None:
     """Saves a data frame of grain and tracing statictics at the folder level.
 
     Parameters
@@ -177,7 +179,6 @@ class LoadScans:
         img_paths: list,
         channel: str,
     ):
-
         """Initialise the class.
 
         Parameters
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -6,7 +6,7 @@ import numpy as np
 import pandas as pd
 import pytest
 
-from topostats.io import read_yaml, find_images, get_out_path, folder_grainstats, LoadScans
+from topostats.io import read_yaml, find_images, get_out_path, save_folder_grainstats, LoadScans
 
 BASE_DIR = Path.cwd()
 RESOURCES = BASE_DIR / "tests" / "resources"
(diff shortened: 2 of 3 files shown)


topostats/io.py
```diff
@@ -131,6 +133,41 @@ def find_images(base_dir: Union[str, Path] = None, file_ext: str = ".spm") -> Li
     return list(base_dir.glob("**/*" + file_ext))
 
 
+def folder_grainstats(output_dir: Union[str, Path], base_dir: Union[str, Path], all_stats_df: pd.DataFrame) -> None:
```
SylviaWhittle, line 136: Suggestion: rename `folder_grainstats` to `save_folder_grainstats`?

Ignore if you disagree!

SylviaWhittle (changes requested):
This works for me on a previously failing dataset. Thank you @ns-rse!

One minor suggestion, resolve the comment if you disagree and I can approve this PR.

Previously:
Pasted Graphic

Now:
Pasted Graphic 2

SylviaWhittle (approved):
Thanks for the edit!

PR #455, opened 2023-02-08
Refactoring plotting (and more)

### LINKED ISSUE #227: Refactor Plotting.py
`topostats/Plotting.py` would benefit from refactoring. Ideally it should be a class with callable methods and would benefit from...

* Use of `pathlib` for filename and directory operations.
* Correct use of logging rather than print statements.

Outstanding issues that mention Plotting...

* #29
* #56 
* #57 
* #87 
* #138 
* #205 
* #224 
* See also comment from @Jean-Du [here](https://github.com/AFM-SPM/TopoStats/issues/224#issuecomment-1189390563)

Something to consider is to move away from running `Plotting` as a script that one tries to configure (although there should be sensible defaults to produce starting points) and instead allow users to run Jupyter Notebooks to produce the exact plots they want, as this would allow users to tweak things as they desire. No template Notebooks are required, but good documentation on how to use the Class would be essential.

### LINKED ISSUE #451: Remove unnecessary images from tests
Tests take ages to run, I introduced comparing Matplotlib generated images to many of the unittests but I think there are too many in there now.

We should leave those in for explicit testing of `plottingfuncs` and `plotting` (work in progress!) but all the other arrays are checked by their sums and so the graphs themselves are a bit unnecessary.

This would also reduce the size of the repository to download (although I fear they will remain buried in the Git history unless I undertake some attach to remove them, not sure I could do that without breaking things).

### LINKED ISSUE #452: Add flake8-print to test suite
The [flake8-print](https://pypi.org/project/flake8-print/) extension adds checks for `print()` statements in the code.

These are widely used when debugging (although we should hopefully be able to do things properly with `LOGGGER.debug()` once the issues with #435 are resolved). They are also used by `pytest-regtest` to print out the results and compare them to expected. Beyond that any `print()` statements are extraneous and shouldn't make it into the code base (I'm very guilty of this!).

Add the following to `setup.cfg`

```
[options.extras_require]
...
dev =
  black
  pre-commit
  pylint
  flake8
  flake8-print
```

#446 would also deal with this as Ruff includes flake8-print and it will also automatically fix it (but we'll have to be careful not to remove the `pytest-regtest` prints.

ns-rse:
Closes #227
Closes #451 
Closes #452

## Refactoring plotting.py

This is the initial refactoring of `plotting.py`. Because of the vast scope for variation that exists in generating plots a deliberate attempt to keep a sensible and minimal set of plots has been taken whilst refactoring the code to be as succinct as possible. Originally there was a mixture of Matplotlib and Seaborn functions, to maintain simplicity the refactoring uses only Seaborn.

* Histograms and Kernel Density plots can be generated (if both are requested then the KDE is overlaid).
* Violin plots can be generated.
* For either type of plot if there is more than one unique `image` in the dataset the data is plotted separately on the same graph (i.e. different colour histograms/kde for each image; multiple violins, one for each image).
* If data comes from more than one image then these are plotted separately by default (currently there is no option to combine them) and Seaborn's `hue` option is used to differentiate these. Because this `hue` option is always used, even with a single image, it overrides the `color` option so it is not possible to explicitly request the color of plots.
* Configuration is via a new YAML file `topostats/summary_config.yaml`. When running as part of `run_topostats` output is to the `default_config.yaml` `output:` location but in the sub-directory `summary_distributions/`.
  * Output file type can be controlled via the `file_ext:` argument.
  * Output file names are constructed as `<statistic>_<plot_type>`.
  * Statistics to plot are controlled by (un-)commenting the variables listed under `stats_to_sum` in the `sumary_config.yaml`.
* Variable names (which are currently aiming to all be `lower_snake_case`) are automatically mapped to variable labels via `topostats/var_to_label.yaml`. New statistics/variables when introduced should be added to this file to ensure plots are labelled correctly.
* Plotting will be done automatically when running TopoStats unless `default_config.yaml` has the option `summary_stats: run: false`. Unless an explicit configuration file for summary_stats (i.e. `summary_config.yaml`) is specified in this section of `default_config.yaml` then the default `topostats/summary_config.yaml` is used.
* An "entry-point" is provided to run plotting retrospectively on generated CSV files. It also includes the option `--create_config_file` that, in a similar vein to `run_topostats --create_config_file` will write a sample YAML configuration file to the specified location for subsequent editing. It can then be passed to `toposum --config_file custom_config.yaml` allowing the user to have an easy way to toggle configuration options.
* An option exists to save the all generated images as [Python
pickles](https://realpython.com/python-pickle-module/). This can then be loaded in a Jupyter Notebook and the images further modified.
* Documentation has been updated under `configuration.md` and `usage.md`.
* A sample Jupyter Notebook has been introduced (`notebooks/01-Summary-statistics-and-plots.ipynb`) that shows how to load a CSV file of statistics and use Pandas and Seaborn to create these images. I think it would be really useful to see this grow with examples of possible plots that may be generated. The file is version controlled and can therefore be
added, but its important to restart the kernel and clear cells before committing otherwise output (text and graphs) are included.

### House Keeping

* Introduces `flake8-print` to the `dev` requirements so that extraneous `print()` statements from debugging are detected by `pre-commit` and removed before Pull Requests are made (#452).
* Removes `@pytest.mark.mpl_image_compare()` from `test-*-minicircle.py` tests. The various steps in processing assess the Numpy arrays for consistency by taking their `sum()` whilst I initially incorporated `pytest-mpl` to compare the plots these would generate that was perhaps over-zealous and unnecessary. It also results in a lot of minor changes in
borders requiring the images to be regenerated (see #451). They have therefore been removed along with the associated `tests/resources/img/*.png` files. Tests for `test_plottingfuncs.py` remain in place with Matplotlib graphs generated.
* Adds `.git-blame-ignore-revs` should we need to ignore blame (see [Who's to
blame](https://rse.shef.ac.uk/blog/git-blame/) and includes a commit made automatically by `pre-commit.ci` which applied `black` to existing files. We don't want the blame to be attributed to a CI script.

### Moving/renaming functions

* `topostats/run_topostats.py` was getting big and unwieldy. The `process_scan()` function has therefore been moved to its own sub-module `topostats/processing.py` and is imported to `run_topostats.py` for usage. Tests have been moved as is appropriate.
* Renamed `find_images()` > `find_files()` since it was requested that the new `toposum` command be able to load multiple CSV files (a feature not yet implemented).
* Command line options to `run_topostats` that are highly unlikely to be used have been removed.

## ToDo (maybe?)

Rather than this work growing any larger I've deliberately left some (small?) tasks undone, maybe not all of them need doing though. Once this PR is merged these can be discussed and written up as individual issues and addressed.

* **A per image summary of statistics**. This is pretty straight-forward, the `self.df` is a Pandas Dataframe to summarise it you can `self.df.groupby("image").describe()` and a summary dataframe is returned. Might not include all metrics (e.g. median, inter-quartile range) but they can also be generated in a similar manner and merged.
* **Automatically Adding Units** - Most units are `nm`, but area and volume differ as to statistics of curvature. Probably the easiest approach will be to modify the labels in `var_to_label.yaml` so that titles carry the correct unit. However it may be desirable to have axes just show the unit and the title not by default. If this is the case the expanding `var_to_label.yaml` so that each `var` has a nested dictionary of `title` and `units` will be the most sensible and cleanest approach.
* **Multiple CSV files** - the entry-point `toposum` should have additional functionality added to scan a `base_dir` for CSV files and these should be read in and combined into one Pandas DataFrame and passed on to the `toposum()` function. This will require adding a `base_dir: ./` option to the configuration file (`summary_config.yaml`) and associated validation to `topostats/validation.py`. The existing `topostats.io.find_files()` function (nee
`find_images()` see below) should be used to find CSV files. Once a list of files exists its possible to use dictionary comprehension in conjunction with `pd.concat()` to read and append all files in one go (e.g. `pd.concat([pd.read_csv(x) for x in csv_files])`).
* **Plotting all molecules across grains** This is another variation that would require a more logic and duplication of code. I think if this is desired it would be best suited to generating via Jupyter Notebooks.
* **Examples of modifying pickled images** It might have some utility to show how to load the pickled images. But this would be minor as its straight-forward but really if someone is going to tweak existing figures they would be best served doing so from scratch based on the examples in the Jupyter Notebook.
* **Joint plots** previous incarnation of `plotting.py` used `sns.jointplot()` to plot two different variables on the same graph. Because of the way in which variables are selected for plotting I think this could be challenging to code up. A list of what to plot is included, how does one then get a list of pairs to plot without introducing another list of which pairs? One could plot all pair-wise combinations but in many instances that may not make sense (e.g. plotting `nm` and `nm^3` on the same plot as the scales are different). This is a case where the requirements are quite specific to the person doing the work and as we can't anticipate all possible combinations Jupyter Notebooks should be used.
* **Seaborn Configuration** - Add configuration options for users to specify the [context](https://seaborn.pydata.org/generated/seaborn.set_context.html) and font via [theme](https://seaborn.pydata.org/generated/seaborn.set_theme.html).


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# pre-commit.ci - ignore updates to codebase made by pre-commit.ci
+5244539071197ce9accfd378bd9f5e557bbc0a49
```
SylviaWhittle, line 2: No new line at end of file. Not sure if this is needed, but GitHub flagged this for me.

docs/usage.md
```diff
@@ -97,6 +97,13 @@ run_topostats
 ...
 ```
 
+On a successful completion you should see a message similar to the following which indicates various aspects of the run
+along with information about how to give feedback, report bugs and cite the software.
+
+``` bash
+
```
SylviaWhittle, line 104: Forgot to add example output statement here, or perhaps this was meant to be removed, since you give an example completion message on lines `191` to `214`.

topostats/plotting.py
```diff
@@ -1,650 +1,362 @@
...
+        """
+        return f"{self.stat_to_sum}_{plot_suffix}"
+
+    def sns_plot(self) -> None:
+        """Plot the distribution of one or more statistics as either histogram, kernel density estimates or both. Uses
+        base Seaborn."""
+        fig, ax = self._setup_figure()
+        if self.hist and not self.kde:
+            outfile = self._outfile("hist")
+            sns.histplot(data=self.melted_data, x="value", bins=self.bins, stat=self.stat, hue=self.image_id)
```
SylviaWhittle, line 172: When running `run_topostats` with default config and data, the histograms do not show bars, I tried editing the histogram 'bins' value between 5 and 200, and this did not fix it. I will investigate further once I have finished this review.

SylviaWhittle (changes requested):

SylviaWhittle:
In summary_config.yaml, banding_angle is a commented-out stat, but has no other references in the codebase aside from in configuration.md, and validation.md. Is this a stat that is yet to be implemented, or possibly an error?

topostats/summary_config.yaml
```diff
@@ -0,0 +1,42 @@
...
+xrange:
+  - 0
+  - 2
+alpha: 0.5
+palette: bright # Seaborn color palette. Options : colorblind, deep, muted, pastel, bright, dark, Spectral, Set2
+stats_to_sum:  # A list of variables (key) to plot uncomment those that are required.
+  - area
+  - area_cartesian_bbox
+#  - aspect_ratio
+#  - banding_angle
```
SylviaWhittle, line 26: I cannot see any other references in the codebase to `banding_angle` aside from in `configuration.md`, `summary_config.yaml`, `validation.md`, and `var_to_label.yaml`. Is this statistic yet-to-be implemented?

SylviaWhittle (changes requested):
Sorry for the piecewise review. Let me know if doing it this way is an issue for you.

topostats/summary_config.yaml
ns-rse: This is a statistic that @Jean-Du has on one of her branches, including it now as she let me know about it already.

.git-blame-ignore-revs
ns-rse: GitHub doesn't like it, but I don't think its necessary, Git doesn't seem to mind.

Jean-Du:
Hi Sylvia,It's supposed to be "bending_angle" and it's a feature I have been working on in my bending angle branch. I have requested Neil to add it as I need to plot statistics for that.On 8 Feb 2023, at 14:53, Sylvia Whittle ***@***.***> wrote:﻿
In summary_config.yaml, banding_angle is a commented-out stat, but has no other references in the codebase aside from in configuration.md, and validation.md. Is this a stat that is yet to be implemented, or possibly an error?

topostats/plotting.py
ns-rse: Hmm, something has gone awry during the merging of `main`, they were working and I forgot to check.

The tests work as these are the test images (which pass locally, bet they're failing on CI!)

image

I'll investigate.

topostats/summary_config.yaml
SylviaWhittle: Thanks for the clarification @Jean-Du 😄

SylviaWhittle: Rename to `bending_angle`

topostats/var_to_label.yaml
```diff
@@ -0,0 +1,35 @@
...
+height_mean: Mean Height
+height_median: Median Height
+height_min: Minimum Height
+radius_max: Maximum Radius
+radius_mean: Mean Radius
+radius_median: Median Radius
+radius_min: Minimum Radius
+smallest_bounding_length: Smallest Bounding Length
+smallest_bounding_width: Smallest Bounding Width
+volume: Volume / $\mathregular{nm^3}$
```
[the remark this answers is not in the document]
ns-rse: Good catch, thanks :+1:

topostats/summary_config.yaml
ns-rse: Corrected spelling.

topostats/plotting.py
ns-rse: Ahha! This is because of the SI units which are now all `e-08` or smaller and so the `xrange` (0-2) is valid, previously they were rescaled as values were outside of the user specified `0-2` range.

I've removed the ability to explicitly specify `xrange` I think its better to let this be determined from the data (which was done anyway if the range wasn't large enough).

notebooks/01-Summary-statistics-and-plots.ipynb
```diff
@@ -0,0 +1,265 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3ae80090-71d3-4c87-9907-fd10b892c962",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df_three_images.index\n"
```
SylviaWhittle, line 239: Is this code block intended to be left in? I'm not sure why it would be printed unless you want to highlight something about the dataframe? Apologies if I'm misunderstanding, just wanted to make sure.

It prints as this:

```
MultiIndex([('smaller', 'upper', 0.0),
            ('smaller', 'upper', 1.0),
            ('smaller', 'upper', 2.0),
            ('smaller', 'upper', 3.0),
            ('smaller', 'upper', 4.0),
            ('smaller', 'upper', 5.0),
            ('smaller', 'upper', 6.0),
            ('smaller', 'upper', 7.0),
            ('smaller', 'upper', 8.0),
            ('smaller', 'upper', 9.0),
            ...
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan),
            (      nan,     nan, nan)],
           names=['image', 'threshold', 'molecule_number'], length=105)
```

notebooks/01-Summary-statistics-and-plots.ipynb
```diff
@@ -0,0 +1,265 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "cf4d92f0-b624-452c-9f51-52fe70dfde88",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df[df[\"circular\"] == False][\"end_to_end_distance\"].plot.hist(figsize=(16,9),\n",
```
SylviaWhittle, line 130: Perhaps add a comment here stating that we ignore entries that are marked as "circular" since they do not have end-to-end distances? Minor suggestion, feel free to ignore.

SylviaWhittle (changes requested):
I've not looked through this as thoroughly as I would like just yet, but the first run-through is done. Will have a more in-depth look tomorrow or Friday hopefully.

topostats/plotting.py
```diff
@@ -1,650 +1,351 @@
...
+        fig, ax = self._setup_figure()
+        sns.violinplot(data=self.melted_data, x=self.image_id, y="value", hue=self.image_id, alpha=self.alpha)
+        plt.title(self.label)
+        plt.xlabel("Image")
+        plt.ylabel(self.label)
+        outfile = self._outfile("violin")
+        self.save_plot(outfile)
+        return fig, ax
+
+    # def sns_jointplot(self, var1: str, var2: str) -> None:
```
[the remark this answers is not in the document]
ns-rse: `jointplot` was one of the features @Jean-Du had in the previous version. I'm undecided as to how easy/sensible it would be to implement given the possiblilty of all pair-wise combinations from a list of selected variables or how one might configure it more manually (thinking we don't want an overly complicated configuration system).

I've noted this in the **ToDo (maybe)** section of the PR along with other things, some of which may become issues/tasks to be completed.

ns-rse:
Thanks @SylviaWhittle completely understand, I've gone against recommended practice and my usual effort of keeping things small.

Jean-Du:
Hi @ns-rse , I haven't looked at everything yet, but one thing I notice is that the plot titles and axis labels for the plot generated by `plotting.py` are smaller than ideal; as these graphs are often used for presentations and posters, it would be useful to have the text in a larger font. 
This can be addressed through [`seaborn.set_context`](https://seaborn.pydata.org/generated/seaborn.set_context.html), as was done in old TopoStats:
```
sns.set_style("white", {"font.family": ["sans-serif"]})
# The four preset contexts, in order of relative size, are paper, notebook, talk, and poster.
# The notebook style is the default
# sns.set_context("notebook", font_scale=1.5)
sns.set_context("poster", font_scale=1.4)
```

SylviaWhittle (approved):
This works for me, and is an excellent addition. It will certainly make viewing the output statistics nice and standardised without the need for users to create individual plotting scripts. Thank you @ns-rse.

I cannot see any issues with this PR from a code perspective, so I am approving this PR, however you may want an approval from a more experimentally experienced person such as @Jean-Du before merging.

Jean-Du:
> This works for me, and is an excellent addition. It will certainly make viewing the output statistics nice and standardised without the need for users to create individual plotting scripts. Thank you @ns-rse.
> 
> I cannot see any issues with this PR from a code perspective, so I am approving this PR, however you may want an approval from a more experimentally experienced person such as @Jean-Du before merging.

@SylviaWhittle  I wouldn't call myself experimentally experienced but I'll review the PR tomorrow and see if there's anything else that needs changing

topostats/summary_config.yaml
```diff
@@ -0,0 +1,39 @@
+output_dir: ./output/summary_distributions
+csv_file: ./all_statistics.csv
+file_ext: png
+pickle_plots: True
```
Jean-Du, line 4: Could we add some comments explaining what "pickle plots" means?

ns-rse: There is a terse explanation in the table I've added to `docs/configuration.md` and I had noted this as a possible area to expand on under *ToDo (maybe)* of the initial post.

I've added the same terse description here for now and we can expand on this in a subsequent issue/work.

I was somewhat hesitant as I'm not sure many people would use this but felt it was a useful way for people to load and edit the pots generated.

notebooks/01-Summary-statistics-and-plots.ipynb
```diff
@@ -0,0 +1,313 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9820d259-680e-49fd-a1eb-e14d200a0b7a",
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "df_three_images[\"contour_lengths\"].groupby(\"image\").plot.hist(figsize=(16,9),\n",
+    "                                bins=20,\n",
```
Jean-Du, line 230: Since different files have different data ranges, having something like `bins=20` would lead to bins that don't match with each other. Would it make sense to add a (possibly commented out) line that says, eg `bins=np.linspace(0, 3e-7, 20)`, so that the user knows how to make plots with aligned bins? 
(The numpy module would need to be imported at the start)

ns-rse: My understanding of `bins` is that if its an integer then the plot has that number of equally spaced bins across the range. `np.linspace(start, stop, num=x)` essentially does the same thing.

Its a tricky balance, I appreciate not everyone knows all the intricacies and options with plotting but don't want to have to document or explain absolutely every possible option when the documentation for the various packages is out there and available. From memory I included links to tutorials and documentation for Matplotlib, Pandas and Seaborn to signpost people to these resources.

Jean-Du: @I see your point on not overcomplicating things, so do feel free to keep it as it is. But just to illustrate the difference this makes, this is an example of bars being not aligned when it's set to `bins=20`
image
And with `bins=np.linspace(0, 3e-7, 20)`, the plot now looks like this:
image

Jean-Du (approved):
Thanks a lot for refactoring this! The plotting with Jupyter notebook seems much easier to use than through the terminal. I made a few suggestions that I think would be helpful to users, but feel free to ignore them if you don't think they are a good idea.

ns-rse:
> Thanks a lot for refactoring this! The plotting with Jupyter notebook seems much easier to use than through the terminal. I made a few suggestions that I think would be helpful to users, but feel free to ignore them if you don't think they are a good idea.

Thanks for finding the time to go through and review this @Jean-Du really useful.

I'm going to make a series of issues from the `ToDo` list and your feedback so that these can be addressed in smaller quicker issues/pull-requests.

Glad you like the Notebook approach to plotting. They're quite flexible and have a [Slide mode](https://towardsdatascience.com/5-slides-for-tips-on-presentation-mode-in-jupyter-notebook-f858b15fff4f) that you may find useful for presentations as you can load the underlying data into the notebook and plot things, saves copying and pasting or inserting images into documents.

notebooks/01-Summary-statistics-and-plots.ipynb
ns-rse: Thanks that is a great example and not something I've ever noticed before. It may be an artifact of Pandas and that using Matplotlib directly doesn't result in this.

Either way I will add this to #462.

ns-rse: Realised the above example has not just different ranges, but different frequencies between the groups. I've added a commented line with a note to that effect.

PR #458, opened 2023-02-09
Maxgamill sheffield/457 mpl interpolation

### LINKED ISSUE #457: MPL interpolation causes lines in MPL Images
**Describe the bug**
Horizontal and vertical lines are seen in the matplotlib.imshow images when they are saved.

MPL image:
20221214_BW_mcc339+1_hTOP1_1000x_0 0_00051_processed (1)
Screenshot 2023-01-04 at 13 59 03

vs using `plt.imsave` to save the image directly.
20221214_BW_mcc339+1_hTOP1_1000x_0 0_00051_processed

**To Reproduce**
Run high resolution image through TopoStats with either, or both, the `colorbar` and `axes` selected to induce a matplotlib plot.
See that on these images, the data is interpolated and so show horizontal and vertical lines.

**Expected behavior**
Image should be clear and un-interpolated.

**Output**
Screenshots above.

**Additional context**
This could be down to the "nearest"-neighbour interpolation or the DPI setting.

MaxGamill-Sheffield:
Resolves #457 

This adds an interpolation option to the config file so that the saved matplotlib `imshow` images at high resolution don't show false horizontal or vertical lines in the data (when `null` option selected).

There are no tests as of yet because this is a matplotib feature. But if people believe one is necessary, let me know how I'd implement a small test for switching out this var in the config file and test it with data which shows the difference between 'nearest' and `None` (as it may not show on small images like [here](https://matplotlib.org/stable/gallery/images_contours_and_fields/interpolation_methods.html)).

There may also be a few messy bits from other PR's because I accidentally `git reset` and I couldn't get it back, even when selecting a `HEAD@{}` from the `git reflog`'s.

As an example, here is an image with the default "nearest" interpolation option (79KB):
none_figuredpi

And here is it with the `null` option at the same DPI (85KB):
20221214_BW_mcc339+1_hTOP1_1000x_0 0_00051


MaxGamill-Sheffield:
The only thing I'd say would be if this could be the default and in that case, not to trouble the user with the decision in the config file?

SylviaWhittle:
I would argue that interpolation in figures shouldn't be a thing? If we want a blur, then it should be in a well defined and typical Gaussian blur function? 

Keen to hear thoughts on this.

MaxGamill-Sheffield:
> I would argue that interpolation in figures shouldn't be a thing? If we want a blur, then it should be in a well defined and typical Gaussian blur function?
> 
> Keen to hear thoughts on this.

I'm inclined to agree. Shall I remove this as a user-set option then?

SylviaWhittle:
I would say so, but if you want to wait for a second-opinion then that's understandable. 👍

MaxGamill-Sheffield:
Note: User config option removed due as we want to avoid modifying data without full explaining it. This also helps to alleviate the cause of this issue. This also means that test images have been updated.

ns-rse:
If you truly want to eliminate any user control of this then the option should be hard-coded in the call to `plt.imshow()`. As it stands if someone were to used the class interactively they could still change the value.

Whether that is something that is desirable to prevent I couldn't say, but then someone who knows Python could always take the array and plot it themselves if they wanted to.

topostats/plottingfuncs.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         image_type: str = "non-binary",
         image_set: str = "core",
         core_set: bool = False,
-        interpolation: str = "nearest",
+        interpolation: str = None,
```
ns-rse, line 41: If you truly want to eliminate any user control of this then the option should be hard-coded in the call(s) to `plt.imshow()`. As it stands if someone were to use the class interactively they could still change the value.

ns-rse (commented):
Whether that is something that is desirable to prevent I couldn't say, but then someone who knows Python could always take the array and plot it themselves if they wanted to.

I'm of the view that options should be configurable with sane defaults. An explanation of why not to use certain interpolation methods in the documentation would perhaps then be prudent.

Reading through the docs [`gaussian` is an option for interpolation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html), so perhaps a default of `None` with a suggestion to use `gaussian` if smoothing is required. :shrug:

topostats/plottingfuncs.py
MaxGamill-Sheffield: Sounds good - I've moved it to be hardcoded in the calls! I think as we've already got the gaussian function in the flattening section which applies the filter to the image, it essentially will do the same thing so I don't think there is a need for it here. What do you think?

ns-rse: The flattening is part of the image processing pathway (and there is also Gaussian methods applied as part of the DNA Tracing processing).

The plotting is very much separate from that in so much as it could be used to plot a `np.ndarray` at any point. I prefer to design software to not have hard-coded values but rather to be open and as flexible as possible. Currently this works through `**kwargs`.

One thing I'd ask if the open and flexible design of making this a configurable option is to be ditched is how likely people are to...

* go with anything but the default settings that are loaded.
* if using a custom configuration file change the field pertaining to `interpolation` method?

I realise these are somewhat rhetorical, but is there any point taking away configuration options because they shouldn't be used if there is a very low likelihood that they will be changed?

Making it configurable but clear what the developers recommend and why that recommendation is made would be my preference.

ns-rse:
> Solves #457
> 

GitHub doesn't recognise _Solves_ as a [keyword](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) but does recognise _Close[s]_ _Fix[es]_ and _Resolve[s]_ and will then close the mentioned issue on merging.

topostats/validation.py
```diff
@@ -223,6 +223,11 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 "core",
                 error="Invalid value in config for 'plotting.image_set', valid values " "are 'all' or 'core'",
             ),
+            "pixel_interpolation": Or(
```
ns-rse, line 226: Currently this would allow `neil` as a valid option but that would then fail when it reached Matplotlib.

It should perhaps list the options so they can be validated (and if there was a desire to limit choices  to `None`, and `gaussian` this would be the place to do it).

topostats/plottingfuncs.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         image_type: str = "non-binary",
         image_set: str = "core",
         core_set: bool = False,
-        interpolation: str = "nearest",
+        pixel_interpolation: str = "nearest",
```
ns-rse, line 41: If the aim is to have the default as `None` then whilst the value carries through from the configuration when called from `run_topostats` it would probably be wise to change the value here to be `None`.

ns-rse (changes requested):
Thanks for reverting the configuration of `interpolation`.

I think it would be prudent to explain  `docs/configuration.md` why this is `null` by default and that whilst other options are available that the only recommended one is `gaussian` and what the side-effects of say `nearest` are.

I realise its a bit of a hassle to update documentation but its easier to do it in small amounts when specific pieces of work are done than trying to do it all retrospectively.

docs/configuration.md
```diff
@@ -90,6 +90,7 @@ Aside from the comments in YAML file itself the fields are described below.
...
 | `plotting`   | `run`                          | boolean    | `true`         | Whether to run plotting. Options : `true`, `false`                                                                                                                                                                                                           |
 |              | `save_format`                  | string     | `png`          | Format to save images in, see [matplotlib.pyplot.savefig](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html)                                                                                                                          |
+|              | `pixel_interpolation`          | string or None  | `None`    | Interpolation method for matplotlib image plots (when one or both of `axes` or `colorbar` is true). See [matplotlib imshow intepolations](https://matplotlib.org/stable/gallery/images_contours_and_fields/interpolation_methods.html).                                                                                                                         |
```
ns-rse, line 93: Ah, you pushed this whilst I was writing my review.

```suggestion
|              | `pixel_interpolation`          | string or None  | `None`    | Interpolation method for matplotlib image plots (when one or both of `axes` or `colorbar` is true). The recommended default of `None` prevents banding that occurs in high resolution images when `nearest` is used. If interpolation is required it is suggested to use `gaussian`. For all possible options refer to the [matplotlib imshow interpolations documentation](https://matplotlib.org/stable/gallery/images_contours_and_fields/interpolation_methods.html).                                                                                                                         |
```

ns-rse (changes requested):
Hope the suggested text makes sense, I haven't explained why `gaussian` is recommended so that could perhaps be expanded on (but I'm not sure why, just that it was mentioned in the discussion).

topostats/plottingfuncs.py
MaxGamill-Sheffield: Makes sense. I'll revert the changes (and learn how to do so in the process!)

topostats/validation.py
MaxGamill-Sheffield: I'll add the options in now. I've gone for all options because if people want them they are there, and if they change the defaults, it can be expected that they understand the changes the interpolation methods incur.

docs/configuration.md
MaxGamill-Sheffield: I like this better than what I put as you were writing this! I'll update it now.

MaxGamill-Sheffield:
Test images pass locally, but not here - I would greatly appreciate if anyone could generate new images from this branch and upload them so the tests pas and we can check image comparisons here

SylviaWhittle:
Tests pass locally for me too.

MaxGamill-Sheffield:
Closing as Sylvia opened #464 which is this but also after resolving merge conflicts

PR #460, opened 2023-02-09
Add Height Sensor channel to validation.py

### LINKED ISSUE #459: Add channel to validation.py
Currently, `validation.py` does not have an option in the channels section for "Height Sensor", a commonly used `.spm` channel. This means that in order to extract image data from `.spm` files that only have the "Height Sensor" channel, one needs to edit `validation.py`, since otherwise the config file gets flagged as being invalid.

Add "Height Sensor" channel to `validation.py` so that users can use this `.spm` channel without editing the code in `validation.py`.

SylviaWhittle:
Closes #459 

This PR adds the commonly used "Height Sensor" channel to `validation.py`.

### DIFF
--- a/topostats/validation.py
+++ b/topostats/validation.py
@@ -56,6 +56,7 @@ DEFAULT_CONFIG_SCHEMA = Schema(
                 "Adhesion",
                 "Deformation",
                 "Dissipation",
+                "Height Sensor",
                 "Height",  # end of spm channels
                 "HeightTracee",
                 "HeightRetrace",


ns-rse (approved):
Nothing to change here, feel free to go ahead and merge, it shouldn't cause any conflicts.

Re: @MaxGamill-Sheffield question about whether this is something to look out for in the future, if I understand correctly these are fields coming directly from the files and so we are at the mercy of those who write the software for that. They could introduce changes, but when we see them might well be contingent on how frequently the software on the machines gets updated. We could set ourselves a task of periodically checking this or if the software has a release cycle set up notifications (could go to `topostats@sheffield.ac.uk` :wink: ) to be informed of when new releases are made and then check the `ChangeLog` if there is one.

Alternatively we could wait for someone to report "_This doesn't work with the new channel, could you please add support_" and drop it in then.

Perhaps a combination of the two might be the best approach, especially as support for more files grows as the tracking task then becomes more onerous.

SylviaWhittle:
@MaxGamill-Sheffield good question. The reason I made this PR, is that I received some data from someone to process, and the only channel available in the `.spm` files was "Height Sensor", but I found that when I tried to use "Height Sensor" in the config, `validation.py` said that it was not a legal channel, so I had a look in `validation.py` and it indeed wasn't in the list. I then added it to the list, and was able to successfully process the `.spm` file. Does this answer your question?

ns-rse:
@MaxGamill-Sheffield I'm not sure that would work since validation currently occurs before any files of the given file extension are detected, and in turn before any are loaded. It could be a function that is called from `validate_config()` when an error occurs but that would require first finding files and loading them to find out what channels are available and somewhat defeats the point of validating a configuration _before_ proceeding to processing.

ns-rse:
#422 relates to the issue of channels.

PR #464, opened 2023-02-12
Maxgamill sheffield/DUMMY PR 457 mpl interpolation updated

SylviaWhittle:
Dummy PR, testing if changes work


topostats/plottingfuncs.py
```diff
@@ -74,8 +74,8 @@ def __init__(
             The set of images to process - core or all.
         core_set : bool
             Flag to identify image as part of the core image set or not.
-        pixel_interpolation: str
-            Interpolation to use (default 'none').
+        pixel_interpolation: Union[str, None]
+            Interpolation to use (default: null).
```
ns-rse, line 78: Very minor `(default: None)`.

Its a shame there is consistency between languages on what "null" is represented by.

ns-rse:
This is good to go now, thanks and apologies for not looking at this in closer detail earlier.

May get a conflict with my PR in the [`update_plotting_config()`](https://github.com/AFM-SPM/TopoStats/pull/527/files#diff-3d7533d8ee7b95a5ac8985e5ccbdafc6af6eb3e31e0f27041ba7a87bd9514935) function. I think the more general solution I've proposed should be opted for as it saves having to manually update the `utils.update_plotting_config()` each time a new configuration option is added, but would be useful to hear other opinions.

ns-rse (approved):

PR #467, opened 2023-02-14
Update Bug template to request commit hash; TopoStats arg to show version

ns-rse:
As discussed in todays meeting (but not noted in a specific issue) it would be useful to report the full version, including a commit hash if we are working on development.

This PR updates the `bug_repor.md` template to request that, including instructions on how to obtain it.

To assist a command line argument `-v/--version` has been added to `run_topostats` which reports the version.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -20,9 +20,10 @@ A clear and concise description of what you expected to happen.
 
 **Output**
 
-If applicable please include the output error, this can be a copy and paste of the output (preferable** or a screenshot.
-If applicable, add screenshots to help explain your problem.
-Output files can be attached to this bug report.
+If applicable please include the output error, this can be a copy and paste of the output (preferable) or a screenshot.
+
+Any images or output files that have been produced can be attached to this bug report (by default they will be under the
+`output` directory unless you have customised the configuration).
 
 ** TopoStats version
 
@@ -34,12 +35,12 @@ have installed TopoStats from source/GitHub and are working on features.
 - `run_topostats --version`
 
 
+** Your computer configuration (please complete the following information):**
 
-**(Not required if using Docker) Your computer configuration (please complete the following information):**
- - OS: e.g. windows, MacOS, linux; please include OS version
- - topostats version: best way to get this at the moment is to paste the results of running `git show --oneline -s` in your topostats folder.
- - Python version: best way to get this is to paste the results of typing `python --version`.
- - Optionally, your installed packages: best way to get this is to paste the results of typing `pip freeze`.
+- OS: e.g. windows, MacOS, linux; please include OS version
+- Python version: paste the results of typing `python --version`.
+- Optionally, your installed packages: the best way to get this is to copy and paste the results of typing `pip freeze`.
 
 **Additional context**
+
 Add any other context about the problem here.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -20,10 +20,21 @@ A clear and concise description of what you expected to happen.
...
+have installed TopoStats from source/GitHub and are working on features.
+
+- `pip show topostats`
+- `run_topostats --version`
+
+
+
 **(Not required if using Docker) Your computer configuration (please complete the following information):**
  - OS: e.g. windows, MacOS, linux; please include OS version
  - topostats version: best way to get this at the moment is to paste the results of running `git show --oneline -s` in your topostats folder.
```
MaxGamill-Sheffield, line 40: Do we also want to update this with something like "if you have cloned from the github repo, use this method instead"?

MaxGamill-Sheffield (dismissed):
Looks good to me! Just a clarification for the users on when to the `-oneline` git command or `run_topostats -v`.

Also, this is the output I get from the `-v`:
(sylv) Max:TopoStats Maxgamill$ run_topostats -v
Installed version of TopoStats : {'version': '2.0.0+324.g811fd5e', 'full-revisionid': '811fd5e60da777b4b48ab0bb14d867510d404e34', 'dirty': False, 'error':
None, 'date': '2023-02-14T12:09:29+0000'}

I'm not sure if you want to make this look prettier by separating each item onto a new line or if this is fine, but I please can you to clarify to me which item I should be quoting - the version or revision ID?

.github/ISSUE_TEMPLATE/bug_report.md
ns-rse: I'd completely overlooked that and written a whole new section on version.

We can also remove reference to Docker, anyone using "old" TopoStats should be directed to the newer one.

I'll re-jig it now.

ns-rse:
> Looks good to me! Just a clarification for the users on when to the `-oneline` git command or `run_topostats -v`.
> 
> Also, this is the output I get from the `-v`: (sylv) Max:TopoStats Maxgamill$ run_topostats -v Installed version of TopoStats : {'version': '2.0.0+324.g811fd5e', 'full-revisionid': '811fd5e60da777b4b48ab0bb14d867510d404e34', 'dirty': False, 'error': None, 'date': '2023-02-14T12:09:29+0000'}
> 
> I'm not sure if you want to make this look prettier by separating each item onto a new line or if this is fine, but I please can you to clarify to me which item I should be quoting - the version or revision ID?

I missed the `git` command and have now removed it as version is important so its now under its own section, thanks for catching that.

We can work out what version/commit from either of the commands I've suggested, but the later is preferable and so I wrote the following...

>  There are several ways of doing this, either with `pip` or `run_topostats`. Please copy and paste all output from either of the following, although the later is preferable if you
have installed TopoStats from source/GitHub and are working on features.

The former (i.e. `pip show topostats`) shows a truncated version of the commit hash. The later (which comes from [versioneers `get_versions()` function](https://github.com/python-versioneer/python-versioneer)) returns a dictionary with this as the first item as well as the full hash and as you found whether its committed or not (the `dirty` bit), the `date` and an `error` field. We can use that information and pick what we need and I find it easier to ask for as much information to be copy and pasted rather than make it confusing by asking for a subset and expect users to select the right thing.

If the above isn't clear please do let me know what is ambiguous and I'll happily adapt it. :+1:

SylviaWhittle (approved):
Works and looks good to me 👍

PR #471, opened 2023-02-16
Make median flattening quantile configurable

### LINKED ISSUE #468: Make median_flatten quantile configurable.
Change the method that `median_flatten` uses to align the rows from a median, to a configurable quantile. This will (I have tested) allow for improvements in images such as are outlined in #432 .

A configurable parameter will be made and will be editable via the `filters` section of the config.

SylviaWhittle:
This PR makes the quantile that is used for the background height duing `median_flatten` in `filters.py` configurable. Reducing this value to 0.2 helps when flattening large structures that take up more than 50% of the pixels in a row.

Closes #468


ns-rse (changes requested):
Great addition, I much prefer having parameters configurable.

One thing that we perhaps should be doing (and you're not doubt tiring of me banging this :drum: ) is to add parametrized tests to check the impact of different values of `row_alignment_quantile` on a small dummy `np.ndarray`. Without tests for each method and its possible options our code coverage starts to slip and its easier to stay on top of tests when we are working on features rather than doing it retrospectively.

ns-rse (approved):
Brilliant, thanks for adding the tests so quickly @SylviaWhittle looks good to merge.

PR #480, opened 2023-02-24
Fixing plotting in light of feedback

ns-rse:
@derollins was testing the `plotting` refactor and found some errors. This PR corrects those.

The `write_yaml()` function had a hard coding to convert `Path()` objects in a configuration dictionary to `str` for writing to file. That is now done by the `path_to_str()` function (tests included) and in doing so removes the need for the presence of `base_dir: ./` in the `summary_config.yaml`.

Also if a `df` is not passed to the `TopoSum` class when it was being instantiated nothing could be plotted. If no `df` is provided in the `config` (which `run_topostats` adds before plotting) then `TopoSum` now loads the provided `csv` file and the plotting can proceed.


Jean-Du (approved):
I tried running `toposum` on its own and it works for me

PR #492, opened 2023-03-06
457 Investigations

ns-rse:
Getting stuck into investigating the artifacts reported in #457.

Wanted an easy way of sharing arrays of different sizes (100, 200 and 300) using _all_ of the different interpolation methods for two colourmaps (`afmhot` and `nanoscope`).

**No intention of merging this its just to share images**


ns-rse:
Closing, GitHub doesn't render all images, they can instead be viewed [here](https://kimura.no-ip.info/~arch/img/matrix.html).

PR #497, opened 2023-03-11
Corrects pprint usage; tidies up runtime arguments

### LINKED ISSUE #496: pprint() show output; -q flag no longer silent
### pprint()

Updating logging to allow `LOGGER.debug()` has left a `pprint()` statement in `run_topostats()` this outputs to `stdout` even though its wrapped within `LOGGER.debug(f"Configuration after update         : \n{pprint(config, indent=4)}")` (see line 213).

The solution appears to be to use [`pprint.pformat()`](https://stackoverflow.com/a/11093247) instead.

### `-q/--quiet` no longer silent

This flag no longer results in silent behaviour. Two options...

1) Remove flag and use `--log-level` which would be my preference (requires updating documentation).
2) Reinstate and have the `-q`/`--quiet` option set the log-level to `warning`. Less desirable to me as its duplication of options.

ns-rse:
Closes #496

* `pformat` used over `pprint` in debug statement so config is no longer printed.
* Removes the `-q`/`--quiet` argument to `run_topostats` in favour of just using `-l`/`--log-level` to set verbosity.
* Updates documentation to reflect this (including removing reference to `warnings` which had already been removed).


SylviaWhittle (approved):
Works for me 😄

PR #520, opened 2023-03-28
Removing errant pkl

ns-rse:
`.pkl` file included by mistake, its better that users running Notebooks generate a `.pkl` first of all to then use as it encourages familiarity with and usage of the `run_topostats` command and configuration files. This is already stated in the Notebooks which use this file.


SylviaWhittle (approved):
Thanks for correcting this mistake! Sorry for inadvertently adding it 😅

ns-rse:
> Thanks for correcting this mistake! Sorry for inadvertently adding it sweat_smile

Nothing to worry about at all. Thanks for the :heavy_check_mark:

PR #527, opened 2023-04-14
Makes DPI configurable via default_config.yaml

### LINKED ISSUE #526: Make DPI configurable
We should have sensible defaults for `dpi` of image scans so that users can configure this if greater (or less) detail is required.

Files to modify...

* [x] `topostats/default_config.yaml`
* [x] `topostats/validation.py`
* [x] `topostats/plottingfuncs.py`
* [x] `docs/configuration.md`

ns-rse:
Closes #526

The `Images` class always had `dpi` as an option but it was not accessible from `run_topostats` as no options were present in the configuration file and therefore didn't cascade through. Dots Per Inch (DPI) can now be specified in the configuration file, its default is set to `100.0` (the default used by `savefig` in Matplotlib). Validation is performed and a test introduced that sets `dpi = 300` for testing to check the value cascades through and plots are made using this value (this also required setting the parameter for the `pytest-mpl` fixture too, otherwise it would save the figure with the default value).


topostats/validation.py
```diff
@@ -262,6 +262,9 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 ),
             ),
             "histogram_bins": lambda n: n > 0,
+            "dpi": Or(
+                lambda n: n > 0, "figure", error="Invalid valud in config for 'dpi', valid values are 'figure' or > 0."
```
SylviaWhittle, line 266:
```suggestion
                lambda n: n > 0, "figure", error="Invalid value in config for 'dpi', valid values are 'figure' or > 0."
```

SylviaWhittle (changes requested):
Minor typo.

Also, I may be misunderstanding, but there seems to be no code here to set the DPI for output figures, only for `test_high_dpi()`? Is this intended?

topostats/validation.py
ns-rse: You weren't mistaken, I corrected it.

ns-rse:
> Just to check...
> 
> I tried running topostats with dpi=1 and it produced exactly the same image as dpi=1000. Is this normal?

That isn't what I'd expect, and sorry for not checking.

I've investigated and realised that I'd forgotten how we pass options to each image (via the `plot_dict` key/value of `plotting_config`). In doing so we update each of these nested dictionaries with the plotting options and I hadn't added `dpi` to this list.

I've therefore modified how this is done in the `update_plotting_config()` function so that in the future new options that are added won't require explicit inclusion, they will be passed through automatically thanks to `**kwargs` expansion (see [27be2fbb](https://github.com/AFM-SPM/TopoStats/pull/527/commits/27be2fbb7109e155dd7c36e85f5c17c37670d0a5) for the specific change if curious).

Thanks for catching this, a glaring omission on my behalf :blush: but hopefully an improved solution.

topostats/utils.py
```diff
@@ -97,18 +97,11 @@ def update_plotting_config(plotting_config: dict) -> dict:
...
-            "save_format": plotting_config["save_format"],
-            "image_set": plotting_config["image_set"],
-            "colorbar": plotting_config["colorbar"],
-            "axes": plotting_config["axes"],
-            "cmap": plotting_config["cmap"],
-            "mask_cmap": plotting_config["mask_cmap"],
-            "zrange": plotting_config["zrange"],
-            "histogram_log_axis": plotting_config["histogram_log_axis"],
-        }
+        plotting_config["plot_dict"][image] = {**options, **main_config}
```
SylviaWhittle, line 104: Very elegant! Thanks for the improvement.

MaxGamill-Sheffield:
I just wanted to clarify behaviour for this. So the dpi was originally put into the images class in preparation for DNAtracing to produce visible single pixel traces with the intention to set these at the plottingdict level.

So my question is if this is now a config parameter, have we given any thought to handling different images separately? On my cats branch where I've implemented this it takes a while to produce a single plot at 1000 dpi (to show representitive skeletons / traces although I've not seen how low of a dpi I could go), so if all 20 plots are set globally like this are we concerned with runtime?

SylviaWhittle:
> I just wanted to clarify behaviour for this. So the dpi was originally put into the images class in preparation for DNAtracing to produce visible single pixel traces with the intention to set these at the plottingdict level.
> 
> So my question is if this is now a config parameter, have we given any thought to handling different images separately? On my cats branch where I've implemented this it takes a while to produce a single plot at 1000 dpi (to show representitive skeletons / traces although I've not seen how low of a dpi I could go), so if all 20 plots are set globally like this are we concerned with runtime?

@MaxGamill-Sheffield Good point. I suppose that this is the base functionality, and in a new issue / PR, we could add the ability to only let it affect the two main output plots?

PR #538, opened 2023-04-21
Fix summary_config undefined in completion message

### LINKED ISSUE #537: Completion message fails when summary stats option is disabled
Writing on behalf of @BillieWard who discovered this issue.

## Checklist

Please try and tick off each of these items when filing the bug report. There are further instructions on each below.

* [ ] Re-run analysis with `run_topostats --core 1`.
* [ ] Describe the bug.
* [ ] Include the configuration file.
* [ ] Copy of the output.
* [ ] The exact command that failed. This is what you typed at the command line, including any options.
* [ ] TopoStats version, this is reported by `run_topostats --version`
* [ ] Operating System and Python Version

## Describe the bug
When running TopoStats with the `summary_stats` `run` option set to `false`, the completion message fails.

## Copy of the output

```
  File "/Users/billieward/opt/anaconda3/envs/topostats-github-env/bin/run_topostats", line 8, in <module>
    sys.exit(main())
  File "/Users/billieward/Documents/PhD/TopoStats/TopoStats-main/topostats/run_topostats.py", line 299, in main
    completion_message(config, img_files, summary_config, images_processed)
UnboundLocalError: local variable 'summary_config' referenced before assignment
```

## Include the configuration file.

Default config with `summary_stats` `run` set to `false`

## To Reproduce

Run topostats with the `summary_stats` `run` option set to `false`.

## TopoStats version

Please report the version of TopoStats you are using. There are several ways of doing this, either with `pip` or
`run_topostats`. Please copy and paste all output from either of the following commands.

- `pip show topostats`
- `run_topostats --version`

```
<-- PASTE TOPOSTATS VERSION -->
```

### Operating System

Please let us know what operating system you are using, if you have used more than one then tick all boxes.

* [ ] Windows
* [ ] MacOS Intel (pre-2021)
* [ ] MacOS M1/M2 (post-2021)
* [ ] GNU/Linux (please include distribution)

### Python Version

Please let us know the version of Python you are using, paste the results of `python --version`

```
3.10.9
```

## Additional context

The issue seems to be that:

`completion_message()` takes `summary_config`:

image

But in `run_topostats.py`, we only set the `summary_config` parameter inside the `if` statement checking for `summary_stats` `run`:

image
 

Solution may be to create an empty string to the variable before the if-statement.

SylviaWhittle:
Closes #537 

This error was caused by the `summary_stats`/`run` option being `false`, and so the variable `summary_config` was never assigned before being accessed by `completion_message`.

- This PR fixes the issue of the call to `completion_message` crashing due to an unassigned variable `summary_config`, by setting `summary_config` to `None` by default.
- It also adds a message that replaces the directory for the summary stats in the completion message (if the summary stats are not generated).

### DIFF
--- a/topostats/run_topostats.py
+++ b/topostats/run_topostats.py
@@ -245,7 +245,6 @@ def main(args=None):
         LOGGER.error(error)
 
     # Summary Statistics and Plots
-    summary_config = None
     if config["summary_stats"]["run"]:
         # Load summary plots/statistics configuration and validate, location depends on command line args or value in
         # any config file given, if neither are provided the default topostats/summary_config.yaml is loaded
@@ -282,6 +281,8 @@ def main(args=None):
                 "There are no results to plot, either you have disabled grains/grainstats/dnatracing or there "
                 "have been errors, please check the log for further information."
             )
+    else:
+        summary_config = None
 
     # Write statistics to CSV
     if isinstance(results, pd.DataFrame):
--- a/topostats/run_topostats.py
+++ b/topostats/run_topostats.py
@@ -245,6 +245,7 @@ def main(args=None):
         LOGGER.error(error)
 
     # Summary Statistics and Plots
+    summary_config = None
     if config["summary_stats"]["run"]:
         # Load summary plots/statistics configuration and validate, location depends on command line args or value in
         # any config file given, if neither are provided the default topostats/summary_config.yaml is loaded
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -350,6 +350,11 @@ def completion_message(config: Dict, img_files: List, summary_config: Dict, imag
     -------
     None
     """
+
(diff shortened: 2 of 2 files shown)


topostats/run_topostats.py
```diff
@@ -245,6 +245,7 @@ def main(args=None):
         LOGGER.error(error)
 
     # Summary Statistics and Plots
+    summary_config = None
```
ns-rse, line 248: An alternative is to use the `if config["summary_stats"]["run"]:` that we have an follow it with...

```
else:
    summary_config = None
```

SylviaWhittle: Yeah I wasn't sure which to use and went with the option that added fewer lines, but now I think about it, this is more readable. Thanks 😄

ns-rse (approved):
Cool, good to go :+1:

PR #548, opened 2023-05-04
Updating documentation

ns-rse:
No issue to close as I didn't create one, tasks grew from the initial fixing of links on `README.md` but tidies things up

* A number of hard coded links in `README.md` failed as a consequence of updating and building documentation for multiple versions and `main` branch (see #528).
* Tidies up some issues within the existing `docs/*.md`, fixing some internal links and rewording/clarifying a few things.
* Introduces [markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2) via a `pre-commit` hook so that all changes to Markdown files (`.md`) will now be linted before commits are possible. Introduces a configuration file which disables line-length checks for tables and code-blocks.
* Rather than expecting others to lint existing code when they touch and `.md` files I have also gone through and linted everything with this configuration. This was done in a separate commit so that I could...
* Ignore the commit from `git blame` by listing it in `.git-blame-ignore-revs` as linting changes which only really affect layout shouldn't be attributed to those who made them but the original author.


docs/index.rst
```diff
@@ -25,19 +25,6 @@ listed below to find out more.
...
-   topostats.grainstats
-   topostats.io
-   topostats.plotting
-   topostats.plottingfuncs
-   topostats.run_topostats
-   topostats.theme
-   topostats.thresholds
-   topostats.tracing
-   topostats.utils
-   topostats.validation
```
SylviaWhittle, line 39: Just interested - I understood these as telling GitHub Actions which pages to render on the GitHub Pages site? I'm guessing you're changing how these are selected?

SylviaWhittle (commented):
I have read through all this and it seems good to me. I clicked every link to verify they work. All the links for local files failed for me, however I think this is just because I rendered it and the local server didn't search for the files and doesn't mean they won't work when published to GitHub Pages?

There is a failing test on the pre-commit so I'll be ready to approve once that is resolved 👍

docs/index.rst
ns-rse: The [sphinx-autoapi](https://sphinx-autoapi.readthedocs.io/en/latest/) package now handles this, it sticks a link into the API reference and I figured since most users won't be reading that removing the explicit links from the front-page would do no harm, those interested can still find them.

ns-rse:
The line-length failure is annoying, it passes locally and I thought the presence of `.markdownlint-cli2.yaml` should be picked up automatically (although I did explicitly include it as an `arg` in the `.pre-commit-config.yaml` just in case).

Will sort that out tomorrow.

SylviaWhittle (approved):
Built successfully (locally) for me, and all links work 😄

PR #549, opened 2023-05-04
Sylvia whittle/pandas reset index error

SylviaWhittle:
Fixes a bug where TopoStats would crash just before saving results to csv, on calling `results.reset_index()`. After a lot of digging, it appears that this is caused by a line in `create_empty_dataframe` that sets the index of the empty dataframe to `molecule_number` without then removing the column name `molecule_number`. This then causes a crash when `reset_index()` attempts to turn the index column `molecule_number` into a regular column, since there is already a regular column of the same name.

This issue only happens if grainstats returns an empty dataframe, and so was difficult and rare to find. The solution that I have proposed brings the dataframe that is returned by `create_empty_dataframe()` in line with the dataframes that are returned by grainstats when grains have been found, by using the `set_index()` method of dataframes, which removes the column after setting the index to that column name. Here is a visual example:

```python
import pandas as pd
import numpy as np

df = pd.DataFrame(data={'col1': [7, 8, 9], 'molecule_number': [0, 1, 2]})
print(df)
df.index.name = 'molecule_number'
print(df)
df_reset = df.reset_index()
print(df_reset)

```
```
col1  molecule_number
0     7                0
1     8                1
2     9                2
                 col1  molecule_number
molecule_number                       
0                   7                0
1                   8                1
2                   9                2
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[7], line 8
      6 df.index.name = 'molecule_number'
      7 print(df)
----> 8 df_reset = df.reset_index()
      9 print(df_reset)

File [/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/util/_decorators.py:331](https://file+.vscode-resource.vscode-cdn.net/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/util/_decorators.py:331), in deprecate_nonkeyword_arguments..decorate..wrapper(*args, **kwargs)
    325 if len(args) > num_allow_args:
    326     warnings.warn(
    327         msg.format(arguments=_format_argument_list(allow_args)),
    328         FutureWarning,
    329         stacklevel=find_stack_level(),
    330     )
--> 331 return func(*args, **kwargs)

File [/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/frame.py:6361](https://file+.vscode-resource.vscode-cdn.net/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/frame.py:6361), in DataFrame.reset_index(self, level, drop, inplace, col_level, col_fill, allow_duplicates, names)
   6355         if lab is not None:
   6356             # if we have the codes, extract the values with a mask
   6357             level_values = algorithms.take(
   6358                 level_values, lab, allow_fill=True, fill_value=lev._na_value
   6359             )
-> 6361         new_obj.insert(
   6362             0,
   6363             name,
   6364             level_values,
   6365             allow_duplicates=allow_duplicates,
   6366         )
   6368 new_obj.index = new_index
   6369 if not inplace:

File [/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/frame.py:4817](https://file+.vscode-resource.vscode-cdn.net/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/frame.py:4817), in DataFrame.insert(self, loc, column, value, allow_duplicates)
   4811     raise ValueError(
   4812         "Cannot specify 'allow_duplicates=True' when "
   4813         "'self.flags.allows_duplicate_labels' is False."
   4814     )
   4815 if not allow_duplicates and column in self.columns:
   4816     # Should this be a different kind of error??
-> 4817     raise ValueError(f"cannot insert {column}, already exists")
   4818 if not isinstance(loc, int):
   4819     raise TypeError("loc must be int")

ValueError: cannot insert molecule_number, already exists
```

My fix:
```python
df = pd.DataFrame(data={'col1': [7, 8, 9], 'molecule_number': [0, 1, 2]})
print(df)
df = df.set_index('molecule_number')
print(df)
df_reset = df.reset_index()
print(df_reset)
```
```
   col1  molecule_number
0     7                0
1     8                1
2     9                2
                 col1
molecule_number      
0                   7
1                   8
2                   9
   molecule_number  col1
0                0     7
1                1     8
2                2     9
```

### DIFF
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -149,11 +149,12 @@ def process_scan(
             )
             grains.find_grains()
         except IndexError:
-            LOGGER.info(f"[{filename}] : No grains were detected, skipping Grain Statistics and DNA Tracing.")
+            LOGGER.warning(f"[{filename}] : No grains were detected, skipping Grain Statistics and DNA Tracing.")
         except ValueError:
-            LOGGER.info(f"[{filename}] : No image, it is all masked.")
+            LOGGER.warning(f"[{filename}] : No image, it is all masked.")
             results = create_empty_dataframe()
         if grains.region_properties is None:
+            LOGGER.warning(f"[{filename}] : No region properties found for grains.")
             results = create_empty_dataframe()
         # Optionally plot grain finding stage
         if plotting_config["run"] and grains.region_properties is not None:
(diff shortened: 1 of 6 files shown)


topostats/processing.py
```diff
@@ -192,91 +193,101 @@ def process_scan(
...
-                            grains=grains.directions[direction]["labelled_regions_02"],
-                            filename=filename,
-                            pixel_size=pixel_to_nm_scaling,
-                            **dnatracing_config,
-                        )
-                        dna_traces[direction].trace_dna()
-                        tracing_stats[direction] = traceStats(trace_object=dna_traces[direction], image_path=image_path)
-                        tracing_stats[direction].df["threshold"] = direction
+        if grainstats_config["run"]:
+            if grains.region_properties is not None:
```
ns-rse, line 197: Much better separation of why processing fails, thank you.

topostats/utils.py
```diff
@@ -237,5 +237,5 @@ def create_empty_dataframe(columns: set = ALL_STATISTICS_COLUMNS, index: tuple =
         Empty Pandas DataFrame.
     """
     empty_df = pd.DataFrame(columns=columns)
-    empty_df.index.names = [index]
+    empty_df = empty_df.set_index(index)
```
ns-rse, line 240: I'm surprised the difference between these two has such an impact, I naively thought they were synonyms.

ns-rse (commented):
Good effort digging into this problem, it was my previous attempt to return a genuinely empty dataframe without rows of all `NaN` that has resulted in this.

I'm going to sound like a broken record and you'll perhaps start loathing me for doing so but a tests to ensure this is covered would be useful.

It shouldn't be too involved though as it can be broken down into...

1. A test that ensures the index name of the empty data frame created by `utils.create_empty_dataframe()` is correct (somewhat redundant as I don't expect Pandas to get things wrong but ensures our function works as expected).
2. A test for `processing.process_scan()` that ensures that when either of

```python
        if grainstats_config["run"]:
```

...are encountered the appropriate log message is seen.

3. Similar for `run_topostats.py` to ensure

```python
if results.shape[0] > 1:
```

...logging is correct.

We already have tests that cover the event that these sections in 2. and 3. actually run.

SylviaWhittle:
Thanks for the comments! I should have noticed this in my review of the previous PR 😬 though in my defence, it only happens in one of 50 images that Eddie was trying today.

Not a problem about the tests. I honestly just keep forgetting / not realising that a test would be good to put in for these, so thank you for pointing out where I can improve my PRs!

topostats/utils.py
ns-rse: Ah, I now realise these aren't synonyms `empty_df.set_index(index)` is saying to make the existing column the index rather than just naming the index with the value of `index`.

ns-rse:
> Not a problem about the tests. I honestly just keep forgetting / not realising that a test would be good to put in for these, so thank you for pointing out where I can improve my PRs!

I've written a test for `create_empty_dataframe()` this morning.

We already have a (parametrized) test in place for testing the logic of whether different stages get run by `process_scan()` (see `tests/test_processing/test_process_stages()`). Just working on mocking a test to capture `if grains.region_properties is not None:`. Is it fair to say that if `grains.region_properties == None` then in essence there are no grains? If so I'll tweak the log message...

```
                LOGGER.info(
                    f"[{filename}] No grains have been detected, skipping calculation of grain statistics."
                )
```

SylviaWhittle:
> > Not a problem about the tests. I honestly just keep forgetting / not realising that a test would be good to put in for these, so thank you for pointing out where I can improve my PRs!
> 
> I've written a test for `create_empty_dataframe()` this morning.
> 
> We already have a (parametrized) test in place for testing the logic of whether different stages get run by `process_scan()` (see `tests/test_processing/test_process_stages()`). Just working on mocking a test to capture `if grains.region_properties is not None:`. Is it fair to say that if `grains.region_properties == None` then in essence there are no grains? If so I'll tweak the log message...
> 
> ```
>                 LOGGER.info(
>                     f"[{filename}] No grains have been detected, skipping calculation of grain statistics."
>                 )
> ```

I think that is the case, yes

ns-rse (approved):
Good idea adding a check that `molecule_number` isn't in the `.columns`. :+1:

PR #555, opened 2023-05-09
Fix no grains found crash

SylviaWhittle:
Fixes the error we found last week where images that had no grains would crash TopoStats. This was due to sometimes returning an empty `defaultdict` object, and sometimes returning `None` instead.

This solution fixes that issue, and another issue, that had not been noticed yet, where if you were to run an image and request grains to be found in both directions `above` and `below`, then it would detect grains in `below` fine, but if when detecting grains in `above`, it found no grains, then it would error out, and skip steps for both the `above` and `below` directions. 

I fix this by checking explicitly for each direction before calling grainstats. This also gives a more informative output message, so you know which direction found no grains:

```
[Tue, 09 May 2023 13:37:26] [INFO    ] [topostats] [minicircle] : Grains found for direction above : 0
```

and 

```
[Tue, 09 May 2023 13:37:26] [INFO    ] [topostats] [minicircle] : *** Grain Statistics ***
[Tue, 09 May 2023 13:37:26] [WARNING ] [topostats] [minicircle] : No grains exist for the above direction. Skipping grainstats and DNAtracing.
```

I have edited @ns-rse 's test to accomodate this.

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -311,12 +311,11 @@ def test_process_stages(
     assert log_msg2 in caplog.text
 
 
-def test_process_scan_region_properties_is_none(
-    process_scan_config: dict, load_scan_data: LoadScans, tmp_path: Path, caplog
-) -> None:
-    """Test capturing disabling GrainStats where no grains have been found."""
+def test_process_scan_no_grains(process_scan_config: dict, load_scan_data: LoadScans, tmp_path: Path, caplog) -> None:
+    """Test handling no grains found during grains.find_grains()."""
     img_dic = load_scan_data.img_dict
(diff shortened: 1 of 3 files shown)


ns-rse (changes requested):
Thanks, I'd forgotten about this needing including (I think that's because we hadn't captured it as an issue in GitHub and added it to the project as it was discussed in Slack).

All looks great. I've made a suggestion that we don't bothering trying to plot grains that don't exist I don't know if people would want to see those images? Otherwise good to go.

ns-rse (approved):
After discussion plotting of grains, even when grains aren't found is desirable, it may be the lack of detected grains is down to inappropriate thresholds for grain sizes and feedback here would be useful as it can show what has been removed by size thresholding.

PR #556, opened 2023-05-09
Sylvia whittle/image stats

### LINKED ISSUE #477: Add output data for whole image
It would be beneficial for TopoStats to output another `.csv` file for quantities associated with the whole image. Two important examples of this are:
- The number of grains
- The density of grains

I imagine there are a lot more metadata that we would want added, perhaps image size, and maximum and minimum heights?

I propose that we made a framework to output such stats like we have for `folder_grainstats`. I am not sure if each of these images should have their own `.csv` or if there should be a `folder_image_metadata.csv`. What do people think?

### LINKED ISSUE #478: Add grain density statistic
For samples such as silicon wafers, being able to calculate the density of grains is very important. This would be stored in the proposed image metadata file, discussed in #477 

Calculating this statistic would be trivial, simply dividing the number of grains by the SI-units area of the image.

### LINKED ISSUE #498: Add roughness measurements
Roughness outputs for images has been requested for our work on materials. Specifically, the RMS output. https://en.wikipedia.org/wiki/Surface_roughness

I think this measurement should be added to the list of statistics that are output in #477 .

SylviaWhittle:
Closes #477 
Closes #478
Closes #498 

This PR adds calculation of statistics that pertain to the whole image being processed, rather than individual grains within it. This has been requested and will be useful in the future, where additional image stats will be required.

I have kept the saving of image stats out of `processing.py` and allowed `processing.py` to pass the results dictionary out to `run_topostats.py` so that the image stats can be properly regression tested.

I am not certain that I have done this in the best way, so please do point out better alternatives to my implementation 😄 

Have I added enough tests? Don't hesitate to suggest more and I'll have a go at writing them.


topostats/roughness.py
```diff
@@ -0,0 +1,17 @@
+"""Contains functions pertaining to the roughness of AFM images."""
+
+import numpy as np
+
+
+def roughness_rms(image: np.ndarray) -> float:
+    """Calculate the rms roughness of a heightmap image.
```
ns-rse, line 7: Is RMS here the [Root Mean Square](https://en.wikipedia.org/wiki/Root_mean_square)? Might be useful to spell this out in the docstring.

topostats/run_topostats.py
```diff
@@ -228,16 +228,30 @@ def main(args=None):
 
     with Pool(processes=config["cores"]) as pool:
         results = defaultdict()
+        image_stats_df_list = []
```
ns-rse, line 231: The `results` is made a dictionary so that its possible to keep track of what image the `result` comes from when it comes to `pd.concat()` the results.

I think this would perhaps be useful to keep track of too so...

```suggestion
        image_stats = defaultdict()
```

topostats/run_topostats.py
```diff
@@ -228,16 +228,30 @@ def main(args=None):
...
-            for img, result in pool.imap_unordered(
+            for img, result, image_stats_df in pool.imap_unordered(
                 processing_function,
                 scan_data_dict.values(),
             ):
                 results[str(img)] = result
                 pbar.update()
+
+                # Add the dataframe to the list for later concatenation
+                image_stats_df_list.append(image_stats_df)
```
ns-rse, line 244: To align with above suggestion...
```suggestion
                image_stats[str(img)] = image_stats_df
```

topostats/roughness.py
```diff
@@ -0,0 +1,17 @@
...
+    Parameters
+    ----------
+    image: np.ndarray
+        2D numpy array of heightmap data to calculate the roughness of
+
+    Returns:
+    float
+        The RMS roughness of the input array.
+    """
+    return np.sqrt(np.mean(np.square(image)))
```
ns-rse, line 17: Just to check this is the metric that is required?

I would have thought "roughness" might measure some element of how much variation there is across an image and so the something like the [Variance](https://en.wikipedia.org/wiki/Variance)  (or it's square root [Standard Deviation](https://en.wikipedia.org/wiki/Standard_deviation) which is the same units as the original heihgts) might be more appropriate as that is quantifying how much variation there is across the sample.

That said scans of biological samples I'd expect (and have seen) tend to be bimodal and such statistics don't provide useful information. Perhaps of more relevance to scans of materials though.

topostats/processing.py
```diff
@@ -249,10 +280,26 @@ def process_scan(
                     # Set tracing_stats_df in light of direction
                     if grains_config["direction"] == "both":
                         grainstats_df = pd.concat([grainstats["below"], grainstats["above"]])
+                        image_stats["grains_number_above"] = grainstats["above"].shape[0]
```
ns-rse, line 283: Here each of the `if:...elif:` statements is doing the same thing, the only thing that differs between them is what is being put in.

This makes it the perfect candidate for a function. But thinking about the `roughness.py` module would this functionality not fit better in the `Grainstats` class as `processing.py` is more about what methods to call and when.

topostats/processing.py
```diff
@@ -317,7 +364,12 @@ def process_scan(
         LOGGER.info(f"[{filename}] Detection of grains disabled, returning empty data frame.")
         results = create_empty_dataframe()
 
-    return image_path, results
+    # Set the index to zero since we are only adding one row. Will crash otherwise since we are not
+    # passing iterables.
+    image_stats_df = pd.DataFrame(image_stats, index=[0])
+    image_stats_df.set_index("filename", inplace=True)
```
ns-rse, line 370: If the below suggestion of making a dictionary of results and `pd.concat()`ing that then this line would no longer be needed.

ns-rse (changes requested):
Code all makes sense, but I wonder if some of the calculations would sit better in the `Grainstats` class, or indeed a class of its own `Imagestats` as `process_scan()` is meant to be more about controlling the processing. 

Some of the code that is repeated could be broken out into small methods that are testable too perhaps.

topostats/roughness.py
SylviaWhittle: So this root-mean-squared roughness calculation is aimed at non-bio samples. An academic I'm working with has assured me that this is what is needed, so I'm taking their word for it. I have checked 😄 👍

topostats/statistics.py
```diff
@@ -0,0 +1,81 @@
...
+        pass
+    try:
+        image_stats["grains_number_above"] = results_df["threshold"].value_counts().get("above", 0)
+        image_stats["grains_per_m2_above"] = image_stats["grains_number_above"] / image_stats["image_area_m2"]
+    except KeyError:
+        pass
+
+    # Convert the dictionary to a dataframe, with "image_name" as the index. Transpose is needed
+    # else the rows and columns will be swapped.
+    image_stats_df = pd.DataFrame.from_dict(image_stats, orient="index").transpose()
```
ns-rse, line 78: I think if you use `orient="columns"` then you shouldn't need to `.transpose()`.

topostats/statistics.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Calculate the RMS roughness of the sample on the flattened image.
+    image_stats["rms_roughness"] = roughness_rms(image=image) * 1e-9
+
+    # Calculate image stats relating to grain statistics. Note that the existence of any of these stats
+    # is not guaranteed
+    try:
+        image_stats["grains_number_below"] = results_df["threshold"].value_counts().get("below", 0)
+        image_stats["grains_per_m2_below"] = image_stats["grains_number_below"] / image_stats["image_area_m2"]
+    except KeyError:
```
ns-rse, line 68: Nice, I hadn't worked out what error might arise and this placates `pylint`.

topostats/statistics.py
```diff
@@ -0,0 +1,81 @@
...
+        dna tracing.
+
+    Returns
+    -------
+    dict
+        Dictionary of image statistics.
+    """
+
+    image_stats = {
+        "Image": filename,
```
ns-rse, line 40: I don't think you need to add the `Image` to the data frame here because you've changed the loop in `run_topostats.py` to save the results into a dictionary which when you `pd.concat` uses the dictionary index as the index for the dataframe.

Might be worth checking but I think this is what happens with the grain/tracing statistics.

topostats/run_topostats.py
```diff
@@ -228,16 +228,30 @@ def main(args=None):
...
                 results[str(img)] = result
                 pbar.update()
+
+                # Add the dataframe to the results dict
+                image_stats_all[str(img)] = individual_image_stats_df
+
+                # Display completion message for the image
+                LOGGER.info(f"[{img.name}] Processing completed.")
+
+    LOGGER.info("Saving image stats dataframe to csv.")
```
ns-rse, line 249: We could log where its saved to with...

```suggestion
    LOGGER.info(f"Saving image stats to : {config['output_dir']}/image_stats.csv.")
```

topostats/roughness.py
ns-rse, line 1: As this is a "statistics" I'd move it to be a function in the `statistics` sub-module.

ns-rse (changes requested):
Thanks for working on this @SylviaWhittle and sorry for not explaining myself clearly in the first place.

This is nice and neat now, we keep code that calculates things out of `process_scan()` and in a test-able function `image_statistics()`.

Would be useful to have a simple test that `image_statistics()` works as expected, including handling of the `KeyError` exceptions.

topostats/statistics.py
SylviaWhittle: I had tried that but it produces an error: 

```
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/Users/sylvi/Documents/TopoStats/topostats/processing.py", line 330, in process_scan
    image_stats = image_statistics(
  File "/Users/sylvi/Documents/TopoStats/topostats/statistics.py", line 79, in image_statistics
    image_stats_df = pd.DataFrame.from_dict(image_stats, orient="columns")
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/frame.py", line 1764, in from_dict
    return cls(data, index=index, columns=columns, dtype=dtype)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/frame.py", line 664, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 493, in dict_to_mgr
    return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 118, in arrays_to_mgr
    index = _extract_index(arrays)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 656, in _extract_index
    raise ValueError("If using all scalar values, you must pass an index")
ValueError: If using all scalar values, you must pass an index
"""

The above exception was the direct cause of the following exception:

  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/bin/run_topostats", line 8, in <module>
    sys.exit(main())
  File "/Users/sylvi/Documents/TopoStats/topostats/run_topostats.py", line 236, in main
    for img, result, individual_image_stats_df in pool.imap_unordered(
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/multiprocessing/pool.py", line 873, in next
    raise value
ValueError: If using all scalar values, you must pass an index
```

I'm by no means experienced with Pandas and so I could be missing something here. I found the transpose method via a tutorial but I can't remember where

topostats/statistics.py
SylviaWhittle: I think it does need it. I tried removing it and...

image

`grainstats.py` also includes the image name:
```
        else:
            grainstats_df = create_empty_dataframe()
        grainstats_df.index.name = "molecule_number"
        grainstats_df["image"] = self.image_name

        return grainstats_df, grains_plot_data
```

topostats/statistics.py
ns-rse: I feel bad for suggesting `pd.DataFrame.from_dict()` now, plain `pd.DataFrame()` should do the trick...

```python
import pandas as pd

test_dict = {"a": 1, "b": 2, "c": 3}
pd.DataFrame.from_dict(test_dict, orient="index").T
Out: 
   a  b  c
0  1  2  3
pd.DataFrame([test_dict])
Out: 
   a  b  c
0  1  2  3
```

Sorry :blush:

topostats/statistics.py
ns-rse: You are of course right, the `image_name` is used in the dictionary which collects results for the parallel processing. We only use the `.values()` of that dictionary when passing to `pd.concat()` though since it only works with objects and doesn't care for the `.keys()`.

ns-rse (approved):
When the tests pass this looks good to go too.

:+1:

ns-rse (approved):
:+1: sorry for the delay.

PR #557, opened 2023-05-11
Adds Mermaid diagram describing extensions to DNA Tracing

ns-rse:
This can be rendered with

```
pip install -e .[docs]
cd docs
make html
```

The output tells you where to find the resulting pages relative to the docs directory (`docs/_build/html/index.html` is the file to open).


MaxGamill-Sheffield (approved):
Looks good

PR #578, opened 2023-05-18
Sylvia whittle/486 area unit tests

### LINKED ISSUE #486: Unit tests for area thresholding
Currently there are no unit tests for area thresholding which excludes grains outside of a specified range.

Some issues have been noted when using area thresholding in practice (hopefully details to follow).

SylviaWhittle:
Closes #486 

This PR adds unit tests for area thresholding functions, `area_thresholding` and `remove_small_objects` in `grains.py`. I think there is something to be said for potentially removing the `remove_small_objects` function since it runs when no area thresholds are set, and so feels a bit counter to what the user might want, but that decision should be left to discussion at a later point.

### DIFF
--- a/tests/test_grains.py
+++ b/tests/test_grains.py
@@ -72,6 +72,45 @@ def test_known_array_threshold(area_thresh_nm, expected) -> None:
 #     assert True
 
 
+def test_remove_small_objects():
+    """Test the remove_small_objects method of the Grains class."""
+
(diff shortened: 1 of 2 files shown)


ns-rse (approved):
Nice succinct tests and I think the renaming of variables makes them a lot clearer, great work, thanks @SylviaWhittle

PR #584, opened 2023-05-24
Sylvia whittle/581 group plots by dir

### LINKED ISSUE #581: Toposums default output should group by folder, and not by image
Currently, when topostats is run, the 'summary distributions' group data by individual spm files, which is not very useful. It would be better if they could be grouped by default based on the folder the spm files are in, i.e. the 'basename'. If all spm files analysed are found within the same folder, there should be no grouping by default.

For example, a user might have two sets of spm files, one under the folder 'DNA_with_protein', the other under the folder 'DNA_without_protein'; when the user runs the data, the plots produced should allow the user compare the distributions of these two data sets, instead of comparing individual files. And if the user only runs topostats within the folder 'DNA_with_protein', no comparison should be made, so the resulting plot would be a simple histogram/kde plot etc.

See above

While I am aware that what I described can be achieved with the notebooks, this isn't obvious to many users. Besides, the default outputs should reflect the most common need of the users, which is to group the distributions by folders, allowing comparison of eg different experimental conditions.

**Additional context**

SylviaWhittle:
Closes #581

As a result of popular request, this PR changes the behaviour of `plotting.py` to group data based on the directory that it comes from, rather than the image.

Example plot:

image

- Changed `melt_data` into a static method, which returns the dataframe, to make testing it easier
- Added tests for the `melt_data` method of TopoSum
- Added `get_paths_relative_to_deepest_common_path` and `convert_basename_to_relative_paths`, which together are used to convert the `basename` column of the dataframe, into having paths relative to the deepest common ancestor rather than global paths, which will mean we don't get unnecessary info in the legends of plots, eg: `usr/sylvi/Documents/topo_data/581-grouped-images/data/subdir1`, `usr/sylvi/Documents/topo_data/581-grouped-images/data/subdir2`, but instead: `subdir1` and `subdir2`.
(these functions also test the limits of how much we should sacrifice conciseness in pursuit of descriptive function names)
- I have edited `conftest.py` so that instead of re-constructing TopoSum objects in many of the tests in `test_plotting.py`, I now just create two instances in `conftest.py`, one for a dataset that only has one unique directory, and one where there are many unique directories, for testing plots where there will be multiple datasets, eg: `test_plot_violin_multiple_images`. I find this change also makes the tests a bit more readable and simple. 
- the legend on the plots is titled 'basename' and this cannot be changed as far as I can tell, since while calling `plt.legend(title='directories')` works for the violin plots, it does not for the hist / kde plots, where it deletes the labels in the legend. There are some posts on SO about this, however they did not work for me when I tried, so for now, unless we want to rename `basename` to a different name in the whole codebase, we are stuck with the legend being titled `basename`.
image


tests/test_plotting.py
```diff
@@ -17,28 +17,74 @@
 RESOURCES = BASE_DIR / "tests" / "resources"
 
 
-def test_df_from_csv(minicircle_all_statistics: pd.DataFrame, toposum_object: TopoSum) -> None:
+def test_melt_data():
```
ns-rse, line 20: Thanks for writing a test for this, bad me for skipping doing so.

topostats/plotting.py
```diff
@@ -272,9 +276,62 @@ def _set_label(self, var: str):
         self.label = self.var_to_label[var]
         LOGGER.debug(f"[plotting] self.label     : {self.label}")
 
-    def summarise_by_image(self):
-        """Summarise statistics by image."""
-        self.summary_data = self.df.groupby(["image", "threshold"]).describe()
+    @staticmethod
+    def get_paths_relative_to_deepest_common_path(paths: list) -> list:
```
ns-rse, line 280: Can be more explicit here with the type hint of `List[Path]` which says its a list of `Path` objects.

SylviaWhittle: Thank you! I did not know about this 😄

topostats/plotting.py
```diff
@@ -177,18 +182,18 @@ def sns_plot(self) -> None:
         fig, ax = self._setup_figure()
         if self.hist and not self.kde:
             outfile = self._outfile("hist")
-            sns.histplot(data=self.melted_data, x="value", bins=self.bins, stat=self.stat, hue=self.image_id)
+            sns.histplot(data=self.melted_data, x="value", bins=self.bins, stat=self.stat, hue="basename")
```
ns-rse, line 185: Hard coding this makes it tricky if people then want to colour/group by different variables.

I think the class `__init__` should take a value for `hue` and in some other plots `x` which become properties. Defaults should be `basename` but its then allows users to specify different groupings should they wish to.

topostats/plotting.py
```diff
@@ -272,9 +278,63 @@ def _set_label(self, var: str):
         self.label = self.var_to_label[var]
         LOGGER.debug(f"[plotting] self.label     : {self.label}")
 
-    def summarise_by_image(self):
-        """Summarise statistics by image."""
-        self.summary_data = self.df.groupby(["image", "threshold"]).describe()
+    @staticmethod
+    def get_relative_paths(paths: List[Path]) -> list:
```
ns-rse, line 282: As this and `convert_basename_to_relative_path` are static methods I was thinking they might be useful functions in `topostats.io` sub-module since they deal with input/output paths as there may be use for them elsewhere.

What do you think?

ns-rse, line 282: Trying to think ahead about typing the return value could also be...

```python
List[str]
```

At some point I'd like to introduce [mypy](https://mypy.readthedocs.io) into the pre-commit hooks and then retrospectively apply it across the code base and then we catch these as we go.

topostats/plotting.py
SylviaWhittle: I think I've addressed this as you suggested, but please do check if I've misinterpreted or missed something

topostats/plotting.py
SylviaWhittle: Done 👍

SylviaWhittle: I was debating whether to do this, but since you flagged it, I might as well, done 👍

topostats/plotting.py
```diff
@@ -218,16 +225,17 @@ def sns_violinplot(self) -> None:
...
         """Melt a dataframe into long format for plotting with Seaborn."""
-        self.melted_data = pd.melt(
-            self.df.reset_index(), id_vars=[self.molecule_id, self.image_id], value_vars=self.stat_to_sum
-        )
-        self.melted_data["variable"] = self.melted_data["variable"].map(self.var_to_label)
-        self.melted_data.rename({"image": "Image"}, axis=1, inplace=True)
-        self.image_id = "Image"
+        melted_data = pd.melt(df.reset_index(), id_vars=["molecule_number", "basename"], value_vars=stat_to_summarize)
+        melted_data["variable"] = melted_data["variable"].map(var_to_label)
+        # melted_data.rename({"image": "Image"}, axis=1, inplace=True)
```
ns-rse, line 233: Could remove this and the next line as they're now redundant.

ns-rse (approved):
Thanks for getting this done and moving the functionality to `io` rather than a static method.

Looks good and ready to merge although I noticed two lines of commented code that could be removed first.

tests/test_plotting.py
```diff
@@ -106,72 +134,60 @@ def test_toposum(summary_config: dict) -> None:
...
     """Test plotting Kernel Density Estimate and Histogram for area for a single image."""
-    _toposum = TopoSum(csv_file=RESOURCES / "minicircle_default_all_statistics.csv", **summary_config)
-    fig, _ = _toposum.sns_violinplot()
+    fig, _ = toposum_object_single_directory.sns_violinplot()
     return fig
 
 
 @pytest.mark.mpl_image_compare(baseline_dir="resources/img/distributions/")
-def test_plot_violin_multiple_images(summary_config: dict, toposum_multiple_images: pd.DataFrame) -> None:
+def test_plot_violin_multiple_images(toposum_object_multiple_directories: TopoSum) -> None:
```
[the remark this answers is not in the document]
SylviaWhittle: Good idea, I hadn't considered this! Done! 👍

topostats/plotting.py
SylviaWhittle: Thanks for spotting this, I knew I would forget to remove these 😆

ns-rse (approved):
Cool, ready to go :+1:

PR #588, opened 2023-05-26
Include comments in config files generated with --create-config-file

### LINKED ISSUE #536: Include field descriptors in autogenerated config file
When using the --create-config-file flag and also in the config file generated at the ends of runs there are none of the comments with the field descriptors that are present in the default config file. I think it would be helpful, especially for new users, to have these comments in the automatically generated file. 

Include the comments in default config file [https://github.com/AFM-SPM/TopoStats/blob/fe3d5c1386bd96f8d9dfc27880b8efd75192b519/topostats/default_config.yaml](https://github.com/AFM-SPM/TopoStats/blob/fe3d5c1386bd96f8d9dfc27880b8efd75192b519/topostats/default_config.yaml) in autogenerated config files.
Alternative could be clearer documentation. 

**Additional context**

SylviaWhittle:
Closes #536 

This PR changes the behaviour of generating config files using the command-line flag `--create-config-file`, so that it uses `shutil` to copy the `default_config.yaml` file directly. This is a trade-off on functionality, but it seems that users would prefer the convenience of having the comments in the file.

If it turns out that the users do not end up liking this change, it should be trivial to reverse it as it's a small change that does not affect any other code as far as I can tell.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -89,9 +89,7 @@ def write_yaml(
     config_yaml = yaml_load(yaml_dump(config))
 
     if header_message:
-        config_yaml.yaml_set_start_comment(
-            f"{header_message} : {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n" + CONFIG_DOCUMENTATION_REFERENCE
-        )
+        config_yaml.yaml_set_start_comment(f"{header_message} : {get_date_time()}\n" + CONFIG_DOCUMENTATION_REFERENCE)
     else:
         config_yaml.yaml_set_start_comment(
             f"Configuration from TopoStats run completed : {get_date_time()}\n" + CONFIG_DOCUMENTATION_REFERENCE
@@ -103,7 +101,7 @@ def write_yaml(
             LOGGER.error(exception)
 
 
-def write_config_with_comments(config: str, filename: str = "config.yaml"):
+def write_config_with_comments(config: str, output_dir: Path, filename: str = "config.yaml") -> None:
     """
     Create a config file, retaining the comments by writing it as a string
     rather than using a yaml handling package.
@@ -112,20 +110,22 @@ def write_config_with_comments(config: str, filename: str = "config.yaml"):
     ----------
     config: str
         A string of the entire configuration file to be saved.
+    output_dir: Path
+        A pathlib path of where to create the config file.
     filename: str
         A name for the configuration file. Can have a ".yaml" on the end.
     """
 
     if ".yaml" not in filename and ".yml" not in filename:
-        create_config_path = "./" + filename + ".yaml"
+        create_config_path = output_dir / f"{filename}.yaml"
     else:
-        create_config_path = "./" + filename
+        create_config_path = output_dir / filename
 
-    with open(f"./{create_config_path}", "w", encoding="utf-8") as f:
+    with open(f"{create_config_path}", "w", encoding="utf-8") as f:
         f.write(f"# Config file generated {get_date_time()}\n")
         f.write(f"# {CONFIG_DOCUMENTATION_REFERENCE}")
         f.write(config)
-    LOGGER.info(f"A sample configuration has been written to : ./{filename}")
+    LOGGER.info(f"A sample configuration has been written to : {str(create_config_path)}")
     LOGGER.info(CONFIG_DOCUMENTATION_REFERENCE)
 
 
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -20,6 +20,9 @@ from topostats.logs.logs import LOGGER_NAME
 LOGGER = logging.getLogger(LOGGER_NAME)
 
 
+CONFIG_DOCUMENTATION_REFERENCE = """For more information on configuration and how to use it:
+# https://afm-spm.github.io/TopoStats/main/configuration.html\n"""
+
 # pylint: disable=broad-except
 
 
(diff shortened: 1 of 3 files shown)


topostats/run_topostats.py
```diff
@@ -163,12 +164,12 @@ def main(args=None):
 
     # Write sample configuration if asked to do so and exit
     if args.create_config_file:
-        write_yaml(
-            config,
-            output_dir="./",
-            config_file=args.create_config_file,
-            header_message="Sample configuration file auto-generated",
-        )
+        default_config_path = pkg_resources.path(__package__, "default_config.yaml")
```
ns-rse, line 167: I understand why this is here but I think we can make things a little simpler by changing some code earlier on.

Currently on lines 146-150 we have...

```python
    if args.config_file is not None:
        config = read_yaml(args.config_file)
    else:
        default_config = pkg_resources.open_text(__package__, "default_config.yaml")
        config = yaml.safe_load(default_config.read())
```

...and so we have two lines that are very similar here as they both have `pkg_resources.open_text(__package__, "default_config.yaml")`.

If we change the logic of the earlier code we only need to have one line for setting the path to the `default_config.yaml`...

```python
    if args.config_file is None:
        default_config = pkg_resources.open_text(__package__, "default_config.yaml")
        config = yaml.safe_load(default_config.read())
    else:
        config = read_yaml(args.config_file)
```

...and the above line is no longer needed (although line 172 would need changing to use `default_config` rather than `default_config_path`).

topostats/run_topostats.py
```diff
@@ -163,12 +164,12 @@ def main(args=None):
...
+        if ".yaml" not in args.create_config_file and ".yml" not in args.create_config_file:
+            create_config_path = "./" + args.create_config_file + ".yaml"
```
ns-rse, line 169: Not tested this on M$-Win but would the hard coding of `./` cause any problems? I've a vague memory that the Windows Linux Shell does interpret/handle these correctly. If not perhaps using `pathlib.Path("./")` might mitigate the problem (if there is one!).

ns-rse (commented):
Thanks for getting this one sorted too @SylviaWhittle 

I've suggested a slight change to avoid duplicated lines simply by changing earlier logic.

I think its a bit of a shame to lose the file header which indicates when the sample config was generated but date/time isn't too reliable given people could be on different branches anyway and besides we ask for the config used at any given point to be included in Bug Reports anyway.

topostats/run_topostats.py
SylviaWhittle: I could be misunderstanding you. 
```python
pkg_resources.open_text(__package__, "default_config.yaml")
```
returns a string of the contents of the config file, which is not usable by `shutil.copy`, which takes a file system path. This could work if I manually saved to the file using `with open() as f`, is this what you intended?

SylviaWhittle:
Thanks for the comments @ns-rse, I've just pushed a re-jig of it, where I use your suggestion above, and I create a function in `io.py`, `write_config_with_comments` to handle the file writing.

I had to add a line to ensure users are not running with `--create-config-file` and `--config`, as this would leave `default_config` undefined before its call.

I've also moved things about a bit with the documentation, so that we keep the `# Config file generated 2023-05-27 01:01:12` and documentation links. 
There were some 404 links in the code that I have removed, leaving only a link that works. This should also mean that it should be easy to have all documentation hints standardised and removed link duplication in the codebase

Let me know what you think and I'll add tests if you approve (don't worry I hadn't forgotten about adding tests). 👍

topostats/io.py
```diff
@@ -67,17 +87,14 @@ def write_yaml(
...
     config = path_to_str(config)
     config_yaml = yaml_load(yaml_dump(config))
-    documentation_reference = (
-        "For more information on configuration : https://afm-spm.github.io/TopoStats/main/configuration.html"
-    )
+
     if header_message:
         config_yaml.yaml_set_start_comment(
-            f"{header_message} : {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n" + documentation_reference
+            f"{header_message} : {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n" + CONFIG_DOCUMENTATION_REFERENCE
```
ns-rse, line 93: Could use the `get_date_time()` function here as you do below.

topostats/io.py
```diff
@@ -86,6 +103,32 @@ def write_yaml(
...
+
+    Parameters
+    ----------
+    config: str
+        A string of the entire configuration file to be saved.
+    filename: str
+        A name for the configuration file. Can have a ".yaml" on the end.
+    """
+
+    if ".yaml" not in filename and ".yml" not in filename:
```
ns-rse, line 119: Good to have this eventually caught :+1:

topostats/io.py
```diff
@@ -86,6 +103,32 @@ def write_yaml(
             LOGGER.error(exception)
 
 
+def write_config_with_comments(config: str, filename: str = "config.yaml"):
```
ns-rse, line 106: Minor pedantic typehint for the return value.

```suggestion
def write_config_with_comments(config: str, filename: str = "config.yaml") -> None:
```

topostats/io.py
```diff
@@ -86,6 +103,32 @@ def write_yaml(
...
+    Parameters
+    ----------
+    config: str
+        A string of the entire configuration file to be saved.
+    filename: str
+        A name for the configuration file. Can have a ".yaml" on the end.
+    """
+
+    if ".yaml" not in filename and ".yml" not in filename:
+        create_config_path = "./" + filename + ".yaml"
```
ns-rse, line 120: Might be overkill using `Path.cwd()` here and under the `else:` but could.

```suggestion
        create_config_path = Path.cwd() / f"{filename}.yaml"
```

ns-rse (approved):
Neat, I like this solution of using a string object rather than `shutil.copy` and that you've retained the file header at the same time.

Not sure its absolutely necessary to use `Path.cwd()` may be overkill. :shrug:

ns-rse (approved):
Looks good to go when the tests complete. :+1:

PR #590, opened 2023-05-30
Make channel a free text field

### LINKED ISSUE #589: Make channel a free string field
The `channel` parameter in the config is restricted to channels that we as developers have set to be allowed. I have had complaints that TopoStats crashes when users try to use less-known channels because of this. I suggested that they could tell us and we could add those channels to the list, however there was pushback, saying that there are so many channels that it would be impractical to do this, and as the userbase of TopoStats grows, this may not be a sustainable solution, especially as this would add a wait time for different labs / users before they could process their images.

The proposed solution is to make the `channel` field a free string field, not restricted by a list of pre-defined options that are validated against in `validation.py`. Is this reasonable? I don't see an issue with it, but could be overlooking something.

SylviaWhittle:
Closes #589 

Small PR to make the `channel` field of the config into a free text field, as it has been requested by several users.

### DIFF
--- a/topostats/validation.py
+++ b/topostats/validation.py
@@ -53,41 +53,7 @@ DEFAULT_CONFIG_SCHEMA = Schema(
             ".gwy",
             error="Invalid value in config for 'file_ext', valid values are '.spm', '.jpk', '.ibw' or '.asd'.",
         ),
-        "loading": {
-            "channel": Or(
-                "ZSensor",
-                "",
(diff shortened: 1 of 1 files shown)


ns-rse (approved):
Saves having to work out what all the channels are called. 

We already have code in `LoadScans` courtesy of @MaxGamill-Sheffield which tells people if the channel they've specified is not present.

PR #597, opened 2023-06-02
Sylvia whittle/596 zscale grains

### LINKED ISSUE #596: Z range of cropped images not set by config file
Currently the z scale of the cropped grains that are generated in the `all` image set is relative, set between the highest and lowest points of the image rather than by the config file. 

It would be better if the z scale of these cropped images could also be set by the z range in the config file.

SylviaWhittle:
Closes #596 

This PR makes it so that the `z-range` set in the config file is applied to all non-binary images, not just the `core` image set. This has been done by editing `update_plotting_config` to no longer remove the `zrange` item from the `plot_dict` dictionaries and instead, sets the item to `[None, None]` for binary images, and sets it to the user defined range for non-binary images.

I have updated the test `test_update_plotting_config` in `test_utils.py` accordingly to check that binary and non-binary images behave as expected.

The new outputs with a `z-range` of `[-3, 4]` applied:
image
image
image
image

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -42,8 +42,14 @@ def test_update_config(caplog) -> None:
 
 
 @pytest.mark.parametrize(
-    "image_name, core_set, title, zrange",
-    [("extracted_channel", False, "Raw Height", KeyError()), ("z_threshed", True, "Height Thresholded", [0, 3])],
+    "image_name, core_set, title, image_type, zrange",
+    [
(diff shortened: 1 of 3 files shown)


derollins:
Ran this code, it works and produces the results I wanted from issue #596

ns-rse (approved):
Looks good and setting to `[None, None]` is the way it should have been in the first place. Thanks @SylviaWhittle

PR #599, opened 2023-06-05
Fixing tpyo's adding in `image`

ns-rse:


SylviaWhittle (approved):
Thank you for updating the docs 😄

PR #603, opened 2023-06-13
Fix typo "contour_lengths" -> "contour_length"

SylviaWhittle:
In several places in the codebase, I missed that `contour_lengths` had not been updated to `contour_length`. As @derollins has found out, this can cause a crash in the plotting stages, when `TopoSum` tries to fetch data about the `contour_lengths` column of the results dataframe, but doesn't find one.

This PR fixes the typo and resolves the issue 🤞

I'm extremely confused about how this passed tests, as I have had to update a test to make this work 🤔


ns-rse (approved):
Thank you @SylviaWhittle and apologies for not doing all of these changes in the recent PR and causing unnecessary work for you.

PR #605, opened 2023-06-15
Sylvia whittle/plotting traces

SylviaWhittle:
Currently a testing ground, I'll fill this in soon


topostats/plottingfuncs.py
```diff
@@ -166,6 +166,7 @@ def plot_and_save(self):
         """
         fig, ax = None, None
         if self.save:
+            # Only continue if the image is in the core set, or the image_set config parameter is set to plot all images
```
ns-rse, line 169: :+1: comments are good!

topostats/tracing/dnatracing.py
```diff
@@ -121,13 +126,28 @@ def trace_dna(self):
...
             self.linear_or_circular(self.ordered_trace)
             self.get_fitted_traces()
             self.get_splined_traces()
             # self.find_curvature()
             # self.saveCurvature()
             self.measure_contour_length()
             self.measure_end_to_end_distance()
+
+            mask = np.zeros(self.image.shape)
+            for coord in self.ordered_trace:
```
ns-rse, line 139: Shouldn't need this I don't think.

Under `trace_image()` we call `prep_arrays()` and that crops the padded bounding boxes out of the original image and the labelled region. Before returning them though it [flips every labelled region to be `1` instead of its label](https://github.com/AFM-SPM/TopoStats/blob/08f807cc6d0acf5e15c7aabb63491f20bee622be/topostats/tracing/dnatracing.py#L781) and so `self.grain` should be a mask of `0` and `1` that can be passed straight into `Images()`.

topostats/tracing/dnatracing.py
```diff
@@ -121,13 +126,28 @@ def trace_dna(self):
...
+                mask[coord[0], coord[1]] = 1
+            Images(
```
ns-rse, line 141: I realise this is work in progress and suspect this is for development and intended to be removed since plotting of traces is done from `processing.process_scan()`?

topostats/plottingfuncs.py
```diff
@@ -201,6 +204,7 @@ def save_figure(self):
                 vmin=self.zrange[0],
                 vmax=self.zrange[1],
             )
+            # If masked array, apply that on top of the plot
```
ns-rse, line 207: Nice, it didn't occur to me that virtually nothing needs doing to the `Images()` class as the traces are in essence just refined/slimmed down grain masks and so the functionality to plot them over the image is already there.

tests/tracing/test_dnatracing_single_grain.py
```diff
@@ -413,6 +429,54 @@ def test_trace_image(image: str, skeletonisation_method: str, cores: int, statis
...
+    #     pkl.dump(all_trace_data, f)
+    # with open(RESOURCES / f"test_trace_image_trace_plot_data_multigrain_{skeletonisation_method}.pkl", "wb") as f:
+    #     pkl.dump(trace_plot_data, f)
+
+    with open(trace_data_expected_file, "rb") as f:
+        trace_data_expected = pkl.load(f)
+    with open(trace_plot_data_expected_file, "rb") as f:
+        trace_plot_data_expected = pkl.load(f)
+
+    print(trace_plot_data)
```
ns-rse, line 445: Debugging

topostats/tracing/dnatracing.py
```diff
@@ -773,13 +894,25 @@ def prep_arrays(image: np.ndarray, labelled_grains_mask: np.ndarray, pad_width:
...
     region_properties = skimage_measure.regionprops(labelled_grains_mask)
     # Subset image and grains then zip them up
-    cropped_images = [crop_array(image, grain.bbox, pad_width) for grain in region_properties]
-    cropped_images = [np.pad(grain, pad_width=pad_width) for grain in cropped_images]
-    cropped_masks = [crop_array(labelled_grains_mask, grain.bbox, pad_width) for grain in region_properties]
-    cropped_masks = [np.pad(grain, pad_width=pad_width) for grain in cropped_masks]
+    cropped_images = []
+    cropped_positions = []
+    cropped_masks = []
+    for grain in region_properties:
```
ns-rse, line 900: I think we might be able to use the already existing `.bbox` that each `grain` within `region_properties` already has such that `crop_array()` retains its single responsibility and doesn't return both the `cropped_array` and the `cropped_position`.

We can keep the existing dictionary comprehension as was and add in...

```python
    cropped_positions = [(grain.bbox[0], grain.bbox[1]) for grain in region_properties]
```

As well as not adding extra functionality to `crop_array()` it also saves having to subtract `pad_width` since the original `bbox` aren't padded.

Arguably this could be made a function of its own...

```python
def extract_grain_anchor(region_properties: List(RegionProperties) -> list:
    """Extract the anchor (min_row, min_col) from all labelled regions which is used to align individual traces over the original image.

    Parameters
    ==========
    region_properties: List(RegionProperties)
        A list of region properties returned by skimage.measure.regionprops()

    Returns
    =======
    List(Tuple):
        A list of tuples of the min_row, min_col of each bounding box.
    """"
    return [(grain.bbox[0], grain.bbox[1]) for grain in region_properties]
```

topostats/tracing/dnatracing.py
```diff
@@ -723,11 +787,16 @@ def trace_image(
     # Check both arrays are the same shape
     assert image.shape == grains_mask.shape
 
-    cropped_images, cropped_masks = prep_arrays(image, grains_mask, pad_width)
+    cropped_images, cropped_masks, grain_crop_coordinates = prep_arrays(image, grains_mask, pad_width)
```
ns-rse, line 790: If getting the `grain_crop_coordinates` is done using a dedicated function then would also need...

```python
    grain_crop_coordinates = extract_grain_anchor(grains_mask)
```

SylviaWhittle: Thanks so much for the feedback on the PR @ns-rse!

Not sure I'm understanding your comment completely, but it seems to me that the crop coordinates/offsets need to be calculated and extracted from the crop_array function because of the padding that crop_array adds being unpredictable due to it increasing the size of the cropped image (and therefore the appropriate grain coordinate/offset) by different amounts depending on if a grain is near the border, so it has to calculate the appropriate offsets to return inside the function.

So I don’t think I could make a separate function for calculating the grain offsets without re-writing the same image-edge proximity logic inside crop_array..?

And so `prep_arrays` needs to handle it because it contains the call to `crop_array`?

SylviaWhittle: Here's a comparison of the current method and using the bounding boxes with
```python
[(grain.bbox[0], grain.bbox[1]) for grain in region_properties]
```
current method:
image

bounding boxes:
image

the bounding boxes traces are offset by the pad_width but only when far from the top left image borders as defined by `crop_array`, so I found I could only get it to behave right if `crop_array` told us by how much it pads each grain (and thus the offset)

It is really messy 😞 but  I wasn't able to think of a better solution myself

topostats/tracing/dnatracing.py
SylviaWhittle: See above comment about this 👍

topostats/tracing/dnatracing.py
```diff
@@ -723,11 +787,16 @@ def trace_image(
...
     LOGGER.info(f"[{filename}] : Calculating statistics for {n_grains} grains.")
     n_grain = 0
     results = {}
+    all_global_traces = {}
+    all_trace_heights = {}
+    all_trace_cumulative_distances = {}
+    trace_plot_data = {}
+    trace_plot_data["cropped_grains"] = {}
     for cropped_image, cropped_mask in zip(cropped_images, cropped_masks):
         result = (
```
ns-rse, line 801: Don't need these parentheses, it puts the dictionary within a tuple and means you then have to index `result[0]` to get at the actual results.

PR #609, opened 2023-06-23
Correcting padding of test grains for tracing

### LINKED ISSUE #606: Correct single grain images for dnatracing tests and investigate padding
Currently the images were cropped using their bounding box and then padded them by a fixed amount as I encountered errors when subsequently trying to trace the images. The arrays in use in the test suite at present are...

image
image

(Its evident where the padding with zeros occur as its the blue uniform border).

These are poor choices and what I should have done was expand the crop beyond the bounding box and then written tests as in the below images..

image
image

This reflects better the workflow as currently there are two cropping stages when processing an image, the first to extend the boundary by a fixed amount (desirable) and this will stop when it reaches the edge of the image, and then a second padding with zeros is _always_ applied. This second padding probably isn't necessary and was a lack of clear thinking on my behalf.

Tests will need updating in light of these changes and possibly the second padding step as described can be removed.

ns-rse:
Closes #606

* Split testing of single grains and multiple grains into separate test files.
* Single grain tests are now padded from the original `minicircle.spm` by a fixed border that expands the bounding box rather than padding with `0`s to give more realistic single grain images.
* Test for `ValueError` being raised if arrays are not the same shape (unlikely but you never know!).
* Splits padding of arrays out of crop_arrays() into its own function along with tests.
* Calculates the (padded) anchor for images and uses these in the construction of a mask of the resulting traces from the individual grains.
* `trace_grain()` returns the `ordered_trace` which in turn means...
* `trace_image()` now returns a dictionary with keys for `statistics`, `ordered_traces` and `image_trace`.
* The `trace_mask()` function takes the `ordered_traces` and creates the image mask of the traces.
* Padding is only done twice, the first time takes a bounding box and crops the original image heights and the mask. The second pads a border with zeros for grains that are close to the edge. This introduced the need to calculate the `grain_anchor()` of the original bounding box _and_ adjust this location when reconstructing a mask of traces to overlay on the original image. Some helper functions are introduced to do that (`trace_mask()`, `trim_array()` and `adjust_coordinates()`)
* The `contour_length` of the one grain `below` has changed, its a bit shorter. Not entirely sure why this is present in the first place (but it is actually useful as it covers that direction and negates the need for separate images to test).


topostats/processing.py
```diff
@@ -266,13 +266,16 @@ def process_scan(
...
+                                tracing_results = trace_image(
                                     image=filtered_image.images["gaussian_filtered"],
                                     grains_mask=grains.directions[direction]["labelled_regions_02"],
                                     filename=filename,
                                     pixel_to_nm_scaling=pixel_to_nm_scaling,
                                     **dnatracing_config,
                                 )
+                                tracing_stats[direction] = tracing_results["statistics"]
+                                # ordered_traces = tracing_results["ordered_traces"]
+                                # image_trace = tracing_results["image_trace"]
```
SylviaWhittle, line 278: Commented out code accidentally left in, or is this courteously hinting at how I can add plotting after this PR? 😄

topostats/processing.py
```diff
@@ -311,7 +314,7 @@ def process_scan(
     else:
         LOGGER.info(f"[{filename}] Detection of grains disabled, returning empty data frame.")
         results = create_empty_dataframe()
-    return image_path, results
+    return image_path, results  # , ordered_traces, image_trace
```
SylviaWhittle, line 317: As above

topostats/tracing/dnatracing.py
```diff
@@ -721,40 +721,129 @@ def trace_image(
...
+            cropped_mask,
+            pixel_to_nm_scaling,
+            filename,
+            min_skeleton_size,
+            skeletonisation_method,
+            n_grain,
         )
         LOGGER.info(f"[{filename}] : Traced grain {n_grain + 1} of {n_grains}")
-        results[n_grain] = result[0]
+        ordered_traces.append(result.pop("ordered_trace"))
```
SylviaWhittle, line 746: Very elegant 😲

SylviaWhittle (approved):
This is great, thank you so much @ns-rse . I feel this clarifies everything wonderfully and makes it extremely modular and therefore easy to implement new features, the docstrings have been super helpful in allowing me to understand what is happening. The data is handled elegantly with no messy returns from functions too, and everything has one job. Rigorously tested too. Looking forward to adding the plotting and trace saving to this, it will be trivial!

topostats/processing.py
ns-rse: I left that in on purpose, figured you would be returning these and need them but `pylint` complained that they were assigned but not used so I just `#` them.

PR #622, opened 2023-07-12
User facing channel names swapped for bts channel names

derollins:
This means the reported available channels in the logger error message are actually what you need to put into the config file.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -497,7 +497,6 @@ class LoadScans:
             # trying to return the error with options of possible channel values
             labels = []
             for channel in [layer[b"@2:Image Data"][0] for layer in scan.layers]:
-                # channel_name = channel.decode("latin1").split(" ")[1][1:-1]
                 channel_description = channel.decode("latin1").split('"')[1]  # incase the blank field raises quesions?
                 labels.append(channel_description)
             LOGGER.error(f"[{self.filename}] : {self.channel} not in {self.img_path.suffix} channel list: {labels}")
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -497,8 +497,8 @@ class LoadScans:
             # trying to return the error with options of possible channel values
             labels = []
             for channel in [layer[b"@2:Image Data"][0] for layer in scan.layers]:
-                #channel_name = channel.decode("latin1").split(" ")[1][1:-1]
-                channel_description = channel.decode('latin1').split('"')[1] # incase the blank field raises quesions?
+                # channel_name = channel.decode("latin1").split(" ")[1][1:-1]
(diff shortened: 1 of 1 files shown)


MaxGamill-Sheffield (approved):
Looks good and tested with channel description outputs from my a cat image.

topostats/io.py
```diff
@@ -497,9 +497,9 @@ def load_spm(self) -> tuple:
             # trying to return the error with options of possible channel values
             labels = []
             for channel in [layer[b"@2:Image Data"][0] for layer in scan.layers]:
-                channel_name = channel.decode("latin1").split(" ")[1][1:-1]
-                # channel_description = channel.decode('latin1').split('"')[1] # incase the blank field raises quesions?
-                labels.append(channel_name)
+                # channel_name = channel.decode("latin1").split(" ")[1][1:-1]
```
ns-rse, line 500: Do we need to retain this line if the correct value should always be `channel_description`?

If not this line can be removed.

The code will still be present in the Git history and referenceable by looking at the Git blame for the line that remains, tracing it back to the commits and then its possible to see what changed if the old value is required.

ns-rse (approved):
Looks good and thanks for getting stuck in and making your first PR @derollins .

On reading this I wondered if this might help with the issue #619 and having seen @MaxGamill-Sheffield comment it seems it is.

topostats/io.py
ns-rse, line 500:
```suggestion

```

ns-rse: Ok, two approaches @derollins 

1) You can if you want directly commit the above suggestion to this Pull Request. The Pull Request has been approved already, this may remove the approval if so it will require approving again. This change won't be reflected on your local branch. That isn't necessarily a problem though as you probably won't touch this branch again after its merged and can delete it.

2) Make the change on your local branch (on your laptop) and push it to GitHub, it will be tagged onto and included in this Pull Request. I would expect this to remove the approval too and it would then require re-approval.

Once re-approvals have been made you can go ahead and "Merge" the pull request, on doing so you'll be offered the opportunity to delete the branch on GitHub which is perfectly fine to do.

ns-rse (commented):
Made a suggestion about how to handle the line removal.

ns-rse (approved):
Brilliant, thanks @derollins this looks ready to merge now. :+1: 

/cc @neumagen this may solve your channel issue. You'll have to setup a new virtual(/conda) environment and install TopoStats from Git to try it out, but we have instructions on how to do this on the [website](https://afm-spm.github.io/TopoStats/main/installation.html#cloning-from-github).

If you encounter any problems you can ask for support in the [Installation](https://github.com/AFM-SPM/TopoStats/discussions/categories/installation) discussion category.

Issue #624, opened 2023-07-12
Notebooks need updating to reflect DNAtracing changes


SylviaWhittle:
@neumagen (I think I got your username right?) has very kindly pointed out that the notebook `00-Walkthrough-minicircle.ipynb` fails to run as it tries to import `traceStats` which no longer exists. I have made a fix and will make a PR shortly but making a note here just in case I forget.

There is also an issue in that one of our test files still has the old "contour_lengths" name which causes `02-Summary-statistics-and-plots.ipynb` to fail. I will amend this in the PR too.

SylviaWhittle:
Resolved in #625, I forgot to tag this issue.

PR #625, opened 2023-07-12
Update notebooks to fix dnatracing issue

SylviaWhittle:
This PR is a quick patch to address the issue flagged by a user regarding notebooks 00 and 02. #624 

Review request:
- It runs on my system but I may have omitted something in my haste. If you could run notebooks 00 and 02 and have a quick skim through the markdown cells that would be greatly appreciated.

Changes:
- Changed the `dnatracing.py` usage in notebook 00 to use the new `trace_image()` function instead of the old method of processing images.
- Ammended the column name for the example statistics output `.csv` file that notebook 02 calls upon to demonstrate statistics plotting.


ns-rse (approved):
Brilliant, thanks for updating these Notebooks @SylviaWhittle and thanks for trying them out and reporting problems @neumagen, very much appreciated.

PR #628, opened 2023-07-17
Add parameter for test_trace_mask for pad_width = 2

### LINKED ISSUE #627: Expand testes for trace_mask() to cover padding > 1
The function `topostats.tracing.dnatracing.trace_mask()` places individual grains that have been skeletonised back onto the original image based on the anchor and the user specified `pad_width` parameter.

The test `tests.tracing.test_dnatracing.py::test_trace_mask()` is only parameterised to work with a `pad_width=[0|1]`. The test should have additional instances where the `pad_width>1` (i.e. `2`) to ensure that the co-ordinates of skeletonised grains are placed back on the images in the correct location, regardless of the `pad_width` value.

ns-rse:
Closes #627

Adds a test for `trace_mask()` where the `pad_width=2` to ensure that grains are aligned in the correct position when reconstructing the skeletonised mask from individual grains.

Could add additional tests for `pad_width > 2` but the code is I feel sufficiently tested by this.


SylviaWhittle (approved):
Does what it says on the tin and the test contents make sense 😄 

Might want to let @MaxGamill-Sheffield have a look at this before merging though as he requested this.

ns-rse:
Thanks @SylviaWhittle, I agree, would be great if you could check this works for your use case @MaxGamill-Sheffield .

ns-rse:
On review I realised I could shorten this by defining the target array outside of the `@pytest.mark.parameterize()` chunk.

MaxGamill-Sheffield:
Is the padding expected to have an effect on the outcome? Other than not hitting errors around pixels around the boarder.

I've seen that some molecules aren't found/plotted when using a padding of 1 and 251 for the image. This could be intentional depending on what might affect the processing:
1 pad
251 pad

ns-rse:
Padding is meant to allow the detection of grains that are near the border.

Which way round are the graphs in terms of padding? 1 first and 251 second?

I got the impression from your original question that raised this that you had some work where you had grains which were traced using the new method and you were going to embark on putting them back as overlay on the original does the function work for that outside of the pipeline?

The bottom image also has traces where some of the masks differ they're not "lasso" like (by which I mean a loop on the end of a straight line). What skeletonisation method was used in these (`topostats` / `zhang` / `lee` / `thin`)?

MaxGamill-Sheffield:
> Padding is meant to allow the detection of grains that are near the border.

Ahhh I think I got it! So it may affect the following calculation as the gaussian filter that is applied to a 10x10 image will appear different than the 10x10 image padded by 251 in the gaussian image result. Thus this may have knock-on effects.

> Which way round are the graphs in terms of padding? 1 first and 251 second?

First is 1, second is 251 

> I got the impression from your original question that raised this that you had some work where you had grains which were traced using the new method and you were going to embark on putting them back as overlay on the original does the function work for that outside of the pipeline?

Sadly not for my images as there was a need to modify the whole `trace_dna` pipeline to handle the multiple molecules that could be found in a single grain. I've made my own which does each step from skeletonisation to splining, but it required some tinkering

> The bottom image also has traces where some of the masks differ they're not "lasso" like (by which I mean a loop on the end of a straight line). What skeletonisation method was used in these (`topostats` / `zhang` / `lee` / `thin`)?

It uses `topostats` but as there's no height info in `tracing_funcs.py` I'm unsure how this downstream differs

PR #633, opened 2023-07-26
Speed up tests

### LINKED ISSUE #508: Speed up tests
Tests take a long time. There are two areas that this can be done (they should be treated as separate tasks/issues, capturing here for now).

* [ ] Trim down the array from `minicircle.spm` to a smaller array to speed up all the processing steps.
* [ ] We can leverage [Python caching in GitHub Actions](https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/) to speed up installation of environments.
* [ ] Reduce the number of times scar removal runs, this should be done by modifying the configuration in `tests/conftest.py` where `@pytest.fixtures` get setup. Good idea @SylviaWhittle .

SylviaWhittle:
Closes #508 

This PR is aimed at increasing the speed of tests. It does this by replacing the `(1024, 1024)` `minicircle.spm` testing image, with a slice of this image, namely `minicircle[-296:,-296:]` and compressed down from `(256, 256)` to `(64, 64)`. This image is saved in the ✨`.topostats`✨  file format instead of `.spm`.

I have checked that it flattens appropriately, grains are detected, and traces are produced despite this low resolution. (It also looks really cute).
image
image
image

## Optimisation

Previously, running the automated test suite on GitHub would take up to 46 minutes:
image

With this PR, the tests now take at maximum, 7 minutes:
image

Which is a 6.5x speed increase (or 85% decrease in time taken).

Previously running locally on my Macbook with one core took 612 seconds:
image

Now, running the tests locally on my Macbook with one core, it takes 55 seconds:
image

The bulk of the testing time is taken up by (courtesy of the `--duration=0` `pytest` flag):
```
============= slowest durations ==============
7.71s call     tests/test_processing.py::test_save_format[png]
7.66s call     tests/test_processing.py::test_save_cropped_grains[all-True]
5.74s call     tests/test_processing.py::test_save_format[tif]
4.31s call     tests/test_run_topostats.py::test_run_topostats_process_debug
3.86s call     tests/test_run_topostats.py::test_run_topostats_process_all
1.88s call     tests/test_plottingfuncs.py::test_high_dpi
0.94s call     tests/test_io.py::test_load_scan_gwy
```

If we look at the timings from before
```
============= slowest durations ===============
50.61s call     tests/test_processing.py::test_save_format[png]
49.70s call     tests/test_processing.py::test_save_cropped_grains[all-True]
45.45s call     tests/test_processing.py::test_save_format[tif]
39.00s call     tests/test_run_topostats.py::test_run_topostats_process_all
38.41s call     tests/test_run_topostats.py::test_run_topostats_process_debug
37.48s call     tests/test_processing.py::test_process_scan_both
35.82s call     tests/test_processing.py::test_process_stages[True-True-True-True-Traced grain 21 of 21-Combining ['above'] grain statistics and dnatracing statistics]
34.78s call     tests/test_processing.py::test_process_scan_above
34.26s call     tests/test_processing.py::test_save_cropped_grains[core-False]
33.38s call     tests/test_processing.py::test_process_scan_below
32.57s call     tests/test_processing.py::test_process_stages[True-True-True-False-Processing grain-Calculation of DNA Tracing disabled, returning grainstats data frame.]
32.02s call     tests/test_processing.py::test_process_stages[True-True-False-False-Calculation of grainstats disabled, returning empty dataframe.-24-labelled_image_bboxes]
31.69s call     tests/test_processing.py::test_process_stages[True-False-False-False-Detection of grains disabled, returning empty data frame.-15-gaussian_filtered]
30.87s call     tests/test_processing.py::test_run_filters
6.45s call     tests/test_processing.py::test_process_scan_align_grainstats_dnatracing
3.29s call     tests/test_processing.py::test_process_scan_no_grains
```

The culprits are the same, but we have significantly reduced their length in this PR.

If we want further optimisation, we could consider dropping `test_processing.py::test_save_format` which would shorten the test timing by a further 25% (55s to 40s locally), and I'm not sure that testing the image output file format by running `processing` is entirely necessary, however at this point further optimisation might be silly. Thoughts appreciated.

## A note on test validity.
Given that I have changed the main testing image from the full `minicircle.spm` to a cropped, compressed 64x64 region of `minicircle.spm`, I wanted to make sure that the tests were all legitimate and made sense, as if I blindly forced all the tests to succeed, I might be inadvertently ruining some of the tests.

To prevent this, I plotted every image for every test to make sure that each test does what it should do. An example of where this was necessary is in `tests/test_grains_minicircle.py::test_remove_small_objects` where I found that the new image had not got any small grains to remove, and so the test verified nothing, despite passing. I thus updated the height thresholds and size thresholds to allow the proteins through so that then the test can verify that they are removed.

I believe I have ensured that all the changes to the tests do not prevent them from testing what they were intended to test, but if you are skeptical then please do ask. It's better that I spend time now verifying that tests are still functioning (not passing, but testing what they should) than to find out later down the line that things are broken 👍

## Other notes

- Fixed several type-hints that were inaccurate in `conftest.py`.
- TopoSum, when run after `process_scan`, no longer takes the command-line arguments. This is essential, as if you pass the flag `--file_ext` for `process_scan`, as, say, `.spm`, then TopoSum would also take that as the file extension to save images as, and config validation would fail. So now TopoSum will only take command line arguments when run directly.


tests/conftest.py
```diff
@@ -557,16 +557,16 @@ def minicircle_grain_threshold_abs(minicircle_grains: np.array) -> Grains:
 
 
 @pytest.fixture
-def minicircle_grain_mask(minicircle_grain_threshold_otsu: Grains) -> Grains:
+def minicircle_grain_mask(minicircle_grain_threshold_abs: Grains) -> Grains:
```
ns-rse, line 560: Why the switch from `otsu > abs` for threhsolding? Is it so that you can get the necessary small grains to test removal of small objects?

SylviaWhittle: Yeah 😄

tests/test_io.py
```diff
@@ -318,11 +318,11 @@ def test_save_folder_grainstats(tmp_path: Path) -> None:
     assert Path(out_path / "processed" / "folder_grainstats.csv").exists()
 
 
-def test_load_scan_spm(load_scan: LoadScans) -> None:
+def test_load_scan_spm(load_scan_spm: LoadScans) -> None:
```
ns-rse, line 321: If we're switching to `minicricle_small` for the vast majority of tests do you think we should switch it for the `load_scan_spm` here too?

Or is it not practical because `minicircle_small` is of type `.topostats` and making the equivalent `.spm` is not possible/a lot of hassle?

tests/test_io.py
```diff
@@ -428,7 +428,7 @@ def test_gwy_read_component(load_scan_dummy: LoadScans) -> None:
 @pytest.mark.parametrize(
     "load_scan_object, length, image_shape, image_sum, filename, pixel_to_nm_scaling",
     [
-        ("load_scan", 1, (1024, 1024), 30695369.188316286, "minicircle", 0.4940029296875),
+        ("load_scan_spm", 1, (1024, 1024), 30695369.188316286, "minicircle", 0.4940029296875),
```
ns-rse, line 431: Could we add in `load_scan_data` as another set of parameters for this test to check the `.topostats`?

topostats/run_topostats.py
```diff
@@ -170,7 +170,10 @@ def run_topostats(args=None):
         else:
             summary_yaml = pkg_resources.open_text(__package__, "summary_config.yaml")
             summary_config = yaml.safe_load(summary_yaml.read())
-        summary_config = update_config(summary_config, args)
+
+        # Do not pass command line arguments to toposum as they clash with process command line arguments
```
ns-rse, line 174: Not sure if this is a wider issue but have noted this in #634 just in case

ns-rse (approved):
Brilliant work sorting this out @SylviaWhittle thank you so much, shortens the development cycle considerably which makes our lives much easier.

All the tests ran fine on my system with similar timings which is a vast improvement.

You'll see I've made a small PR (#635) to merge into this branch which should help reduce friction when setting up a new development environment. These were very minor and unrelated, technically I should probably have made a separate issue/branch/pr but was going through the process of checking things so made the changes. Hope you don't mind.

tests/test_io.py
SylviaWhittle: I tried my best to construct a `.spm` from `minicircle_small`, but Gwyddion does now allow saving to `.spm`, so it's effectively a read-only file. 😢

topostats/run_topostats.py
SylviaWhittle: Thank you 😄

ns-rse:
Closes #508 

(Not sure if that works from comments but lets find out!).

tests/test_io.py
SylviaWhittle: 👍 
I've added `load_scan_gwy` and `load_scan_topostats` (I forgot to add `load_scan_gwy` from my gwy file format PR 😨 )

ns-rse: Good catch on the `load_scan_gwy`, wasn't even thinking about that.

ns-rse (approved):
Cool, thanks for adding `load_scan_topostats` and good spot on `load_scan_gwy`. All looks ready to merge now.

PR #653, opened 2023-08-07
Improve linear splining

### LINKED ISSUE #607: Splining Linear Molecules
@Jean-Du has done some work on the [JeanDu/Spline-linear-molecules](https://github.com/AFM-SPM/TopoStats/tree/JeanDu/Spline-linear-molecules) for splining of linear molecules which currently isn't in the `main` branch.

This needs incorporating along with tests.

SylviaWhittle:
Closes #607 

This PR aims to improve the splining of linear molecules. Previously we just skipped splining for linear molecules, however @Jean-Du got it working through use of the smoothing and periodicity parameters to scipy's splprep (splining) function.

I wanted to make `get_splined_traces()` into a function rather than a method, since it doesn't really need the `self` keyword in it and could just as easily take all its arguments as parameters (better for documentation too!) so I've left it in a functional style, but still being a method in case we want to switch it later on.

I've also added unit tests for `get_splined_traces()` 😄 

I've had to update a lot of tests for this change. I have *NOT* sanity checked all of them, but I have plotted the effects on the splined traces of `minicircle.spm`:

Previous splines:
image

Updated splines:
image

Thanks to @Jean-Du for the code here, I just copied it across, applied pylint adherence and updated / created tests.


topostats/tracing/dnatracing.py
```diff
@@ -744,20 +714,31 @@ def trace_image(
...
+        image_spline_trace = trace_mask(grain_anchors, rounded_splined_traces, image.shape, pad_width)
     except ValueError as error:
         LOGGER.error("No grains found in any images, consider adjusting your thresholds.")
         LOGGER.error(error)
     return {
         "statistics": results,
         "ordered_traces": ordered_traces,
+        "splined_traces": splined_traces,
         "cropped_images": cropped_images,
         "image_trace": image_trace,
```
Jean-Du, line 740: Hi, I had a quick look at the code, and I noticed that you removed line 428 in processing.py, `image_trace = tracing_results["image_trace"]`, but you kept the `image_trace` here. Is this intentional, and is this `image_trace` supposed to be different from both the ordered and the splined trace?

Jean-Du:
Closes #607

topostats/tracing/dnatracing.py
SylviaWhittle: Yeah this is intentional as I foresee a need to have `image_trace` later on for image output, or for development purposes but we don't currently need to plot it. It is also useful for people using notebooks to have access to.

tests/tracing/test_dnatracing_methods.py
```diff
@@ -387,3 +388,194 @@ def test_crop_array(bounding_box: tuple, target: np.array, pad_width: int) -> No
...
+                    [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
+                    [0, 1, 0, 0, 0, 0, 0, 0, 1, 0],
+                    [0, 0, 1, 0, 1, 0, 1, 0, 1, 0],
+                    [0, 0, 0, 1, 0, 1, 0, 1, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                ]
+            ),
+        ),
+        (
```
ns-rse, line 470: This is the key skeleton for testing the linear splining I feel, but here the target is the same as the origin. Not a problem in and of itself but I'm wondering if we can devise a test case where the splining alters (i.e. improves) the skeleton trace as that is after all the ultimate aim of using splining so we should be testing this.

ns-rse (commented):
Some probably not very well formed thoughts.

I've made a note to ensure I rebase for the branch I'm working on the pruning on.

tests/tracing/test_dnatracing_methods.py
SylviaWhittle: I've added a linear spline that is changed by the splining function as you suggested, however I have been unable to construct a circular trace that is affected by the splining process.

MaxGamill-Sheffield:
Sorry if this has been waiting on me, I've been testing to see if we see unwanted oscillations in the linear splines like we do in the nodeStats branch, i.e. the following which give incorrect contour lengths:
20230601_Nicked1D_6ng_MgNi_eph 0_00009_above_splines

In the nodestats branch it may be due to the height-biased skeletonisation differences. I'll ignore this issue for now and I've not seen it occur for this branch but just a note for the future the solution may be to just use the ordered traces as these will already be height-aligned.

topostats/tracing/dnatracing.py
```diff
@@ -314,101 +315,65 @@ def get_splined_traces(self):
...
+        # Fitted traces are Nx2 numpy arrays of coordinates
+        # All self references are here for easy turning into static method if wanted, also type hints and short documentation
+        fitted_trace = self.fitted_trace  # : np.ndarray - boolean 2d numpy array of fitted traces to spline
+        step_size_m = self.step_size_m  # : float - the step size for the splines to skip pixels in the fitted trace
+        pixel_to_nm_scaling = self.pixel_to_nm_scaling  # : float - pixel to nanometre scaling factor for the image
+        mol_is_circular = self.mol_is_circular  # : mol_is_circular - whether or not the molecule is classed as circular
+        n_grain = self.n_grain  # : int - the grain index (for logging purposes)
+
+        # Calculate the step size in pixels
+        step_size_px = int(step_size_m / pixel_to_nm_scaling)
```
MaxGamill-Sheffield, line 327: Apologies I just noticed the actual splining func takes those instead. But what happens if in a scan our `px_2_nm_scaling` > 2*`step_size_nm` = 14nm 
-> will the int force a 0 and thus not spline any larger images regardless of what is in them?

topostats/tracing/dnatracing.py
```diff
@@ -314,101 +315,65 @@ def get_splined_traces(self):
...
+
+        # Splines will be totalled and then divived by number of splines to calculate the average spline
+        spline_sum = None
+        nbr = len(fitted_trace)
+
+        fitted_trace_length = fitted_trace.shape[0]
+
+        if fitted_trace_length < 4:
```
MaxGamill-Sheffield, line 335: Why 4? Wouldn't this depend on the step size in pixels?:
```suggestion
        if fitted_trace_length < step_size_px:
```

MaxGamill-Sheffield (changes requested):
Code looks good and clean. Thanks for making this change so quick and sorry for the lack of review on this until now.

One thing I want to ask is why some of the circular molecule lengths have changed? My thought is that it shouldn't be touched as only the params (and therefore splining) of the linear mols have changed? The amounts are small enough that it shouldn't be too much of an issue - is it the step_size?.

I also am curious about the shepherd hooks you have for the minicircles image and why despite they don't terminate on or nearer the trace - is it because of the spline averaging where the last coordinate is only obtained for 1 of 4 splines and as such has less weighting? Or am I overcomplicating this hahaha

topostats/tracing/dnatracing.py
```diff
@@ -314,101 +315,69 @@ def get_splined_traces(self):
...
+        # by averaging the splines. Think of this like weaving a lot of splines together along the course of
+        # the trace. Example spline coordinate indexes: [1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 3, 4], where spline
+        # 1 takes every 4th coordinate, starting at position 0, then spline 2 takes every 4th coordinate
+        # starting at position 1, etc...
+        for i in range(step_size_px):
+            x_sampled = np.array([fitted_trace[:, 0][j] for j in range(i, fitted_trace_length, step_size_px)])
+            y_sampled = np.array([fitted_trace[:, 1][j] for j in range(i, fitted_trace_length, step_size_px)])
+
+            # Use scipy's B-spline functions
+            tck, _ = interp.splprep([x_sampled, y_sampled], s=smoothness, per=periodicity, quiet=1, k=3)
```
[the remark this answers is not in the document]
SylviaWhittle: Added as values that can be given when initialising the class. 👍

topostats/tracing/dnatracing.py
SylviaWhittle: Yes, `int` would cause this to be zero for images that have a low `step_size_m` and a high number of nanometres per pixel (low resolution images). 

I could be wrong, but I think for the default case of 7nm for `step_size_m`, the resolution would need to be worse than 7nm per pixel before this would be calculated as zero.

Regardless, I've changed this so that `step_size_px` is always at least 1 to fix this potentiality.

topostats/tracing/dnatracing.py
SylviaWhittle: This is the minimum number of points it's okay to ask it to spline for the hard-coded degree polynomial of 3. Since I've now made the polynomial degree a configurable parameter, I've updated this accordingly. 👍

SylviaWhittle:
Just a note, mainly for future me: The linear molecules look different because previously we just returned the non-splined, fitted traces since we didn't spline the linear traces. We now spline them so they are different.

SylviaWhittle:
@MaxGamill-Sheffield the difference in circular molecule stats is due to a suggested change by @Jean-Du to set the smoothing of circular molecules to `2.0` (previously `0.0`). I have reverted it back to `0.0` now just for simplicity. We can always change it later.

SylviaWhittle:
I discovered and fixed a bug that caused splining to fail in some cases where there are duplicate, consecutive points in the sampled points arrays. Added a function to remove these duplicates and added tests for it. This results in more grains succeeding.

topostats/tracing/dnatracing.py
```diff
@@ -93,6 +93,7 @@ def __init__(
         self.number_of_rows = self.image.shape[0]
         self.number_of_columns = self.image.shape[1]
         self.sigma = 0.7 / (self.pixel_to_nm_scaling * 1e9)
+        self.step_size_m = 7e-9
```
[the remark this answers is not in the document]
SylviaWhittle: Made it configurable 👍

tests/tracing/test_dnatracing_methods.py
SylviaWhittle: They change now 😄

SylviaWhittle:
Oh and @ns-rse I added a contingency (plus test) to your `trace_mask` function, where it will skip grains that have pixels outside the bounds of the image. This happens to some splined traces and so to utilise your function I needed to add a contingency to it. (Tested of course).

SylviaWhittle:
> Sorry if this has been waiting on me, I've been testing to see if we see unwanted oscillations in the linear splines like we do in the nodeStats branch, i.e. the following which give incorrect contour lengths: 20230601_Nicked1D_6ng_MgNi_eph 0_00009_above_splines
> 
> In the nodestats branch it may be due to the height-biased skeletonisation differences. I'll ignore this issue for now and I've not seen it occur for this branch but just a note for the future the solution may be to just use the ordered traces as these will already be height-aligned.

I don't think this has been fixed, as the actual splining functionality has not changed. This could be a new issue if it needs addressing.

ns-rse:
> @MaxGamill-Sheffield the difference in circular molecule stats is due to a suggested change by @Jean-Du to set the smoothing of circular molecules to `2.0` (previously `0.0`). I have reverted it back to `0.0` now just for simplicity. We can always change it later.

I've not digged in to check but if this is a parameter it would make sense for it to be configurable, with a sensible default.

tests/tracing/test_dnatracing_multigrain.py
```diff
@@ -193,6 +193,28 @@ def test_image_trace_unequal_arrays() -> None:
...
     [
+        # Ensure that grains whose traces are outside of the image bounds are skipped
+        (
+            [
+                [0, 0],
+                [7, 7],
+            ],
+            [
+                np.asarray([[1, 1], [1, 2], [1, 3]]),  # Grain 0's points plus anchor 0 is inside image bounds
+                np.asarray([[1, 1], [1, 2], [1, 3]]),  # Grain 1's points plus anchor 1 are outside image bounds
```
ns-rse, line 204: Nice, good to have tests that meet exceptions/are outside of bounds. :+1:

tests/tracing/test_dnatracing_multigrain.py
```diff
@@ -301,7 +323,7 @@ def test_trace_mask(
                 {
                     "molecule_number": [0, 1],
                     "image": ["multigrain_topostats", "multigrain_topostats"],
-                    "contour_length": [1.2382864476914832e-07, 7.617314045334366e-08],
+                    "contour_length": [5.684734982126663e-08, 7.574136072208753e-08],
```
ns-rse, line 326: Grain lengths look like they are roughly/broadly halving but the image examples don't look like that for most grains. Just wanted to check that this reduction in length is correct.

tests/tracing/test_dnatracing_methods.py
```diff
@@ -386,3 +387,392 @@ def test_crop_array(bounding_box: tuple, target: np.array, pad_width: int) -> No
...
+                    [0, 0, 0, 1, 0, 1, 0, 0],
+                    [0, 0, 0, 0, 1, 1, 0, 0],
+                    [0, 0, 0, 0, 1, 0, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0, 0],
+                ]
+            ),
+        ),
+    ],
+)
+def test_get_splined_traces(
```
ns-rse, line 700: This test is doing a lot of work. Some of the steps could I think be moved out and a separate fixture for this specific test.

For example the instantiated `dnatrace` fixture could be setup with....

```
@pytest.fixture()
def dnatrace_spline(dnatrace: dnaTrace) -> dnaTrace:
   """dnaTrace object for splining tests."""
   dnatrace.pixel_to_nm_scaling = 1.0
   dnatrace.n_grain = 1
   return dnatrace
```

...and then use `dnatrace_spline`.

That said I don't think much else could be broken out as the reordering and other steps need performing conditional on the skeleton that is being passed in.

Has made me realise that the refactoring I did is incomplete as there aren't tests for things like `reordertrace.linearTrace()` or `reordertrace.circularTrace()`, captured in #706.

ns-rse (approved):
Sorry I've taken so long to get round to reviewing this, its a big PR.

Some minor comments but nothing blocking this from being merged.

tests/tracing/test_dnatracing_multigrain.py
SylviaWhittle: Thank you for being so vigilant! I believe this is correct. I saved the splined trace for that molecule in main and my branch, and this is the difference in the splines:

image

So the spline is much shorter now since it is not jumping about all over the place, artificially increasing the length.

Does this seem okay to you?

In my investigations, I have found that contour length is possibly being calculated badly, and providing systematically incorrect results, but that's for more investigation and a separate issue focusing on measure_contour_length.

tests/test_processing.py
```diff
@@ -497,4 +497,4 @@ def test_run_dnatracing(process_scan_config: dict, tmp_path: Path) -> None:
 
     assert isinstance(dnatracing_df, pd.DataFrame)
     assert dnatracing_df.shape[0] == 13
-    assert len(dnatracing_df.columns) == 23
+    assert len(dnatracing_df.columns) == 26
```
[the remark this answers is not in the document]
SylviaWhittle: So previously to this PR, during running of `test_run_dnatracing()`, the lines inside `get_splined_traces()` that have the `.pop()` methods, are called.


image


This would cause an AttributeError and the test would fail, except it doesn't since we have a broad `except` around dna tracing in `processing.py`.

image

Dna tracing as a result returns our `create_empty_dataframe()` data frame which doesn't contain the 3 columns for dna tracing data, namely `contour_length`, `circular` and `end_to_end_distance`.

This means that it doesn't crash, and the test runs successfully.
I stupidly didn't check the columns when writing this test, and assumed that since it didn't crash, it worked as intended. (many apologies for this).

So the test currently asserts that the returned data frame has the columns without the dna tracing columns.

I verified this by saving the data frames during the test, here we see that the old one was missing the columns:

Screenshots for ease of seeing the difference:

Old:
image

New:
image

Copy & Paste of the columns direct from the csv files:


New:
molecule_number,centre_x,centre_y,radius_min,radius_max,radius_mean,radius_median,height_min,height_max,height_median,height_mean,volume,area,area_cartesian_bbox,smallest_bounding_width,smallest_bounding_length,smallest_bounding_area,aspect_ratio,threshold,max_feret,min_feret,image,contour_length,circular,end_to_end_distance,basename


Old:
molecule_number,centre_x,centre_y,radius_min,radius_max,radius_mean,radius_median,height_min,height_max,height_median,height_mean,volume,area,area_cartesian_bbox,smallest_bounding_width,smallest_bounding_length,smallest_bounding_area,aspect_ratio,threshold,max_feret,min_feret,image,basename

tests/tracing/test_dnatracing_methods.py
```diff
@@ -386,3 +387,392 @@ def test_crop_array(bounding_box: tuple, target: np.array, pad_width: int) -> No
...
+                [
+                    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                    [0, 0, 1, 1, 1, 0, 0, 0, 0, 0],
+                    [0, 1, 0, 0, 0, 1, 0, 0, 0, 0],
+                    [0, 1, 0, 0, 0, 0, 1, 0, 0, 0],
+                    [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
+                    [0, 0, 1, 0, 0, 0, 0, 0, 1, 0],
+                    [0, 0, 1, 0, 0, 0, 0, 0, 1, 0],
+                    [0, 0, 1, 0, 0, 0, 0, 0, 1, 0],
+                    [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
```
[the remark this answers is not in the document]
SylviaWhittle: Yeah it is weird, though not in my power to fix without digging into the code of `scipy.splprep` I don't think ☹️

tests/tracing/test_dnatracing_methods.py
SylviaWhittle: I've added the fixture, but I likewise cannot think of a way to break out the logic for this rather messy test ☹️

tests/test_processing.py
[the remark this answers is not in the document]
ns-rse: :+1: Thanks for taking the time to explain all that, I completely agree the broad `try: .... except:` causes more pain that it solves, but there was urgency to get something running that didn't continually fail before refactoring of `dnatracing` was undertaken.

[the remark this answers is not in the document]
SylviaWhittle: Oh I don't deny that using the broad exception should have been used, we needed to get things done 😄

ns-rse (approved):
Good to go :white_check_mark: :+1:

PR #654, opened 2023-08-07
Fix broken reference to 'image', now 'image_original'

SylviaWhittle:
This PR is a hotfix for the notebooks, as a previous PR changed the structure of how we store the loaded data from AFM files. 
The raw image used to be called `'image'`, but not it's `'image_original'` which I forgot to update in the notebooks. I will make tests for the notebooks soon.


MaxGamill-Sheffield (approved):
Seems like a simple enough change, thanks for this Sylvia!

PR #658, opened 2023-08-22
Fix plotting in pixels for topostats files

### LINKED ISSUE #657: .topostats file plotting in pixels not nanometres
IMG_1496

## Describe the bug

Pixel units on images x and y axes instead of nanometre units

## Include the configuration file

Any config

## To Reproduce

Run grains and dnatracing (not filters) on a .TopoStats file.
Notice the images output are in pixels not nanometres on the x and y.

## TopoStats version

Topo Version : 2.1.1

### Operating System

Please let us know what operating system you are using, if you have used more than one then tick all boxes.

Windows
MacOS

### Python Version

All versions

SylviaWhittle:
Closes #657 

This PR fixes the issue where TopoStats would output images scaled in pixels if `Filters` had not been ran.
The issue happened because the code that adds the `pixel_to_nanometre_scaling` factor into the `plotting_dictionary` was inside the `run_filters` function. I have created a function for this code, added a test for it, and moved it out front into `process_scan` as it should always be run.

See here `minicircles.spm` successfully plotted in nanometres without `Filters` being ran.
image

### DIFF
--- a/topostats/plottingfuncs.py
+++ b/topostats/plottingfuncs.py
@@ -25,6 +25,31 @@ LOGGER = logging.getLogger(LOGGER_NAME)
 # pylint: disable=dangerous-default-value
 
 
+def add_pixel_to_nm_to_plotting_config(plotting_config: dict, pixel_to_nm_scaling: float) -> dict:
+    """Function to ensure that the plotting config has the pixel to nanometre scaling factor accessible
+    for plotting. This ensures that the plot scaling are in nanometres and not pixels.
+
+    Parameters
+    ----------
+    plotting_config: dict
+        TopoStats plotting configuration dictionary
+    pixel_to_nm_scaling: float
+        Pixel to nanometre scaling factor for the image.
+
+    Returns
+    -------
+    plotting_config: dict
+        Updated plotting config with the pixel to nanometre scaling factor
+        applied to all the image configurations.
+    """
+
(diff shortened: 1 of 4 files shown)


topostats/plottingfuncs.py
```diff
@@ -25,6 +25,31 @@
 # pylint: disable=dangerous-default-value
 
 
+def add_pixel_to_nm_to_plotting_config(plotting_config: dict, pixel_to_nm_scaling: float) -> dict:
+    """Function to ensure that the plotting config has the pixel to nanometre scaling factor accessible
```
ns-rse, line 29: I'm probably being overly pedantic but I don't think docstrings need prefixing with "_Function to_".

```suggestion
    """Ensure that the plotting config has the pixel to nanometre scaling factor accessible
```

topostats/plottingfuncs.py
```diff
@@ -25,6 +25,31 @@
...
+    for plotting. This ensures that the plot scaling are in nanometres and not pixels.
```
ns-rse, line 30:
```suggestion
    for plotting so that plots are in nanometres and not pixels.
```

ns-rse (approved):
Simple and effective, thanks for fixing this @SylviaWhittle 

Very minor, pedantic suggestions on the docstring but otherwise ready to merge.

ns-rse (approved):
Sorry for the delay in looking at this, its good to go @SylviaWhittle :+1:

PR #662, opened 2023-09-12
Moves DPI configuration from default_config to plotting_dictionary

### LINKED ISSUE #645: Move DPI setting to plotting dictionary
We only need high resolution images for traced plots and having it as a global setting means that plotting of scans slows down too much.

By moving the configuration of DPI to the plotting dictionary we will have finer grained control over the resolution of images.

ns-rse:
Closes #645

Moves specification of a global DPI for all images from `topostats/default_config.yaml` to per-plot in
`topostats/plotting_config.yaml` with default of `100` for most values but `800` for existing plots of traces.

Updates validation rules and documentation and I've taken the opportunity to add missing Description to two rows in the
documentation covering the `summary_config.yaml`.


SylviaWhittle (approved):
Simple and works, tested locally with different DPIs. 👍
Thanks @ns-rse !

PR #665, opened 2023-09-14
Changed dict crop_size -1 names to match processing

MaxGamill-Sheffield:
Noticed that no grainstats were being made when cropped_size = -1

Was due to incorrect dictionary keys being called


ns-rse (approved):
Good spot, thanks for the fix @MaxGamill-Sheffield :+1:

Issue #487, opened 2023-03-01
Add upper threshold option for grain height


SylviaWhittle:
Currently, grains are detected as being more than `threshold` above or below the background. Sometimes however, we want to be able to prevent the masking of data that lie too far from the background. For example, 'blobs' are often a lot higher than DNA, and we don't want them detected as parts of grains. 

Here is an attempt at drawing (side on) what this would look like for DNA data.

                                       XX            blob height
-----------------------XX------  "upper" threshold for data above the background
      oooo            oo        XXX          DNA height
--ooooooo----ooo---XXXX----- "lower" threshold for data above the background
======================== Background level

  ~~~~~~~      ~~~                         Data Mask

derollins:
Any update on this? I would like to implement this in my ["haribo rings"](https://github.com/AFM-SPM/TopoStats/tree/derollins/hariborings) project.

SylviaWhittle:
I'll do that now 👍

derollins:
To clarify, I'd like to just unmask parts of a grain that are above a threshold, not get rid of the whole grain.

SylviaWhittle:
@derollins Yep!

PR #666, opened 2023-09-14
Add upper limit to thresholding

SylviaWhittle:
This PR adds the ability to set both a minimum and maximum when thresholding. See #487.

From a user-perspective this changes the config files from having a single value in the thresholding sections, to having two values like how `absolute_area_thresholds` have:

```yaml
  threshold_std_dev:
    below: [10.0, null] # Minimum and maximum thresholds for data below the image background
    above: [1.0, null] # Minimum and maximum thresholds for data above the image background
  threshold_absolute:
    below: [-1.0, null] # Minimum and maximum thresholds for data below the image background
    above: [1.0, null] # Minimum and maximum thresholds for data above the image background
```

Note that:
- Otsu method still takes a single fudge factor, as the nature of this threshold is to perform binary segmentation.
- `null` is an accepted value when you don't want to set an minimum and or maximum limit to the thresholds.

See here where `minicircle.spm` has been processed with `threshold_std_dev` `upper` as `[1.0, 3.5]`. So this is masking points that are between `1.0` and `3.5` standard deviations from the mean of the image.
image

Technical notes:
Internally the way thresholding is done has been overhauled.
- Tests have been added for previously untested 😱  `utils.py`'s `_get_mask` and `get_mask` (now called `get_and_combine_directional_masks` for specificity)
- The `"above"` and `"below"` keys in threshold dictionaries are now considered to be always existent / expected.
- Skipping directions is now handled by detecting if the threshold in `"above"` or `"below"` is `None` or not. This makes it a little cleaner I think.
- `utils.py`'s `get_threshold` is monolithic and huge, but I tried refactoring it in clever and elegant ways, and the solutions I came up with required some rather less clear code, so I have purposely left the function verbose as to be more accessible / readable. Happy to have pushback on this though!
- As you can see, no image related tests have changed, so this does not affect the results of processing unintentionally.
- I had to get rid of the lovely lambda statements in `validation.py` as I do not know how to get lambda statements to be okay with having a `None` type too (required to allow `null` values in the config).
- The tests are rather verbose now, as to compare dictionaries with `np.Infinity` in them, we **must** unpack it and use the `np.isposinf()` or `np.isneginf()` methods where infinities are encountered. **This is because `np.Infinity != np.Infinity`**.


SylviaWhittle:
I'll update the notebooks & documentation now, I almost forgot!

SylviaWhittle:
I'd appreciate a preliminary review so that I can update the documentation after we have settled on the right course of action if that's okay 😅 

@derollins I'd really like your input too from a user-perspective if you have time 🙏

topostats/validation.py
```diff
@@ -73,24 +73,52 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
             ),
             "otsu_threshold_multiplier": float,
             "threshold_std_dev": {
-                "below": lambda n: n > 0,
-                "above": lambda n: n > 0,
+                "below": [
```
ns-rse, line 76: We want to avoid having negative numbers passed  and I've not tried this but could you perhaps put the lambda within the `Or()`?

```
Or(
    None,
    lambda n: n > 0,
    error=("Whoops!")
)
```

topostats/utils.py
```diff
@@ -110,7 +110,7 @@ def update_plotting_config(plotting_config: dict) -> dict:
     return plotting_config
 
 
-def _get_mask(image: np.ndarray, thresh: float, threshold_direction: str, img_name: str = None) -> np.ndarray:
+def _get_mask(image: np.ndarray, thresholds: list, threshold_direction: str, img_name: str = None) -> np.ndarray:
```
ns-rse, line 113: `thresholds` is a `dict` here as the contents are referenced by keys rather than position.

```suggestion
def _get_mask(image: np.ndarray, thresholds: dict, threshold_direction: str, img_name: str = None) -> np.ndarray:
```

topostats/utils.py
```diff
@@ -199,20 +209,68 @@ def get_thresholds(
...
     thresholds = defaultdict()
     if threshold_method == "otsu":
-        thresholds["above"] = threshold(image, method="otsu", otsu_threshold_multiplier=otsu_threshold_multiplier)
+        thresholds["above"] = {
+            "minimum": threshold(image, method="otsu", otsu_threshold_multiplier=otsu_threshold_multiplier),
+            "maximum": np.Infinity,
+        }
+        thresholds["below"] = None
     elif threshold_method == "std_dev":
         try:
```
ns-rse, line 218: This a long block of nested `if: ... else:` statements that are easy to get lost in.
Perhaps it could be broken down into a couple of methods that are easier to reason about. This would help address the linting errors that have been disabled. Perhaps `_get_mean_min_max()`.

Similarly break the code for `absolute` out to its own function too `_get_absolute_min_max()`.

derollins:
Tried this, seems to work although there is a ring around the objects I don't want to mask... which is inevitable and not the fault of the code.
20230413_Cas9_RNA_complex_DNA_MCON_4ng_NiCl2_3mM_HEPES_20mM 0_00027_above_masked

topostats/validation.py
SylviaWhittle: Sadly that causes this error for all of them:
```
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/schema.py", line 405, in validate
    nvalue = Schema(svalue, error=e, ignore_extra_keys=i).validate(value, **kwargs)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/new/lib/python3.10/site-packages/schema.py", line 409, in validate
    raise SchemaError([message] + x.autos, [e.format(data) if e else None] + x.errors)
schema.SchemaError: Key 'threshold_std_dev' error:
Key 'below' error:
<lambda>([10.0, None]) raised TypeError("'>' not supported between instances of 'list' and 'int'")
```

topostats/utils.py
SylviaWhittle: Done 👍 sorry for the delay

SylviaWhittle:
Do we think this is an actually useful feature to add, @derollins? It adds complexity and so I'm hesitant to add it unless we are sure it is necessary or needed.

ns-rse:
I'm all for following the KISS principle.

SylviaWhittle:
Tests updated and coverage has been made comprehensive thanks to CodeCov's ability to detect per-line coverage. 🔧

SylviaWhittle:
I think I have addressed all but one of your points @ns-rse. The schema doesn't seem to like using lambda functions in Or sadly 😢

SylviaWhittle:
Note to self: make the config take both single values (counted as the lower threshold) and two values (lower and upper), parsed when reading the config. This is to reduce complexity of the config file and because two thresholds will be an extremely niche feature.

SylviaWhittle:
Closing as this still has not been requested by users and it would be easier to re-make rather than update this old PR. Can easily be ressurected later if needed.

PR #667, opened 2023-09-16
Lower default DPI of individual molecule plots

SylviaWhittle:
Changes the DPI for individual molecule plots to `100` as the previous setting of `800` caused RAM issues and slow plotting. Thanks to @MaxGamill-Sheffield for pointing this out and @ns-rse for enabling easy configuring of DPI for each type of plot!

I have also added the DPI into the logging statement for plotting as this may be helpful for debugging.

### DIFF
--- a/topostats/plottingfuncs.py
+++ b/topostats/plottingfuncs.py
@@ -224,7 +224,8 @@ class Images:
                     else:
                         self.save_array_figure()
         LOGGER.info(
-            f"[{self.filename}] : Image saved to : {str(self.output_dir / self.filename)}" f".{self.save_format}"
+            f"[{self.filename}] : Image saved to : {str(self.output_dir / self.filename)}.{self.save_format}\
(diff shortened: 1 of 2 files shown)


topostats/plottingfuncs.py
```diff
@@ -224,7 +224,8 @@ def plot_and_save(self):
                     else:
                         self.save_array_figure()
         LOGGER.info(
-            f"[{self.filename}] : Image saved to : {str(self.output_dir / self.filename)}" f".{self.save_format}"
+            f"[{self.filename}] : Image saved to : {str(self.output_dir / self.filename)}.{self.save_format}\
+ | DPI: {self.dpi}"
```
ns-rse, line 228: Nice idea to add details of DPI used to logging :+1:

ns-rse (approved):
Simple and effective, and my apologies for inflating the DPI in my recent PR which necessitated this correction.

PR #675, opened 2023-09-28
Add .asd file support

### LINKED ISSUE #304: Support  ASD file format

### LINKED ISSUE #647: Fork libasd and release to PyPI
The builds for [libasd](https://github.com/ToruNiina/libasd) are not correct on PyPI and despite contacting the author multiple times we haven't seen this resolved.

We want to support `.asd` format and so have decided that we will fork the repository, rename and get the releases working so that libasd is installable across architectures and we can support `.asd`.

SylviaWhittle:
Closes #647 
Closes #304 

This PR adds support for loading `.asd` files into TopoStats, processing all the frames automatically and saving each frame as an individual image output.

This is not going to be the final workflow for `.asd` files, however I want to split this task up into smaller PRs to get feedback from users on how it should be improved going forward.

The actual loading of the `.asd` file is done using my [topofileformats](https://github.com/AFM-SPM/topofileformats) package that I wrote to decode and load `.asd` files into numpy arrays.

## Technical details
- I have added a test for loading `.asd` files.
- I had to change how files are added to the image dictionary in `io.py` as now each file might have more than one image. I hacked this in by checking the file extension and if it's `.asd` then iteratively add all the frames. Far from tidy, would welcome suggestions on better methods.


SylviaWhittle:
This implementation uses **a lot** of RAM. I ran a 57mb .asd file and it used about 2GB of RAM with grains, grain stats and dnatracing disabled. It would be a lot more with those enabled.

Note to self: is there a way to free memory in between frames being processed?

SylviaWhittle:
The RAM issue seems to be okay actually. While it does use a lot of RAM, it does not increase when running with image set `all` and maintains its stability at 2GB of RAM. This is reasonable given the weight of the files being asked to process.

tests/test_io.py
```diff
@@ -93,7 +93,12 @@ def test_write_yaml(tmp_path: Path) -> None:
 
 def test_path_to_str(tmp_path: Path) -> None:
     """Test that Path objects are converted to strings."""
-    CONFIG_PATH = {"this": "is", "a": "test", "with": tmp_path, "and": {"nested": tmp_path / "nested"}}
+    CONFIG_PATH = {
```
ns-rse, line 96: Curious what linter (`black` or `ruff>=0.1.4`-ish) have decided to wrap these lines as they and others in this file are all <120. :thinking:

tests/test_io.py
```diff
@@ -441,6 +480,7 @@ def test_load_scan_get_data(
     scan = request.getfixturevalue(load_scan_object)
     scan.get_data()
     assert len(scan.img_dict) == length
+    print(scan.img_dict)
```
ns-rse, line 483: Errant debug `print()`.

Surprised `ruff` doesn't pick this up (see [print (T201)](https://docs.astral.sh/ruff/rules/print/)) either locally 
when running the `pre-commit` hook or here on when `pre-commit.ci` runs.

I'll look into that as we've had to explicitly disable this in `tests/test_processing.py` where `pytest.regtest` requires a `print()` statement to write the CSV files for the regression tests.

tests/test_io.py
```diff
@@ -465,7 +505,7 @@ def test_load_scan_get_data_check_image_size_and_add_to_dict(
     load_scan_spm.filename = "minicircle"
     load_scan_spm.img_path = tmp_path
     load_scan_spm.image = np.ndarray((x, y))
-    load_scan_spm._check_image_size_and_add_to_dict()
+    load_scan_spm._check_image_size_and_add_to_dict(image=load_scan_spm.image, filename=load_scan_spm.filename)
```
ns-rse, line 508: Good spot, presumably this test wasn't doing anything or failed in some manner?

topostats/io.py
```diff
@@ -889,44 +914,55 @@ def get_data(self) -> None:
...
             else:
                 raise ValueError(
                     f"File type {suffix} not yet supported. Please make an issue at \
                 https://github.com/AFM-SPM/TopoStats/issues, or email topostats@sheffield.ac.uk to request support for \
                 this file type."
                 )
 
-    def _check_image_size_and_add_to_dict(self) -> None:
+    def _check_image_size_and_add_to_dict(self, image: np.ndarray, filename: str) -> None:
         """Check the image is above a minimum size in both dimensions.
```
ns-rse, line 930: I think its useful to explain the action taken in light of the check.

```suggestion
        """Check the image is above a minimum size in both dimensions. If not the image is not processed further.
```

ns-rse (changes requested):
Looks good @SylviaWhittle I like that you've moved a lot of the leg-work to its own package in [`topofileformats`](https://github.com/AFM-SPM/topofileformats/) (hope you don't mind I've added myself as a collaborator on that project and started looking through the code).

Some comments in line.

tests/conftest.py
```diff
@@ -387,6 +387,12 @@ def load_scan_topostats() -> LoadScans:
     return LoadScans([RESOURCES / "file.topostats"], channel="dummy_channel")
 
 
+@pytest.fixture()
+def load_scan_asd() -> LoadScans:
+    """Instantiate a LoadScans object from a .asd file."""
+    return LoadScans([RESOURCES / "file.asd"], channel="TP")
```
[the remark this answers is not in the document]
ns-rse: 2023-12-19 - Only these three channels are known about.

ns-rse:
Three different standards for different ASD files, 0, 1 and 2, only have test files for 0 and 1 but not 2.

ns-rse:
@ns-rse to round out.

ns-rse (approved):
All comments addressed (most were just that and didn't require any changes and one was redundant I'd just not read the subsequent line).

Removed errant `print()` I'm unsure why `ruff` didn't pick this up (either locally or via `pre-commit.ci` as [T201](https://docs.astral.sh/ruff/rules/print/) should have complained about its presence and the file isn't explicitly excluded.

Otherwise good to go.

ns-rse (approved):
Good to go, can I approve having now contributed though? :thinking:

PR #678, opened 2023-10-04
Disable scars by default and fix KDE plotting bug

### LINKED ISSUE #677: Disable scar removal by default
After some discussion, I think it might be best to disable scar removal by default.

My reasoning is that scar removal can affect the quality of the data, so should only be enabled if entirely necessary. It also is extremely slow and makes up 70% of TopoStats' processing time.

Of course I am not suggesting removing the feature, only changing the default behaviour in `default_config.yaml` to not run it.

SylviaWhittle:
Closes #677 

This small PR does two things:
- Disables scars by default (See #677 for my reasoning). I have kept the tests running with scar removal on, as the way I see it, it's more comprehensive to test *with* scar removal than without.
- Fixes an issue that I found while checking the tests, that causes a `numpy.linalg.LinAlgError` where when plotting KDE plots, if all values are 0, then the standard deviation is zero and a `DivisionByZero` error causes the aforementioned error. I handle this by adding a `None` return option to the `sns_plot()` method and handle the `Optional[Tuple]` in the calling `run_toposum()` function.


ns-rse:
I think we've encountered a problem with a new plotting version here, I'll investigate that today.

With regards to tests being on by default, broadly yes that is probably what we want. I'll have to check that the results with scar removal off are also tested as tests are meant to check the impact of state/options, so perhaps its worth adding a `test_scars_off()` to `tests/test_scars.py` to check this (separate issue though, keeps issues/PRs small and focused and easier to work on and review).

topostats/plotting.py
```diff
@@ -139,31 +139,56 @@ def _outfile(self, plot_suffix: str) -> str:
         """
         return f"{self.stat_to_sum}_{plot_suffix}"
 
-    def sns_plot(self) -> None:
+    def sns_plot(self) -> Optional[Tuple[plt.Figure, plt.Axes]]:
```
ns-rse, line 142: Nice, thanks :+1:

topostats/plotting.py
```diff
@@ -139,31 +139,56 @@ def _outfile(self, plot_suffix: str) -> str:
...
         """Plot the distribution of one or more statistics as either histogram, kernel density estimates or both. Uses
-        base Seaborn."""
+        base Seaborn.
+
+        Returns
```
ns-rse, line 146: :grin: thanks for adding this, I should have done it myself.

topostats/plotting.py
```diff
@@ -139,31 +139,56 @@ def _outfile(self, plot_suffix: str) -> str:
...
+        -------
+        Optional[Union[Tuple[plt.Figure, plt.Axes], None]]
+            Tuple of Matplotlib figure and axes if plotting is successful, None otherwise.
+        """
+
+        # Note: Plotting KDEs with Seaborn is not possible if all values are 0. This is because the KDE is calculated
+        # using a Gaussian kernel and if all values are 0, the standard deviation is 0 wich results in a
+        # ZeroDivisionError. To avoid this, we check if all values are 0 and if so, we skip the KDE plot.
```
ns-rse, line 154: Sorry for not reading this in detail earlier but if we observe a `ZeroDivsionError` then the natural/Pythonic thing to do would be to use `try: ... except:`

This needs to fit within the logic of whether `self.kde` is true or not.

And so...

```python
if self.hist and not self.kde:
    outfile = self._outfile("hist")
    sns.histplot(data=self.melted_data, x="value", bins=self.bins, stat=self.stat, hue=self.hue)
if self.kde and not self.hist:
    outfile = self._outfile("kde")
    try:
        sns.kde.plot(data=self.melted_data, x="value", hue=self.hue)
    except ZeroDivsionError:
        LOGGER.info(f"[plotting] All values for {self.label} are 0. KDE plots cannot be made as there is no variance, skipping."
        return None
if self.hist and self.kde:
    outfile = self._outfile("hist_kde")
    try:
        sns.histplot(
            data=self.melted_data,
            x="value",
            bins=self.bins,
            stat=self.stat,
            hue=self.hue,
            kde=True,
            kde_kws={"cut": self.cut},
        )
    except ZeroDivisionError:
       LOGGER.info(f"[plotting] All values for {self.label} are 0. KDE plots cannot be made as there is no variance, skipping."
       return None
```

...is one way of doing that. I realise its perhaps not _that_ efficient as its possible to determine if the variance is zero before trying to plot, but it capture instances where all values are the _same_ rather than just zero.

topostats/plotting.py
```diff
@@ -139,31 +139,56 @@ def _outfile(self, plot_suffix: str) -> str:
...
-            sns.histplot(
-                data=self.melted_data,
-                x="value",
-                bins=self.bins,
-                stat=self.stat,
-                hue=self.hue,
-                kde=True,
-                kde_kws={"cut": self.cut},
+        # Otherwise, KDE is requested, so check if all values are 0
+        elif (self.melted_data["value"] == 0).all():
```
ns-rse, line 166: This is an awkward situation but you would also get errors raised if all values were `!= 0` but all the same, because the variance (and in turn the standard deviation which is the square-root of the variance) would be zero.

So a better test if sticking with this structure rather than the suggested `try: ... except:` would be...

```suggestion
        elif self.melted_data["value"].std() == 0:
```

ns-rse (changes requested):
Sorry for not looking at this in greater detail earlier.

Hope the suggestions are useful, I think at the very least the test should be whether the standard deviation is zero rather than all values as its broader on the off chance all values are the same (unlikely but you never know!).

Whether to go with the Pythonic `try: ... except:` is more a matter of style.

SylviaWhittle:
🤦‍♀️ I should have updated it to check for the same value instead of all zeros, thank you. (I originally thought that the error was from only having zeros and forgot to update the check).

I've taken the suggested `try` `except` method as I think that's simpler from a reading perspective and the speed here is not important. I updated the note and comment since the error raised is actually a `np.linalg` error, since I think the division by zero is handled but then causes an error later on. A little fuzzy on the specifics but this works!

topostats/plotting.py
```diff
@@ -184,6 +187,13 @@ def sns_plot(self) -> Optional[Tuple[plt.Figure, plt.Axes]]:
                     kde=True,
                     kde_kws={"cut": self.cut},
                 )
+            except np.linalg.LinAlgError:
```
ns-rse, line 190: That is an obscure error I've not come across before!

SylviaWhittle:
I've just seen CodeCov's useful comments saying that there are now lines added that aren't tested. I can add tests for this if we think it's not overly detailed?

ns-rse (approved):
Thanks @SylviaWhittle for these changes. :+1: 

As to whether to add tests, you can either add them into this PR or make another short issue and add it separately, but its worth doing either way I think.

PR #685, opened 2023-10-04
Switch to igor2 for .ibw and unpeg NumPy

### LINKED ISSUE #682: Unpeg numpy==1.23.4 by switching to igor2
Currently we have pegged `numpy==1.23.4` in our requirements.

This is because in `numpy>=1.23` the `np.complex` which was an alias for the builtin `complex` is deprecated.

We encounter this error courtesy of the `igor` library.

```bash
❱ pip show numpy | head -n20
Name: numpy
Version: 1.26.0
Summary: Fundamental package for array computing in Python
Home-page: https://numpy.org
Author: Travis E. Oliphant et al.
Author-email: 
License: Copyright (c) 2005-2023, NumPy Developers.
❱ pytest -n 14
ImportError while loading conftest '/home/neil/work/projects/topostats/TopoStats/tests/conftest.py'.
tests/conftest.py:14: in <module>
    from topostats.io import read_yaml, LoadScans
topostats/io.py:15: in <module>
    from igor import binarywave
/home/neil/.virtualenvs/topostats/lib/python3.11/site-packages/igor/binarywave.py:110: in <module>
    1:_numpy.complex, # NT_CMPLX, makes number complex.
/home/neil/.virtualenvs/topostats/lib/python3.11/site-packages/numpy/__init__.py:324: in __getattr__
    raise AttributeError(__former_attrs__[attr])
E   AttributeError: module 'numpy' has no attribute 'complex'.
E   `np.complex` was a deprecated alias for the builtin `complex`. To avoid this error in existing code, use `complex` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.complex128` here.
E   The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
E       https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

A drop-in replacement that may solve this is to switch from [igor==0.3](https://pypi.org/project/igor/) which was last updated 2016-08-02 to [igor2==0.5.3](https://pypi.org/project/igor/) which was last updated 2023-07-21.

From [Igor2 README](https://github.com/AFM-analysis/igor2)...

> Igor2 is the continuation of the inactive igor project, forked from W. Trevor King and originally written by Paul Kienzle (see git history).

...and...

> This package is a direct replacement of igor. Your scripts should work without any issues if you replace:
> `import igor`
>
> with:
> 
> `import igor2 as igor`
>
> See the docstrings and unit tests for examples using the Python API.

ns-rse:
Closes #682

We had pegged `numpy==1.23.4` because the unmaintained [igor](https://pypi.org/project/igor/) used the now deprecated `np.complex`. By switching to the continuation project [igor2](https://pypi.org/project/igor2/) which still sees active maintenance we are able to remove the fixed numpy version.

I've also taken the liberty to remove `flake8` packages from the list of `dev` optional-dependencies as we no longer have `pre-commit` hooks in place and the task is done by `ruff` and added `ipython` as its a commonly used alternative to plain Python interface (I prefer it as it has additional/commands but always forget to install it in new VirtualEnvs I setup for testing).

## Why were tests failing

My initial thought on why tests were failing with my attempt to upgrade to `igor2` and remove `numpy==1.23.4` were wrong (I thought I'd branched off an old commit, but that wasn't the case, good to know I'm not going mad or doing silly things!).

It seems that for some reason under `python==3.8` `igor2==0.5.2` gets installed rather than `igor==0.5.3` as happens on other Python versions (3.9, 3.10, 3.11), we see the line....

```
INFO: pip is looking at multiple versions of igor2 to determine which version is compatible with other requirements. This could take a while.
Collecting igor2 (from topostats==0.1.dev1+geaa9029)
101
  Downloading igor2-0.5.2-py3-none-any.whl (31 kB)
```

_after_ `igor2-0.5.3` has been installed in the logs for Python 3.8 under Linux/OSX/Windows but not for other versions of Python across platforms they all install.

A bit of digging however reveals that if I try to force `igor==0.5.3` then `pip` can't resolve dependencies because....

```bash
The conflict is caused by:
    topostats[tests] 0.1.dev1+g3ef231c depends on numpy
    igor2 0.5.3 depends on numpy>=1.25.1
```

...and versions `numpy>=1.25.0` only supports python>=3.9 (see [release notes](https://numpy.org/doc/stable/release/1.25.0-notes.html)).

I've therefore set the minimum Python >= 3.9 for TopoStats and will with the update to support recent versions of Seaborn and once #678 and 681 are merged in make a new release to PyPI, but only after I add support for [Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html) which was released two days ago (see #686)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.9', '3.10', '3.11']
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,9 +19,9 @@ jobs:
         uses: actions/setup-python@v4.3.0
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Upgrade pip and install test dependencies
         run: |
-          pip install -e .
+          pip install --upgrade pip
           pip install -e .[tests]
       - name: Test with pytest
         run: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,9 +37,9 @@ keywords = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-  "igor",
+  "igor2",
   "matplotlib",
-  "numpy==1.23.4",
+  "numpy",
   "pandas",
   "pySPM",
   "pyyaml",
@@ -78,10 +78,8 @@ docs = [
   "sphinxcontrib-napoleon",
 ]
 dev = [
-  "Flake8-pyproject",
   "black",
-  "flake8",
-  "flake8-print",
+  "ipython",
   "pre-commit",
   "pylint",
   "pyupgrade",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ keywords = [
 ]
 requires-python = ">=3.8"
 dependencies = [
-  "igor2",
+  "igor2>=0.5.3",
   "matplotlib",
   "numpy",
   "pandas",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,9 @@ keywords = [
   "afm",
   "image processing"
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 dependencies = [
-  "igor2>=0.5.3",
+  "igor2",
   "matplotlib",
   "numpy",
   "pandas",
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -12,7 +12,7 @@ from typing import Any
 import numpy as np
 import pandas as pd
 import pySPM
-from igor import binarywave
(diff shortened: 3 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -12,16 +12,16 @@ jobs:
...
       - uses: actions/checkout@v3
       - name: Set up Python
         uses: actions/setup-python@v4.3.0
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Upgrade pip and install test dependencies
         run: |
-          pip install -e .
+          pip install --upgrade pip
```
SylviaWhittle, line 24: Just curious, why remove the `pip install -e .`? I would have thought it was needed?

SylviaWhittle (approved):
Ran locally and seems to work 👍 
Installing automatically installs `igor2` now and `numpy` is updated.

The comment is just curiosity.

.github/workflows/tests.yaml
ns-rse: Its because I now understand `pip` a little better and `pip install -e .[tests]` installs not just the tests but the optional dependencies defined under `tests` in addition to the the `dependencies`.

PR #695, opened 2023-10-12
Sylvia whittle/height trace redo

SylviaWhittle:
Adds the option of saving grain trace heights & grain trace cumulative distances to JSON file.

Trace heights mean the pixel heights of every coordinate in the ordered trace. It is an array / list.

Trace cumulative distances are the distances for each pixel as you traverse the ordered trace. Distances only go up by 1.0 or sqrt(2) since pixels are either adjacent or diagonal. Hope that makes sense.

They are saved in files like this:

image

The structure of the JSON is a dictionary where the keys are the molecule number and the values are the lists of data (heights or cumulative distances).

[Technical notes]

Internally I kept the arrays as lists, since JSON cannot serialise numpy arrays, and I thought it would be preferable to avoid adding a script to deconstruct the dictionary, then convert the arrays to lists and then re-constructing the dictionary again. Let me know if this is silly 😋 

I have kept the saving in `processing.py` as I plan for `run_topostats` to be generic and not handle extra output from workflows in the future. Feel free to disagree with me on this, but I think maybe we should let workflows save data as they see fit during processing of individual images without them having to return it to `run_topostats` since I foresee workflows saving **a lot** of random stuff that we don't necessarily want to write extra code to handle in `run_topostats`?


SylviaWhittle:
Adding a test for the `.json` output is blocked by #653

ns-rse:
Not looked at the code yet, and this may be part of the JSON output but there's no test to see what it looks like (sorry for banging the same :long_drum: again) but I'm wondering what the utility of knowing the heights without the locations is?

Sure summary statistics could be calculated, but if anyone wanted to plot or visualise the data elsewhere the location of the points would be required.

ns-rse:
> I have kept the saving in `processing.py` as I plan for `run_topostats` to be generic and not handle extra output from workflows in the future. Feel free to disagree with me on this, but I think maybe we should let workflows save data as they see fit during processing of individual images without them having to return it to `run_topostats` since I foresee workflows saving a lot of random stuff that we don't necessarily want to write extra code to handle in `run_topostats`?

I agree with this, each module of analysis should save whatever it should/is configured to do at each stage.

SylviaWhittle:
No longer blocked by #653

SylviaWhittle:
Put all stats into one file with image and grain number as the index. Also rename the index to a string of "grain_54" etc instead of "54"

SylviaWhittle:
There has been some debate about what format the files should be saved in, and if all traces should be in one file or multiple etc, and so I'm thinking before trying to refine this further, push to main and let people use it for a while to see what issues come up?

Happy to change how this works if people feel strongly on it though. Just this has been delayed due to indecision and also me not being available enough 😅

ns-rse:
:+1: Yes, definitely isolate the code into its own testable function. Test with small examples rather than full data set if possible too (even smaller than the cropped picture is likely possible, see examples of tests for `write_yaml()` in [`tests/test_io.py::test_write_config](https://github.com/AFM-SPM/TopoStats/blob/43410ab06bf366764271553a16f88d97f3c29999/tests/test_io.py#L116) which uses a dummy [`CONFIG`](https://github.com/AFM-SPM/TopoStats/blob/43410ab06bf366764271553a16f88d97f3c29999/tests/test_io.py#L39)).

SylviaWhittle:
Note to self: Make the heights and distances all in one JSON and change molecule index to a string "grain_54".

temp_json_experiment.ipynb
ns-rse, line 1: This Notebook can probably be removed.

topostats/default_config.yaml
```diff
@@ -51,6 +51,7 @@ grainstats:
   cropped_size: 40.0 # Length (in nm) of square cropped images (can take -1 for grain-sized box)
 dnatracing:
   run: true # Options : true, false
+  save_trace_json_data: false # Whether to save things such as height traces to JSON files.
```
ns-rse, line 54: Would you be able to add an equivalent row into `docs/configuration.md` for this new option please? :pray:

topostats/tracing/dnatracing.py
```diff
@@ -154,6 +158,86 @@ def trace_dna(self):
...
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        None
+        """
+
+        # Get the heights of each pixel in the ordered trace from the gaussian filtered image
```
ns-rse, line 175: Comment is redundant as the docstring should contain this information.

topostats/tracing/dnatracing.py
```diff
@@ -154,6 +158,86 @@ def trace_dna(self):
...
+            if abs(coordinates[i] - coordinates[i + 1]).sum() == 2:
+                distance += 1.41421356237  # Saves a sqrt call
+            else:
+                distance += 1.0
+            cumulative_distance_list.append(distance)
+        # Convert to array first to allow for multiplication by px_to_nm
+        # Then convert back to list since JSON cannot save as numpy array
+        # Better to do the conversion here than have to do it in the
+        # JSON saving script.
+        return list(np.asarray(cumulative_distance_list) * px_to_nm)
```
ns-rse, line 239: No need to convert to an array as can just use list comprehension
```suggestion
        return [distance * px_to_nm for distance in cumulative_distance_list]
```

ns-rse: Also does this make redundant the `recursive_numpy_array_to_list()` function if as noted `Better to do the conversion here than have to do it in the JSON saving script`?

ns-rse (changes requested):
Some comments in line but one thought that occurred to me whilst reviewing is whether we are creating additional boiler plate data for a very specific function.

The `.topostats` file which is used for saving results to uses HDF5 which is a very similar structure to JSON in terms of nested data. Would it not save us a large amount of work to have the height trace data saved to the HDF5 if people need to access it?

I'm also not sure about whether JSON is the best format for people to use (something we've discussed before) as its unfamiliar to many who are used to columnar/tabular data and this data itself isn't highly nested in nature.

Issue #692, opened 2023-10-10
Standardised plotting format


ns-rse:
It would be sensible to have a consistent style for TopoStats plotting of images.

This would entail setting up [Customizing Matplotlib with style sheets and rcParams — Matplotlib 3.8.0 documentation](https://matplotlib.org/stable/users/explain/customizing.html) so that there is a consistent and acceptable format.

+ Large fonts on axes
+ 4 x/y tick bins instead of 5 as with the font size increase, this can look messy

If there are other features please add them to this issue.

Issue #513, opened 2023-03-21
GitHub Actions fail on Windows systems


ns-rse:
Second time I've seen this now, tests fail with...

```
_______________________ test_run_topostats_process_all ________________________
caplog = <_pytest.logging.LogCaptureFixture object at 0x0000018608C1B070>
def test_run_topostats_process_all(caplog) -> None:
        """Test run_topostats completes without error when no arguments are given."""
        caplog.set_level(logging.INFO)
        # Explicitly force loading of topostats/default_config.yaml as I couldn't work out how to invoke process_all()
        # without any arguments as it defaults to 'sys.argv' as this is wrapped within pytest it picks up the arguments
        # pytest was invoked with (see thread on StackOverflow at https://stackoverflow.com/a/55260580/1444043)
>       run_topostats_main(args=["--config", f"{BASE_DIR / 'topostats' / 'default_config.yaml'}"])
tests\test_run_topostats.py:29: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
topostats\run_topostats.py:276: in main
    toposum(summary_config)
topostats\plotting.py:309: in toposum
    figures[var]["dist"]["figure"], figures[var]["dist"]["axes"] = topo_sum.sns_plot()
topostats\plotting.py:177: in sns_plot
    fig, ax = self._setup_figure()
topostats\plotting.py:156: in _setup_figure
    fig, ax = plt.subplots(1, 1, figsize=self.figsize)
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\matplotlib\pyplot.py:1501: in subplots
    fig = figure(**fig_kw)
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\matplotlib\_api\deprecation.py:454: in wrapper
    return func(*args, **kwargs)
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\matplotlib\pyplot.py:840: in figure
    manager = new_figure_manager(
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\matplotlib\pyplot.py:384: in new_figure_manager
    return _get_backend_mod().new_figure_manager(*args, **kwargs)
[...]

PR #707, opened 2023-10-23
Adds a complete custom matplotlibrc params file

### LINKED ISSUE #704: Expand plotting defaults and option for users to load their own mplstyle
#701 hard codes the `font.size` and `axes.titlesize` and make configurable the `ax.set_{x,y}ticks` in plotting.

To avoid future technical debt of having to remove or modify hard-coded variables if/when users require something other than the default's that are hard coded in `topostats/images.mplstyle` we should expand the defaults defined to include more variables and make it possible for users to specify their own `image.mplstyle` file. Further we should document for users how to find the `images.mplstyle` file that comes with TopoStats and how to use it. This seems like a middle ground of making a default "brand" for images generated by TopoStats but also making them configurable without further bloating `

Other values to consider including beyond those mentioned above are listed below, for a full list see the [the-`matplotlibrc`-file](https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file) example...

+ [ ] `axes.grid` - whether to show a grid, this might perhaps be useful when eye-balling plots to get a feel for the size of molecules (personally I find grids really useful on most graphs).
+ [X] `axes.titlesize`
+ [ ] `axes.titlelocation`
+ [ ] `axes.labelsize`
+ [ ] `font.family` the broad class of font to use.
+ [ ] `font.style`
+ [ ] `font.weight`
+ [X] `font.size`
+ [ ] `{x,y}tick.major.size`
+ [ ] `{x,y}tick.labelsize`
+ [ ] `grid.color`
+ [ ] `grid.linestyle`
+ [ ] `grid.linewidth`
+ [ ] `grid.alpha`
+ [ ] `legend.loc` - location of legend
+ [ ] `legend.frameon` - whether to frame the legend
+ [ ] `legend.fontsize`
+ [ ] `image.interpolation` - refer back to #457 
+ [ ] `image.cmap` - see [`topostats/theme.py`](https://github.com/AFM-SPM/TopoStats/blob/main/topostats/theme.py)
+ [ ] `savefig.dpi` - see #527 and #667
+ [ ] `savefig.format`
+ [ ] `savefig.transparent`

ns-rse:
Closes #704

Further to #692 and #701 this PR now includes a full example [matplotlibrc params file](https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file) at `topostats/topostats/mplstyle`. Only a handful of parameters have been enabled for now...

+ `image.cmap`
+ `hist.bins`
+ `figure.figsize`
+ `savefig.format`

Specific Changes

+ The `.mplstyle`  is also now included as part of the package data which is required otherwise it wouldn't have been bundled up into PyPI packages and none of the styles added in #701 would have been applied (I should have clocked this when reviewing, sorry).
+ A new function to load either the `topostats.mplstyle` or users specified file is introduced with simple tests (`tests/test_plottingfuncs.py::test_load_mplstyle` and associated `tests/resources/my_custom.mplstyle`).
+ Images generated by `tests/test_plottingfuncs.py` (which are saved in `tests/resources/img/*.png`) have all been updated. I'm not sure why this wasn't required when #701 was submitted as that should have updated the axis labels in all the tests, perhaps the font resizing wasn't actually being applied. These images have reduced in size slightly from 800x800 to 640x640, quite why I'm not sure since the `figure.figsize` is set to `8.0, 8.0` (which was the hard coded value in the `Images()` class) and DPI is set on a per-image basis via `plotting_dictionary.yaml`. If this is a serious issue please [create an issue](https://github.com/AFM-SPM/TopoStats/issues/new?assignees=&labels=enhancement&projects=&template=feature_request.md&title=). The reductions have been included as it reduces the size of the test-suite fractionally, if high resolution images of regions are required I would suggest loading the saved numpy arrays and plotting them anew in Matplotlib (see the plotting Notebooks for examples of how to do this).
+ In order to get the `nanoscope` and `gwyddion` colormaps that are defined in the `topostats/theme.Colormap()` recognised as a valid `cmap` by Matplotlib when specified as such in the `topostats.mplstyle` required registering the colormap when the class is initialised in `__init__.py` (see [solution](https://stackoverflow.com/a/74997755/1444043)).

This makes it easy to further customise additional aspects of the default plots produced by TopoStats and aligns more smoothly with Matplotlib configuration styles, rather than having to add configuration options to the `Images()` class which carries further maintenance overhead and complexity.

Where fields were specified as individual fields in the `default_confg.yaml` most have been removed in favour of having all parameters for plotting specified in a single file. A few have been left for convenience of modifying such as the file-type.

It is possible for users to specify their own style file either as an argument in `default_config.yaml` under `plotting: style: <file_path>` or via the command line argument `--matplotlibrc`.  Further individual parameters that have been set can also be over-ridden when instantiating the `Images` class itself. Note not all options are configurable by rcParams.

If people feel it would be useful to generate such a file in the same manner as configuration files are created please [create an
issue](https://github.com/AFM-SPM/TopoStats/issues/new?assignees=&labels=enhancement&projects=&template=feature_request.md&title=).


topostats/entry_point.py
```diff
@@ -52,6 +52,12 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Path to a YAML configuration file for summary plots and statistics.",
     )
+    process_parser.add_argument(
```
SylviaWhittle, line 55: Just commenting here to make sure you're aware that you haven't added this argument for the legacy `run_topostats` entry point. I agree with this, as I think we should discourage use of the old way, but wanted to make sure this was intentional 👍

topostats/plottingfuncs.py
```diff
@@ -75,6 +75,14 @@ def dilate_binary_image(binary_image: np.ndarray, dilation_iterations: int) -> n
     return binary_image
 
 
+def load_mplstyle(style: dict) -> None:
```
SylviaWhittle, line 78: Should the type hint here be `str`? While `plt.style.use()` can take dictionaries, they don't appear to be used in this PR for styles I think? Possibly left in from a previous iteration where you used dictionaries?

It would be odd (but technically allowed in python, as I have just found out when testing this function in a notebook) to test equality between a dictionary and a string.

If allowing dictionaries in this function is intended / wanted, then perhaps it would be good to add a check (and tests) for what type the incoming `style` is, to prevent comparison between a dictionary and a string, which could maybe cause unexpected behaviour, just a thought!

tests/test_plottingfuncs.py
```diff
@@ -31,6 +37,24 @@ def test_add_pixel_to_nm_to_plotting_config(plotting_config_with_plot_dict):
...
 
+@pytest.mark.parametrize(
+    ("style", "axes_titlesize", "font_size", "image_cmap", "savefig_format"),
+    [
+        ("topostats.mplstyle", 18, 16.0, "nanoscope", "png"),
+        ("tests/resources/my_custom.mplstyle", 34, 5.0, "viridis", "svg"),
+    ],
+)
+def test_load_mplstyle(
+    style: dict, axes_titlesize: int, font_size: float, image_cmap: str, savefig_format: str
```
SylviaWhittle, line 48: Type hint `str` for "style" I think?

SylviaWhittle (changes requested):
I am now able to use my wonderful, publication-ready matplotlib style file!
image


This PR clears up a lot of the ambiguity between what is a TopoStats plotting option, and what is a matplotlib option. It also prevents the inevitable expansion of the config file to eventually containing every matplotlib option inside it 😆 thank you!


Just a couple things I thought I should mention and check (an issue with validation), but aside from that, as you can see above, it works perfectly!

topostats/entry_point.py
ns-rse: Yep, trying to steer people to the newer approach. I'd almost be inclined to remove the older `run_topostats` entry point to do so. Captured in #712

topostats/validation.py
```diff
@@ -215,6 +215,14 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 False,
                 error="Invalid value in config for 'plotting.run', valid values are 'True' or 'False'",
             ),
+            "style": And(
+                str,
+                Or(
+                    "topostats.mplstyle",
+                    None,
```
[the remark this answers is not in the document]
ns-rse: Brilliant, thanks for your diligence checking this @SylviaWhittle :+1: 

Will look at this later, got to teach in a couple of hours.

topostats/plottingfuncs.py
ns-rse: Good spot, it should indeed be `str | Path` since in the first instance its `topostats.mplstyle` otherwise its the path to a the file that should be loaded.

Ta :+1:

ns-rse:
Tests failed again due to the Tcl error on Windows which randomly crops up (#513) re-running those that failed and they all passed. :shrug:

SylviaWhittle (approved):
Looks good to me! Thanks for the edits!

PR #711, opened 2023-10-28
Fix flattening step not running when scar removal disabled

SylviaWhittle:
This PR just fixes an issue I introduced accidentally when implementing `nonlinear_polynomial_removal`, where that flattening step is not run when scar removal is disabled. I initially only incorporated the nonlinear polynomial removal into the pipeline when scar removal is enabled.

This happened because I forgot to add the "initial_nonlinear_polynomial_removal" and "masked_nonlinear_polynomial_removal" to the `else` branch of the checks to see if scar removal is enabled. This resulted in the scar removal result images, "initial_scar_removal" and "secondary_scar_removal" to be set to the flattening step before the nonlinear polynomial removal, ignoring the nonlinear polynomial removal.

I have fixed this by simply switching the images that "initial_scar_removal" and "secondary_scar_removal" are set to when scar removal is disabled, from "initial_quadratic_removal" and "secondary_quadratic_removal" to "initial_nonlinear_polynomial_removal" and "masked_nonlinear_polynomial_removal". This means that nonlinear polynomial removal is now incorporated when skipping scar removal.

I've tried to explain this clearly, but there's a lot of words! Let me know if I've not been clear.

### DIFF
--- a/topostats/filters.py
+++ b/topostats/filters.py
@@ -431,7 +431,7 @@ processed, please refer to <url to page where we document common problems> for m
             )
         else:
             LOGGER.info(f"[{self.filename}] : Skipping scar removal as requested from config")
-            self.images["initial_scar_removal"] = self.images["initial_quadratic_removal"]
+            self.images["initial_scar_removal"] = self.images["initial_nonlinear_polynomial_removal"]
 
         # Zero the data before thresholding, helps with absolute thresholding
         self.images["initial_zero_average_background"] = self.average_background(
--- a/topostats/filters.py
+++ b/topostats/filters.py
@@ -424,7 +424,7 @@ processed, please refer to <url to page where we document common problems> for m
         run_scar_removal = self.remove_scars_config.pop("run")
         if run_scar_removal:
             LOGGER.info(f"[{self.filename}] : Initial scar removal")
-            self.images["initial_scar_removal"], _scar_mask = scars.remove_scars(
+            self.images["initial_scar_removal"], _ = scars.remove_scars(
                 self.images["initial_nonlinear_polynomial_removal"],
                 filename=self.filename,
                 **self.remove_scars_config,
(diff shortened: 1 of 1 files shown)


topostats/filters.py
```diff
@@ -431,7 +431,7 @@ def filter_image(self) -> None:
             )
         else:
             LOGGER.info(f"[{self.filename}] : Skipping scar removal as requested from config")
-            self.images["initial_scar_removal"] = self.images["initial_quadratic_removal"]
+            self.images["initial_scar_removal"] = self.images["initial_nonlinear_polynomial_removal"]
```
ns-rse, line 434: In the previous `if run_scar_removal:` section we use tuple unpacking to assign the results of `scars.remove_scars()` to `self.images["initial_scar_removal"]` and `_scar_mask`.

But in this `else:` section nothing gets assigned to `_scar_mask`.

Do we need to assign an array to `_scar_mask`? If not the tuple unpacking in the `if run_scar_removal:` should perhaps be `self.images["initial_scar_removal"], _ = ...` as if `_scar_mask` isn't used I would have expected `pylint` to throw an error saying that variable was assigned but not used.

Same is true of the other change in this PR.

SylviaWhittle: Fixed 👍

topostats/filters.py
```diff
@@ -424,14 +424,14 @@ def filter_image(self) -> None:
         run_scar_removal = self.remove_scars_config.pop("run")
         if run_scar_removal:
             LOGGER.info(f"[{self.filename}] : Initial scar removal")
-            self.images["initial_scar_removal"], _scar_mask = scars.remove_scars(
+            self.images["initial_scar_removal"], _ = scars.remove_scars(
```
ns-rse, line 427: One for subsequent work but if the scar mask is not required we could remove it from being returned (#714).

ns-rse (approved):
Great, thanks @SylviaWhittle 

If we don't use the mask I don't think there is any point in returning it, but perhaps its useful for development. Created #714 to discuss and action if required.

PR #740, opened 2023-11-29
Merging confidence value work into cats branch

MaxGamill-Sheffield:
Adds confidence and average confidence values into the nodestats and grainstats dataframes


ns-rse:
For reference these merges can also be done locally. Its not too dissimilar from merging `main` into your local branch, but since you are developing them locally there is no need to `git pull` the branch you want to merge in...

```
git checkout maxgamill-sheffield/cats
git merge maxgamill-sheffield/cats-confidence
```

...should do the same as making a Pull Request on GitHub to merge the branches.

PR #749, opened 2023-12-05
Add snoop as a dependency but disable by default

ns-rse:
I came across [snoop](https://github.com/alexmojaki/snoop) which provides a powerful set of debuggin tools and think it would be useful to use it in TopoStats.

This PR adds it as a dependency and imports during package initialisation. By default though it is disabled so that performance in not impacted, this is done by `snoop.install(enabled=False)` in `__init__.py`.

To enable and using during development add the `@snoop` decorator to a function, method or class and in the same file re-enable dynamically by calling `snoop.install(enabled=True)`. See the documentation for more information on usage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,7 @@ dev = [
   "pyupgrade",
   "pytest-durations",
   "pytest-xdist",
+  "snoop",
 ]
 pypi = [
   "build",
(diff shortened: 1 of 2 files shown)


ns-rse:
Thanks for the review @SylviaWhittle I'm liking how useful `snoop` is too, just going to take a little bit of practice to get used to it.

I've not added any dependencies though, I just alphabetised them :wink:

Issue #750, opened 2023-12-07
Use scikit-image to calculate minimum and maximum feret and dependent statistics


ns-rse:
We currently calculate our own `min_feret` and `max_feret` but these are methods to the `Grains()` class.

Whilst looking to implement #748 I discovered that the `skimage.measure.regionprops` which we calculate on all grains includes the `max_feret` property, but no `min_feret`.

However an issue from a few years ago on scikit image (see [thread](https://github.com/scikit-image/scikit-image/issues/4466)) requested the ability to pass user-defined functions to `regionprops` so that additional metrics can be calculated (the original proposer mentions the minimum feret distance as being one desirable metric).

This feature has been implemented in Scikit Image by [#4810](https://github.com/scikit-image/scikit-image/pull/4810) and is available to use.

We can therefore leverage the functionality to reduce our codebase and perhaps improve the speed of calculations as the methods implemented in Scikit-Image are vectorised, including the calculation of convex hulls.

ns-rse:
Closing, scikit-image doesn't directly calculate minimum feret anyway, will implement a module and functions to do so as part of #748 and may push up-stream.

Issue #748, opened 2023-12-05
Add horizontal grain profiles


SylviaWhittle:
**Is your feature request related to a problem? Please describe.**
It has been requested that topostats have a feature similar to Gwyddion where the horizontal (in the image) height profile of a grain would be taken. In other words, a horizontal line is placed over the molecule, and the pixels that form the line have their heights measured to form a profile for that molecule.

**Describe the solution you'd like**
During grainstats, a horizontal, one pixel thick line of pixels should be formed, with the length sufficient to cover the entire molecule from left to right, and the line intersecting the centroid of the molecule. The heights for those pixels should be obtained and saved as a line plot. Probably only saved with image_set all.

While it does seem arbitrary to just take one line through the centre, and we will miss a lot of information, this is is not an issue, as this line profile is just to get an idea of the height profile of the molecule.

This is a heavily requested feature and should be quite high priority.
It should be relatively trivial to implement. (I think?)

slackline:
Ultimately the profile of any horizontal line is available as the Numpy array for the given grain so its "just" a case of finding the centroid of a grain (and skimage has functions for doing this) and extracting that row and the plotting.

Some random questions off the top of my head (not necessarily directed at you @SylviaWhittle ).

At what point in the processing should we extract this information, is it from the untraced or traced grain? I suspect the former but would like to check

What should be done for linear molecules?

In the future is any other profile likely to be required other than that which passes through the centroid? If so will these always be horizontal or are other angles/paths required such as the profile through the widest part of a circular molecule[^1]?


[^1]: I would be surprised if the orientation of a molecule is always as desired in the horizontal plane and is essentially random.

alicepyne:
I think this would be an option - like “crop” in the config file, Where if the suer wants to they would plot a line of “user defined length” or through the centre of the blob from the centre only. This is to use as single examples alongside the numbers obtained from topostats e.g. height, are volume, partly for sense checking, partly for visualisation. 

If you did this over all the molecules, you would get an average. Which would show you heterogeneity you display below.

ns-rse:
Wouldn't you have heterogeneous measurements across a single molecule depending on the orientation though?

Would it perhaps be useful to use a slice along the line parallel to the minimum feret diameter which passes through the centroid as that would then pick up the height profile along (roughly[^1]) the longest slice?

I'm probably over thinking this though so please do say but I'm trying to anticipate other scenarios and think about general solutions at the outset.

[...]

PR #755, opened 2023-12-11
New submodule for feret calculations

ns-rse:
New submodule `measures.feret` with functions for calculation of min and max feret and the associated co-ordinates.

Utilises code from Skan developer see [gist](https://gist.github.com/VolkerH/0d07d05d5cb189b56362e8ee41882abf) and as [suggested](https://github.com/scikit-image/scikit-image/issues/4817#issuecomment-758900250) it adds tests. In doing so I found sorting points prior to calculation of upper and lower convex hulls missed some points.

I'm also not sure about the `rotating_calipers()` function as it doesn't seem to return all pairs formed across the points from the upper and lower convex hulls and so have included but not used the `all_pairs()` function which does, although if used in place this doesn't return the minimum feret correctly, rather just the smallest distance.

Currently some of the tests for the `curved_line` fail too.

The intention is to use this without instantiating the `GrainStats` class to be used in profiles (#748) and could serve as a concise stand-alone set of functions outside of `GrainStats` class which currently has static methods for the calculations (#750).

## Benchmarking

In light of #750 and re-implementing feret calculations as a new sub-module I wanted to see how this compares in terms of performance to those in `GrainStats()` class so have run some very basic benchmarking. Note this is not the full pipeline of taking labelled images and finding the outlines/edge points which are required as inputs to the calculations, its purely on the calculation of min-max feret from the edge points.

```
import timeit

import numpy as np
from skimage import draw

from topostats.measure import feret
from topostats.grainstats import GrainStats

holo_circle = np.zeros((14, 14), dtype=np.uint8)
rr, cc = draw.circle_perimeter(6, 6, 5)
holo_circle[rr, cc] = 1

holo_circle_edge_points = np.argwhere(holo_circle == 1)


%timeit feret.min_max_feret(holo_circle_edge_points)

83 µs ± 686 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit GrainStats.get_max_min_ferets(holo_circle_edge_points)

1.06 ms ± 10.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

So this new implementation appears faster.


ns-rse:
Please hold back on this, I've been working on simplifying the tests and think there is a problem with the Rotating Caliper algorithm so will be checking that next week.

ns-rse:
The inverse gradient method of calculating the triangle height formed when determining the minimum feret results in `ZeroDivisionError` when the base of the triangle is horizontal so switching methods (thanks for the comprehensive suggestions of both @SylviaWhittle).

In order to check everything I wrote a plotting function (see previous commit) and whilst many of the minimum feret distances are correct there are still instances where the triangle height is calculated _outside_ of the convex hull which is baffling. For the most part not an issue but in the `holo_ellipse_angled` this **is** an issue.

Figures below show the problem.

The same Graham Scan algorithm is used for the rotating caliper pairs as in GrainStats so unsure if this is an issue there and can't check as co-ordinates aren't currently returned (possibly bounding box may be informative as I think I saw that was calculated for the feret distances, but inclined to solve this in current code base).

Slightly confusing plots but they show the lower and upper hulls in red/green (apologies if anyone is colourblind) with the rotating calipers in yellow. The Triangle heights are in blue and it is these that are wrong for some, in most cases this isn't an issue as the min feret (in magenta as is maximum feret) are elsewhere)....

quadrilateral_triangle_outside
tiny_ellipse_triangle_outside
ellipse_triangle_outside
horizontal_ellipse_triangle_outside
curved_line_triangle_outside

BUT that won't always be the case as the following shows, the minimum feret is determined to be a line outside of the convex hulls and taking the trace here would give an incorrect set of heights.

angled_ellipse_triangle_outside_IMPORTANT_AS_ITS_WRONG_COORDS

These tests may seem/feel like over-simplifications but if the code isn't correct for these then we have no guarantee it will be correct for more complex shapes.

I feel like I'm close to solving this and its important to me that its correct as I would hate to have results published that contained a corrigendum.

Once I've fixed these I will reinstate the tests for filled objects and images with multiple masks (the later isn't essential but it may be useful to users using the functions interactively and will "just work (TM)" once everything else is correct as its just a wrapper).

ns-rse:
Started investigating this and have worked out where the problem arises and have some thoughts on how to resolve it.

Taking this small ellipse...

image

The `rotating_calipers()` function finds the first caliper between `[5,2]` and `[1,2]` that is the horizontal line across the middle of the ellipse.

The next point on the adjacent hull is `[4,1]` thus the triangle is formed by `[[5,2],[4,1],[1,2]` which is a scalene triangle

ellipse_triangle_001

Extend the base and we have...

ellipse_triangle_002

The height of the triangle, which is from the apex to the perpendicular intersection with the base _is_ outside of the original triangle...

ellipse_triangle_003

## But this doesn't matter its not the minimum feret

In this example its not but as noted above this tilted ellipse has a similar error and the minimum feret (short pink line) _is_ outside of the shape.

image


Checking this we have...

### Caliper/Triangle 1

ellipse_angled_triangle1_003

### Caliper/Triangle 2

ellipse_angled_triangle2_003


### Caliper/Triangle 3

**NB** This is a right-angled triangle so the height is within the bounds of the triangle and what we want.

ellipse_angled_triangle3_003

### Caliper/Triangle 4

ellipse_angled_triangle4_003


### Caliper/Triangle 5

ellipse_angled_triangle5_003

### Caliper/Triangle 6

**NB** This is a right-angled triangle so the height is within the bounds of the triangle and what we want, you can't see it because its the black side of the triangle and I only wrote a short hacky function to draw these plots.

ellipse_angled_triangle6_003




Ok, this is doing what we have asked which is find the height of the triangle formed by the calliper pairs and the next point on the convex hull, but its _not_ what we want to know and its actually wrong if the triangle formed is scalene.




### Why is the minimum feret the co-ordinates _outside_ of the triangle and not the height of caliper/trianlge 2?

The `rotating_calipers()` function returns a generator, for each pass round the rotating calipers it then yields the `min_feret`, the `caliper` co-ordinates and the co-ordinates of the `min_feret` (the apex to the point perpendicular to the base of the triangle, which as we have seen can be  outside of scalene triangles).

The `min_ferets` and `min_feret_coords` are then `zip()` up and passed to the `min()` function which returns the minimum. It does so based on the first element, if any of the first elements (i.e. `min_feret`) match then they are sorted by the `min_feret_coords`.

This list looks like this...

```
[
    (5.0, [[6.0, 2.0], [1, 2]]),
    (5.0, [[1.0, 5.0], [6, 5]]),
    (2.82842712474619, [[3.0000000000000004, 2.0000000000000004], [1, 4]]),
    (2.82842712474619, [[0.0, 3.0], [2, 1]]),      
    (7.071067811865475, [[6.661338147750939e-16, 2.999999999999999], [5, 8]]), 
    (7.071067811865475, [[6.0, 7.0], [1, 2]])
]
```

The third and fourth elements have the _same_ minimum feret distance (and you can see that the `min_feret` in the coloured plots in pink is indeed the same distance as the caliper/triangle 3 height but because the _co-ordinates_ of the feret line in element 4 are smaller it is selected as the minimum feret.

### What to do?

The Graham Scan algorithm is the basis for this implementation (as it is in the current [`grainstats.GratinStats.get_max_min_feret()`](https://github.com/AFM-SPM/TopoStats/blob/6de80e534367990248d5595d310ae12a83a31851/topostats/grainstats.py#L961) function, note the calls to `GrainStats.get_triangle_height()` which is taking the caliper and the next point on the convex hull).

If algorithms don't work on small examples we can't expect them to generalise well and I would never have come across this issue if I hadn't created the various small simple shapes and tested them! 

I think a sensible approach here might be to have a check made before finding the minimum feret that checks if the co-ordinates of the triangle height are on the convex hull. If either of these points are outside of the shape we can drop that feret distance/co-ordinates from the list prior to finding the minimum which means we should only ever get feret distances that are within the convex hull and we therefore get not just the correct distance but the correct co-ordinates through which we can take a profile (which is what has stimulated this work in the first place!).

I know you're incredibly busy @SylviaWhittle but as you tussled with this problem in the past your thoughts would be appreciated as would any thoughts from @MaxGamill-Sheffield or @llwiggins .

ns-rse:
## Notes on solutions for excluding triangle heights from scalene triangles that are outside of the convex hull

### Matplotlib Path

Gave this method a whirl but as noted in the [documentation](https://matplotlib.org/stable/api/path_api.html#matplotlib.path.Path.contains_point).


> The current algorithm has some limitations:
> 
> + The result is undefined for points exactly at the boundary (i.e. at the path shifted by radius/2).
> + The result is undefined if there is no enclosed area, i.e. all vertices are on a straight line.
> + If bounding lines start to cross each other due to radius shift, the result is not guaranteed to be correct.

Had a play with the value of `radius` but without success which excludes this method as many of the points we will be assessing will be on the boundary.

Did like the fact you could assess a vector of co-ordinates rather than having to loop over them though.

ns-rse:
This is ready for review.

Once approved and merged I shall proceed with extracting the profiles along the min/max feret lines (the main feature from #748) and addressing #798 as I think the code implemented in `GrainStats` suffered from the above mentioned problem that scalene triangles causes when calculating minimum feret distances.

Test coverage is low on this PR as I've not tested the `topostats.measure.feret.plot_feret()` function I wrote to aid investigation. Can address this but its really for debugging and would bloat the repository adding figures for each scenario (although could easily keep them as small as possible in order to minimise that side effect).

ns-rse:
**Update** - Decided to try the [ray tracer algorithm](https://www.algorithms-and-technologies.com/point_in_polygon/python) and as we want to include points on the convex hull have used `>=` and `<=` but have a slight problem with tests failing for points on the left and top edge of shapes as the `odd` value gets flipped twice. I think since these are convex hulls rather than concave we might be able to to use the number of times crosses are made being `>= 1` as indicative of a point being on or within but will check next week.

ns-rse:
After discussion with @SylviaWhittle  and @llwiggins it was decided that because of the issue of minimum feret coordinates sometimes falling outside of the convex hull that we should not use it as a consistent line through the grain/image.

Instead we will use the maximum feret and a line perpendicular to the mid-point of the  maximum feret.

As such all attempts to work out if minimum feret coordinates/lines were inside or on the edge of a polygon have now been removed and the function returns a dictionary of...

+ `min_feret`
+ `min_feret_coords`
+ `max_feret`
+ `max_feret_coords`

The dictionary should be conducive to being saved in the HDF5 output (perhaps a separate issue to do so).

@SylviaWhittle : Would it be useful to have an additional element in the returned dictionary for the triangle co-ordinates from which the minimum feret (triangle height) co-ordinates are derived?

SylviaWhittle (approved):
I ran this locally using some test shapes that I generated and it works perfectly. Returning the coordinates of the min and max ferets is really useful, thank you.

The code reads beautifully and is very comprehendible. I also like the several options that you've added for running it.

Thank you for your patience and sorry for not reviewing it last week as requested.

PR #758, opened 2023-12-13
Adds create-config entry point

### LINKED ISSUE #757: Entry point to `create-config`
As we re-model the interface we will need to add an entry-point to `create-config`.

ns-rse:
Closes #757

Adds a `topostats create-config` entry point and removes reference to `run_topostats`.

As we move towards a new release having a modular interface is desirable and that includes generation of sample configuration files. This commit brings the "swiss army knife" approach of `topostats <subcommand>`to parity with the older `run_topostats` by adding `topostats create-config` to generate configuration files for users to edit.

Documentation has been updated to reflect these changes, including the `README.md`. Because we have versioned documentation users will still be able to refer to the older documentation.

For now the `run_topostats` method remains but raises a deprecation warning (`UserWarning`) if the `--create-config-file` option is used.

## Future Extension

Currently TopoStats is designed to work primarily with nucelic acid samples and in most cases the `topostats/default_config.yaml` which serves as a basis for generating configuration samples is a good starting point.

However, there are other types of samples that may be scanned, silicone, large molecular structures such as proteins, prokaryotic and eukaryotic cells. These might require a different set of default parameters.

With this in mind I have implemented what I hope is a method for provisioning such configurations once sensible defaults have been determined. This would work via the `topostats create-config --config <sample_type>` option and for each `<sample_type>` there would be a `topostats/<sample_type>_config.yaml` in the TopoStats repository/package from which the resulting output would be derived. 

To this end a check is made whether the specified configuration is available and it fails telling the user that no such config is available if incorrectly specified. An appropriate test has been added to the test suite to cover this.

**NB** This work was undertaken as a consequence of re-writing the `.github/ISSUE_TEMPLATES/` so that the instructions could use this new entry point (see #759 ).


ns-rse:
I've addressed #761, once this PR is approved and merged I'll add that in (trying to stay focused and avoid drift on PRs and keep it to small chunks of work).

PR #752, opened 2023-12-07
reduced max trace coord idx by 1 so within image bounds

MaxGamill-Sheffield:
fixes #751

Ensures that if the max fitted coord looks outside the image, it's max is the end of the image. 

Previously this matched the max shape of the array i.e. arr.shape=(10,10), but index was set to 10 which is outside the max index.

Seems like it was only an issue now because Eddie's images are very zoomed in and the `index_width` was set to 3nm which is quite a few pixels and exceeds the padding value of 1.

PR #760, opened 2023-12-14
reduced max trace coord idx by 1 so within image bounds

ns-rse:
Further to #752 I've worked through this for my own clarity, apologies for the protracted length and [Rubber Duck](https://en.wikipedia.org/wiki/Rubber_duck_debugging) debugging.

From #752 [comment](https://github.com/AFM-SPM/TopoStats/pull/752#issuecomment-1855640458)

>
> I agree that the coords should not be < 0.

This is the easy bit! Given image...

``` python
import numpy as np

test = np.asarray([[0, 0, 0, 0, 0],
                   [0, 1, 0, 0, 0],
                   [0, 0, 1, 0, 0],
                   [0, 0, 0, 1, 0],
                   [0, 0, 0, 0, 0],
                   ])
co_ordinates = np.argwhere(test == 1)
index_width = 2
nrows, ncols = test.shape
trace_coordinate = co_ordinates[0]
```
Taking the first co-ordinate, currently what we test is...

```python
trace_coordinate[0] < 0
```

We know that `trace_coordinate[0]` (and in turn `trace_coordinate[1]`) will never be < 0 so the inequality currently in place `if trace_coordinate[0] < 0:` doesn't make sense.

What we want to know is if the value of `trace_coordinate[0]` _minus_ the `index_width`, which will be the left/upper padding area, will be outside of the image, so we can update our inequality to reflect this.

```python
trace_coordinate[0] - index_width < 0
```

Now lets consider the other side of the image (the right side and bottom)...

> So my initial thought was that because number_of_rows and number_of_columns come from the np.ndarray.shape property,
> the index of this would be off by 1 when indexing.
>
> i.e. with a shape of (10,10), the variables number_of_rows etc would be 10 but only be indexed upto 9. Thus if the
> coord value satisfied: trace_x >= (self.number_of_rows - index_width), the value cannot be set to self.number_of_rows
> - index_width as the diagonals would exceed the bounds. If index_width=0, condition: trace_x >= self.number_of_rows >>
> trace_x = self.number_of_rows and x would exceed the index bounds.

I'll stick with the smaller example above and take the last point

``` python
trace_coordinate = co_ordinates[-1]
```

Now further down the code in lines 259-304 the `x_coords` and `y_coords` are calculated using `trace_coordinate[0] - index_width` and  `trace_coordinate[0] + index_width` (and the same for columns `trace_coordinate[1] - index_width` and  `trace_coordinate[1] + index_width`). It is this second value (the resulting of adding the `ìndex_width`) that we want to test in a similar manner is not outside the bounds of the image, i.e. `trace_coordinate[0] + index_width` does not exceed the width (in indices!) of the image.

Thus I think the `elif:` should be the following (NB `nrow` is equivalent to `self.number_of_rows`)...

``` python
trace_coordinate[0] + index_width > nrow - 1
```

As we don't want either the `x_coords` nor `y_coords` to exceed the image dimensions (in terms on Numpy indexing). Ok we now have that check in place, but what do we want to do if this inequality is met? We want it to be equal to the number of rows/columns minus the `index_width` but since we obtain the number of rows/columns from the `nrow`/`ncol` we have to subtract another `1` from this, which is the solution you came up with @MaxGamill-Sheffield. :+1:

At this point the tests still fail, but I've no way of knowing if the logic is correct. Further we have duplication of code, there are two sets of essentially the same thing, one dealing with rows and one dealing with columns. How to check this is working and reduce the amount of code? Simple we break it out into its own function and write a test!

To this end I've introduced a new function `topostats.utils.bound_padded_coordinates_to_image()` which does this and have tried to explain in the docstring why this function is needed and what it does (slightly challenging without knowing what the subsequent steps taken are but hopefully clear enough). The tests demonstrate that the co-ordinate is adjusted such that if the padding goes outside of the image boundary the position of the initial point is adjusted so that the subsequent padding to get the heights of a 3nm boundary will stop at on the edge of the image.

The tests all pass for this function and now that I've convinced myself that the co-ordinates are correctly adjusted if the padding (`image_width`) would take the points used subsequently outside of the images boundaries I'm happy to substitute this function in the `get_fitted_traces()` and run our tests that were previously failing again  (`tests/tracing/test_dnatracing_multigrain.py::test_trace_image`).

On doing so we find that the first three of the parametisation work, but the last (which uses thinning method of skeletonisation) fails as the co-ordinates of the start of first image (not the array sizes as I incorrectly wrote in #752) are slightly different (they were previously as `71, 78` but are returned as `71, 81` on my system and were `71, 77` in the CI that failed). BUT as noted in the docstring to the function that test is actually incomplete and it only checks the start and end coordinates of the ordered trace and not the fitted trace that these changes impact.

This is a _very_ roundabout way to have...

1) Discovered why the tests were failing.
2) That the tests are incomplete and need improving. 
3) That we had some incorrect logic in place that meant fitting of traces was incorrect.

Fortunately 1) and 3) have been addressed in this Pull Request. Its not entirely clear to me why the thinning algorithm should have changed the end location of the thinned skeleton but I do have a vague recollection that some of the thinning algorithms didn't give consistent results.

However, this has been a very useful exercise as it identified problems and hopefully by breaking out the code into its own function and writing tests we can have greater confidence that the padding used in fitting traces is more robust and behaves correctly when points are very close to the border of the image.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2,10 +2,12 @@
 from pathlib import Path
 
 import numpy as np
+import numpy.typing as npt
 import pytest
 
 from topostats.utils import (
     ALL_STATISTICS_COLUMNS,
+    bound_padded_coordinates_to_image,
     convert_path,
     create_empty_dataframe,
     get_thresholds,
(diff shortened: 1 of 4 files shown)


MaxGamill-Sheffield:
Yep sounds good to me! I'll merge this into my branch and then we can push them into main!

PR #768, opened 2023-12-20
Adds numpydoc-validation to pre-commit

ns-rse:
Further configuration can be undertaken by adding options to [`docs/conf.py`](https://numpydoc.readthedocs.io/en/latest/install.html).


ns-rse:
Needs a bunch of linting to address current doc strings which don't conform to numpydoc style. Will address next year.

PR #773, opened 2024-01-04
Adds entry point to create matplotlibrc parameters file

### LINKED ISSUE #761: Generate matplotlibrc params file
As anticipated in #707 it would be helpful to be able to generate a matplotlibrc params file in a similar way to the config file. It took a while for me to work out how I could change the style of the image output and it currently requires searching around in the code folders to find the file to copy. 
Being able to generate your own would make changing important image formatting options, such as dpi and colour, a lot more user friendly. 

It would be helpful to be able to quickly generate a matplotlibrc params file that can be easily edited in a similar way to `--create_config_file` .

Otherwise the options could be returned to the config file. I certainly change these options fairly regularly while using topostats.

ns-rse:
Closes #761

Introduces the `topostats create-matplotlibrc` sub-command/entry point to write the default style file (`topostats/topostats.mplstyle`) to disk for modification and subsequent usage. The path and filename can be customised when invoking.

This doesn't necessarily mean that all aspects of image plotting are controlled since we set default filenames, titles and DPI and whether an image is in the `core` set via the `topostats/plotting_dictionary.yaml` file.

Documentation has been added that, hopefully, clarifies how to use this in light of this and points users to the relevant Matplotlib documentation for customising parameters in the generated configuration file.

@derollins : I've included you as a reviewer as you requested this feature in #761. I've tested and it seems to work from my perspective in that I can create a sample configuration file of a given name in a given location but I would be very grateful if you could give it a whirl and provide feedback.

To do so you need to something similar to the following...

```bash
cd path/to/TopoStats/git/repo  # modify to reflect your system
git checkout main
git pull
git checkout ns-rse/761-create-maplotlibrc
conda create matplotlibrc   # assuming you are using Conda environments
pip install -e .
topostats create-matplotlibrc --help    # Instructions should be useful to guide usage
```

Once done you can remove the conda environment with...

```bash
conda env remove -n matplotlibrc
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ _build/
 
 # Ignore topostats/_version.py as per setuptools_scm
 topostats/_version.py
+
+# default output directory, often common from testing
+output/
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -134,6 +134,59 @@ Configuration files are validated against a schema to check that the values in t
 expected ranges or valid parameters. This helps capture problems early and should provide informative messages as to
 what needs correcting if there are errors.
 
+## Matplotlib Style
+
+TopoStats generates a number of images of the scans at various steps in the processing. These are plotted using the
+Python library [Matplotlib](matplotlib.org/stable/). A custom
+[`matplotlibrc`](https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file) file is included in
+TopoStats which defines the default parameters for generating images. This covers _all_ aspects of a plot that can be
+customised, for example we define custom colour maps `nanoscope` and `afmhot` and by default the former is configured to
+be used in this file. Other parameters that are customised are the `font.size` which affects axis labels and titles.
+
+If you wish to modify the look of all images that are output you can generate a copy of the default configuration using
+`topostats create-matplotlibrc` command which will write the output to `topostats.mplstyle` by default (**NB** there are
+flags which allow you to specify the location and filename to write to, see `topostats create-matplotlibrc --help` for
+further details).
+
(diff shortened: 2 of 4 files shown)


derollins:
The `create-matplotlibrc` entry point worked for me and I was able to create and edit the file and change the colour scale of the images I was processing. 

This does allow important formatting changes to be made and it is good that all the various aspects of the images can be edited now. 

However from a broader usability perspective the key variables are hidden somewhat within a very large file, I had to try a few searches before I could find the right section to change the colour scale and the other setting I wanted to change, `dpi`, is either in a different section of the document or different file (I'm still not sure which). This makes these changes feel like 'advance user' features and I think setting the colour map and dpi are pretty basic features that almost all users will want to change.

derollins (approved):
Works as described. Allowing the generation of a matplotlibrc file with the new subcommand `create-matplotlibrc`

ns-rse:
The generated file is, in terms of variable order, a carbon copy of the default Matplotlib from [here](https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file) a small number of fields have been changed.

The DPI settings are handled on a per-image basis. I tried to describe how these are handled in the  addition to the `configuration.md` page (which renders to the website once this is merged).

I'm disinclined to make a highly configurable plotting interface to TopoStats when one already exists in Matplotlib which is what the rc-file is doing.

Images/graphs/figures have such a wide array of choices and tweaking them is in my view best done in a "Read-evaluate print loop" (REPL) manner in Jupyter Notebooks where instant feedback on the image is obtained. If multiple files need processing with the same parameters then once these have been determined they can be transferred to the configuration file and applied to a batch of images.

PR #777, opened 2024-01-12
Restore cmap to config; simplify config of  cmap/dpi/image format

### LINKED ISSUE #776: Move some params back to config file

Some commonly used parameters are found in the "intimidating" `.mplstyle` file and it can be hard to locate these due to it's size.

Users have requested that the image colourmap, savefig.format and savefig.dpi return back to the original config file

Even better might be if the dpi can be moved to the plotting config so it can be done on a per image basis making the plotting quicker as very high dpi's only need to be done for certain images e.g. dnatracing images.

_No response_

### Additional context

_No response_

ns-rse:
Closes #776

## TL;DR

I would really appreciate feedback on the documentation as without clear documentation it is perhaps confusing how the components interact and work and can be modified and getting this as clear as possible will be really helpful. This is the [configuration.md file](https://github.com/AFM-SPM/TopoStats/pull/777/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180).  Ignore the changes to the table at the top and please review and feedback on the section `### Further customisation`.

It won't be added to/rendered to the website until this PR is merged so I can't easily provide it as a web-page for review.

## Details

+ Aligns command line and configuration field names with those in matplotlibrc files.
+ Restores the `cmap` configuration option to `default_config.yaml` and introduces `savefig_dpi` option.
+ Adds command line options for setting DPI (`--savefig-dpi`), Colormap (`--cmap`)and Output file format (`--savefig-format`).
+ Expands documentation on how to use custom configuration files or command line options to set the DPI/Colormap/Output format.
+ Updates the header to `topostats.mplstyle` to explain how to use it as typically users will have created a copy of the file (after the convenience function `topostats create-matplotlibrc` was introduced with #773).
+ To achieve this the dictionary `config["plotting"]` needed explicitly updating as the `update_config()` function doesn't update nested configurations (since this is the first PR that introduces command line options that modify any of the values in the nested dictionaries).
+ Updates options for `topostats toposum`` to align with `savefig_format` and adds flag to entry point so output format is consistent.
+ Updates and expands the configuration documentation explaining how to use these conveniences.

As a consequence quite a few files are touched to ensure that validation and processing functions all have variables that align with those in the configuration.

## Testing

If users could test this it would be very much appreciated, if you use the Git installed version something like the following would switch branches and allow you test it.

```
conda create --name topostats-config # Create and activate a virtual env specific to this
conda activate topostats-config
cd ~/path/to/TopoStats
git pull
git checkout ns-rse/776-config-jigging
pip install -e .
topostats process --output-dir base
topostats create-config test_config.yaml   # Create test_config.yaml to try changing parameters
topostats process --config test_config.yaml --output-dir test1
topostats process  --output-dir test2 --savefig-dpi 10 --cmap rainbow --savefig-format svg
topostats process --config test_config.yaml  --output-dir test3 --savefig-dpi 80 --cmap viridis --savefig-format pdf
```

Each invocation of `topostats process` will save output to its own directory (either `base`, `test1`, `test2` and `test3`) for comparison. There should be differences between the scan plots under each run, i.e. `base` v `test_config.yaml` and saved under `test1` and those under `test2` and `test3` should also differ.


topostats/default_config.yaml
```diff
@@ -61,13 +61,15 @@ dnatracing:
 plotting:
   run: true # Options : true, false
   style: topostats.mplstyle # Options : topostats.mplstyle or path to a matplotlibrc params file
-  save_format: png # Options : see https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html
+  savefig_format: null # Options : null (defaults to png) or see https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html
+  savefig_dpi: null # Options : null (defaults to format) see https://afm-spm.github.io/TopoStats/main/configuration.html#further-customisation and https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html
```
derollins, line 65: Typo? Does it default to `figure` rather than `format`?

derollins:
Tested this branch as suggested above, it worked as described and both the updated config file and the new command line options all behave as expected.

ns-rse:
Thanks for checking @derollins and glad the tests work for you. 

Did you edit the `savefig_dpi`/`savefig_format`/`cmap` fields in the generated `test_config.yaml` so that output differed from `base` output?

Is this the desired functionality for changing these options (i.e. three methods firstly via a custom Matplotlibrc style file; secondly changing a `custom_config.yaml`; thirdly via command line options)?

Does the documentation make sense? (If you want a rendered view of it you can copy and paste [this](https://raw.githubusercontent.com/AFM-SPM/TopoStats/ecd6a2a037e0b10962a2c4f9fda77b9d4582bc02/docs/configuration.md) into [here](markdownlivepreview.com) then scroll down to the _Further Customisation_ section).


I've spent 4-5hrs digging through this and making the changes and would like to round it off whilst its fresh in my mind if its not right so I can get back to debugging some other work I'm developing with my TopoStats time next week.

derollins:
> Did you edit the `savefig_dpi`/`savefig_format`/`cmap` fields in the generated `test_config.yaml` so that output differed from `base` output?

Yes, changed all three of these in the generated test config file. So it is different from all the other folders.

This is a good improvement in functionality, it's really helpful to have all the most used parameters in one place (the main config file) and accessible from the command line for routine analysis and image generation. It's still helpful to have the matplotlib style file for more detailed changes and for matching presentation or journal style guides etc.   

The documentation seems clear, the only small thing I thought of was making it clear that the plotting dictionary is not a file to copy/ generate then edit like the config file and matplotlib style file but a single file to edit. Although I don't know if that would actually overcomplicate. 

The new header on the style file is very clear and much less intimidating! I'm no longer scared I will break everything haha.

Everything seems to be working how it should and this is a really helpful change.

ns-rse:
Thanks for the clarification @derollins 


What is most used might change between users and what they are trying to achieve. I've always seen TopoStats as a batch processing tool that produces standardised output for work in progress and expected that where bespoke images need customisation they would be plotted outside of this batch processing framework.

Glad you find the changes useful.

I'll await feedback/approval from other users before merging.

derollins:
> What is most used might change between users and what they are trying to achieve. I've always seen TopoStats as a batch processing tool that produces standardised output for work in progress and expected that where bespoke images need customisation they would be plotted outside of this batch processing framework.

This is true, hopefully the upcoming training sessions may help identify the needs of the 'general' users which may feedback to usability development moving forward.

topostats/topostats.mplstyle
```diff
@@ -1,21 +1,22 @@
...
 ## default Matplotlib values.
 ##
-## If you wish to change your default style, copy this file to your work directory
-## you MUST rename it to something _other_ than topostats.mplstyle. Edit fields as
-## required and include it by either editing the style field in config.yaml to point
-## to the desired file or with the command line option --matplotlibrc <filename>
-##
+## Once you have saved changes to your copy you can run topostats with it using
+## the --matplotlibrc option. If for example your copy is 'my_custom.mplstyle'
+## you would use it with the following
```
MaxGamill-Sheffield, line 17:
```suggestion
## Once you have saved changes to your copy you can run topostats with it using
## the --matplotlibrc option or updating the sytle sheet path in the plotting section
## of the configuration file. If for example your copy is 'my_custom.mplstyle' you
## would use it with the following
```

topostats/validation.py
```diff
@@ -229,7 +230,17 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
...
-            "save_format": str,
+            "savefig_format": Or(
+                None,
+                str,
+                error="Invalid value in config for plotting.savefig_format" "must be a value supported by Matplotlib.",
+            ),
+            "savefig_dpi": Or(
+                None,
+                "figure",
+                int,
```
MaxGamill-Sheffield, line 241: To follow as the same as below (and is a tad more specific)
```suggestion
                lambda n: n > 0,
```

docs/configuration.md
```diff
@@ -54,55 +54,58 @@ above:
...
+|                 | `save_format`                     | string         | `null`                      | Format to save images in, `null` defaults to `png` see [matplotlib.pyplot.savefig](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html)                                                                                                                                                                      |
+|                 | `savefig_dpi`                     | string / float | `null`                      | Dots Per Inch (DPI), if `null` then the value `figure` is used, for other values (typically integers) see [#further-customisation] and [Matplotlib](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html).                                                                                                    |
```
MaxGamill-Sheffield, line 97: Suggestion to mention this affects all images
```suggestion
|                 | `savefig_dpi`                     | string / float | `null`                      | Dots Per Inch (DPI), if `null` then the value `figure` is used, for other values (typically integers) see [#further-customisation] and [Matplotlib](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html). Low DPI's improve processing time but can reduce the plotted trace (but not the actual trace) accuracy.         |
```

MaxGamill-Sheffield (changes requested):
From my testing, I see that the defaults are 'nulls' for the `configuration` field output to the terminal which could be confusing.  Could we move the references to `mpl.rcParams` from `plottingfuncs.py` into `run_topostats.py` or the `utils.update_plotting_config()` so that the dictionaries are updated earlier, and thus the defaults are reflected in terminal output?
Screenshot 2024-01-22 at 10 26 31

Additionally, none of the matplotlib colourmaps are available due to the validation restriction. I propose we swap this to just `str`, make it clear in the config file of any additional ones we've added + mpl ones, and let mpl deal with the error when it hits a plot . I'll add these suggestions during the code review. The alternative is to label all mpl cmaps here and keep up-to-date with any additional ones they add which may be tedious.

Also I think we should mention in the config file that the savfig_dpi affects all figures and thus having them too low may affect the traces seen on the dnatracing images (maybe one for after the cats branch merge though).

Other than that it all seems to work well, thanks Neil.

ns-rse:
Thanks for the feedback @MaxGamil-Sheffield

I hadn't thought about `None` being listed in the `completion_message()` for DPI/Output Format/cmap and appreciate this
would be confusing so thanks for highlighting that. The solution I've gone for (removing the additions that reported
these) is different from that suggested (update the `config` dictionary with parameters from `mpl.rcParams` early in
processing). My reasoning being...

+ Previously we didn't report these, no one has ever asked to see them in the logging output.
+ We write configuration options to YAML file via `write_yaml()` at the end of processing. Its a verbatim copy of that
  which was used (either user specified or `default_config.yaml`)and it contains the settings used. If a user didn't
  specify DPI/cmap/format I'm not sure we should alter this. It could be argued it is useful to provide them but then
  that would also require writing _all_ other configuration/plotting options to be consistent should the default values
  ever change in the future.

Currently only a handful of parameters are read from `topostats.mplstyle` and this is conditional on
whether any of these are being over-ridden or not when instantiating the `Images()` class. Currently we

```
plottingfuncs.Images() > load_mplstyle() > Set DPI/cmap/format based on arguments to Images()
```

It is certainly possible to change this process as suggested and..

```
load_mplstyle() > Update DPI/cmap/format > plottingfuncs.Images()
```

...but that is a larger amount of work to undertake and introduces scope drift to this PR. If it is desirable to report this information in logging and/or ensure the configuration file that is written contains the default parameters from `topostats.mplstyle` then we can address that as a separate issue.

One challenge there might be is that as TopoStats evolves the configuration files will become dated (e.g. new options may be added, some may be removed, defaults might change). That said specific versions of TopoStats should always work with an associated configuration file, so perhaps we need to give this more consideration and include TopoStats versions in configs too to ensure results are reproducible. :thinking: :exploding_head:

MaxGamill-Sheffield (changes requested):
Looks good thanks Neil! 

Just a suggestion to the docs and 2 other things I've just forgotten while writing this. Lmk what you think

topostats/validation.py
```diff
@@ -229,7 +230,17 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
...
+            "savefig_format": Or(
+                None,
+                str,
+                error="Invalid value in config for plotting.savefig_format" "must be a value supported by Matplotlib.",
+            ),
+            "savefig_dpi": Or(
+                None,
+                "figure",
+                lambda n: n > 0,
+                error="Invalid value in config for plotting.savefig_dpi, valid" "values are 'figure' or integers",
```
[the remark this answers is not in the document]
ns-rse: Checked [docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html) gone with `float` rather than `integers or floats.`

docs/configuration.md
```diff
@@ -165,28 +168,133 @@ through the basics.
 
 ### Further customisation
 
-Whilst the broad overall look of images is controlled in this manner there is one additional file that controls how
+Whilst the overall look of images is controlled in this manner there is one additional file that controls how
 images are plotted in terms of filenames, titles and image types and whether an image is part of the `core` subset that
 are always generated or not.
```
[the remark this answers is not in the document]
ns-rse: Also added details of what the core set is to the relevant row of the configuration options table.

topostats/validation.py
```diff
@@ -271,6 +282,12 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 [None, And(int, lambda n: n > 1)],
                 error="Invalid value in config plotting.for 'num_ticks', valid values are 'null' or integers > 1",
             ),
+            "cmap": Or(
+                None,
+                str,
+                error="Invalid value in config for 'plotting.cmap', valid values are 'afmhot', 'nanoscope', "
+                "'gwyddion' or values supported by Matplotlib",
```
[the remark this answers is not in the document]
ns-rse: Somewhat hesitant to start putting links in error messages as the documentation already links to this (see `configuration.md` at lines 104 and 242 of this PR). Its duplication putting everything that is in the documentation into the code too and I'm inclined to assume[^1] that if someone is specifying a Matplotlib colormap then they will likely already be familiar with those colormaps and/or have already consulted the documentation.

[^1]: "_To assume makes an ass out of u and me_" :zany_face:

MaxGamill-Sheffield (approved):
Good to go! And sorry I didn't see the Newsletter comment until just now.

Issue #787, opened 2024-01-29
Pytest >= 8.0.0 breaks tests


ns-rse:
Issues with the `pytest-lazy-fixture` plugin have cropped up with the release of [pytest-8.0.0](https://docs.pytest.org/en/stable/changelog.html) on 2024-01-27 (see [issue](https://github.com/TvoroG/pytest-lazy-fixture/issues/65)).

There are some solutions on lazy fixture alternatives in that thread and may be possible to [`request.getfixturevalue()`](https://ns-rse.github.io/posts/pytest-param/#request.getfixturevalue) instead.

In the interim suggest blocking new `pytest` versions by modifying `pyproject.toml` to...

```toml
  "pytest<8.0.0",
```

...until there is time to address this properly.

PR #788, opened 2024-01-29
pytest<8.0.0

ns-rse:
Temporary fix for #787


SylviaWhittle (approved):
Provided the tests, pass this is good, thank you

PR #801, opened 2024-02-12
Added 'pm' and 'mm' for very small and large scans

ns-rse:
As #796 seeks to address an issue with non-square images that were 1 pixel wide it makes sense to have such scenarios
tested.

+ The `pytest.param()` function is used to add meaningful identifiers to each of the tests.
+ Adds a `1x512` pixel scan to the parameters (denoted `pn units; rectangular (thin)`).
+ Conversely adds a `512x1` pixel scan to the parameters (denoted `pm units; rectangular (tall)`)

I'm unsure...

+ whether the `tall` test makes sense but then I'm also don't know why there would be an image that is only 1 pixel wide.
+ why the `expected_px2nm` is always based on the x-axis size and whether this would be an issue?

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -481,10 +481,12 @@ def test_gwy_read_component(load_scan_dummy: LoadScans) -> None:
 @pytest.mark.parametrize(
     ("unit", "x", "y", "expected_px2nm"),
     [
-        ("mm", 0.01, 0.01, 10000),
-        ("um", 1.5, 1.5, 1500),
-        ("nm", 50, 50, 50),
-        ("pm", 233, 233, 0.233),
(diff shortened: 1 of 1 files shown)


MaxGamill-Sheffield:
The error and associated fix in #796 were only discovered by the 1x512px image, not the cause of it. The microscope thought it foolish to record the 0.87nm long direction in 'nm' but instead 'pm'., not because of the non-squareness of the image, and thus the 'pm' test has already been added. Apologies if this wasn't clear.

As the function currently requires each pixel to be square (the indexing of 0 in `io.py` ~ line 530), modifying these values currently does nothing. I'll merge and add the tests as they should be useful for the non-square pixel work that is currently being worked on and to avoid writing these tests again (only modifying them) at a later date 👍

PR #790, opened 2024-01-31
Save grain trace data to HDF5 format

SylviaWhittle:
Closes #488 

This PR replaces #695. Instead of saving the trace data to JSON, we have decided to save it to the `.topostats` file in HDF5 format.

Experimentalists have requested that the grain trace data be accessible and readable in plain text format if possible, but we will add functions to make it easier to work with later on, adding a JSON dump function if absolutely necessary.

PR #804, opened 2024-02-20
Remove saving of gaussian filtered arrays to .npy files

### LINKED ISSUE #802: Stop generating at .npy when running Topostats

A `.npy` file is generated for each processed image along side an image and an HDF5 `.topostats` file. This is no longer necessary since the image data is all head in the `.topostats` file and the `.npy` file just takes up extra space.

Stop the generation of the `.npy` file.

Could also be made optional in the config file.

### Additional context

_No response_

ns-rse:
Closes #802

Arrays and data are now saved in HDF5 files (#790) and so `.npy` arrays are somewhat redundant. This PR leaves the `io.save_array()` function in place should interactive use require saving of arrays but removes its use from `topostats.processing.run_filters()` so that the `.npy` files are no longer saved to disk.

Users wishing to access processed data should load it from the HDF5 formatted `.topostats` files that are saved during processing.

### DIFF
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -12,7 +12,7 @@ from topostats import __version__
 from topostats.filters import Filters
 from topostats.grains import Grains
 from topostats.grainstats import GrainStats
-from topostats.io import get_out_path, save_array, save_topostats_file
+from topostats.io import get_out_path, save_topostats_file
 from topostats.logs.logs import LOGGER_NAME, setup_logger
 from topostats.plottingfuncs import Images, add_pixel_to_nm_to_plotting_config
 from topostats.statistics import image_statistics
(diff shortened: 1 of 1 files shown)


MaxGamill-Sheffield (approved):
Looks good! Will approve :)

Just a small best practice question though - is it typical for a library to have it's own save function like this even though it's basically a wrapper for another library's function (numpy)?

ns-rse:
Thanks.

> Just a small best practice question though - is it typical for a library to have it's own save function like this even though it's basically a wrapper for another library's function (numpy)?

Depends, by making it a wrapper you can add more functionality to a small atomic function (`save_array()`) and keep it focused on doing one thing and one thing alone, then update the call you make to it from elsewhere in the code base. For example if there is a labelled image but you wanted to convert to a mask for all objects (ignoring the fact we already have these available!) then you could add an option to the `def save_array(..., mask: bool = False)` and have it call whatever function is used for masking. Then in a place during processing where you do want to mask before saving you would add `bool = True` and be done. Of course you could do the masking as a step in the processing before calling `save_array()` but over time the steps in processing would grow and grow (its already pretty long) and so my preference is to add options in this manner.

This is pretty much what has been done with all the refactoring, breaking long chunks of code into smaller functions with options that are then called. It might seem faff when there is no immediate benefit but further down the line it makes it easier to extend functionality I think.

PR #808, opened 2024-02-28
Fixing DPI settings

### LINKED ISSUE #807: DPI settings from config override the plotting dictionary
The value was added to `topostats/default_config.yaml` to allow users to over-ride the defaults.

Ideally we want...

`if plotting["savefig_dpi"] is None`... use the values in `topostats/plotting_dictionary.yaml`.

`else:` ... use the values from `topostats/default_config.yaml`/command line.

ns-rse:
Closes #807

Current process is...

1. `topostats/entry_point.create_parser` : get arguments from commandline via the `process_parser` subparser (includes DPI)
2. `topostats/run_topostats.run_topostats()` : loads the dictionary and updates it with any values from `arg`, will update `savefig_dpi` with `--savefig-dpi` value.
3. Validate the updated configuration.
4. Load the default `topostats/plotting_dictionary.yaml` to the main `config["plotting"]["plot_dict"]"`. This adds in all the image specific settings such as title and importantly DPI values for each image.
5. Validate the `config["plotting"]` configuration.
6. Update the plotting configuration with via `utils.update_plotting_config()`. This adds all the options _not_ defined in the `plotting_dictionary` with those from `topostats/default_config.yaml`.

This was wrong because we have `plotting_config["plot_dict"][image] = {**options, **main_config}` and so the options from `**main_config` override those from `**options` as you can not have dictionaries with duplicate keys.

This has been fixed by...

1. Updating the values from `plot_dictionary` with any that exist in the `default_config.yaml` under the `plotting` section. Importantly this means that if any are `None` they do not update, logic that already existed in `update_config()`.
2. Then adding to each image within `plot_dictionary` any values from the `plotting` section of `default_config.yaml` that do _not_ already exist.

This avoids replacing the defaults from `plot_dictionary` with `None` values from the `plotting` section of `default_config.yaml`.

Tests included which check equality  of dictionaries and that `LOGGER.info()` is included when changes are made to the DPI.

Other Changes

+ `LOGGER.info()` to `LOGGER.debug()` in `topostats.io.dict_to_hdf5` as I don't think printing the keys is that informative.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,4 +1,4 @@
-"""Test utils."""
+"""Test the utils module."""
 
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


MaxGamill-Sheffield (approved):
Looks good, and tests well! 

I've just added a commit to change the "Options: .." bit in the config file so it no longer talks about "figure".

P.S. thanks for changing the .topostats logger setting too, it was driving me up the wall hahaha

PR #810, opened 2024-03-04
Tidies up Images.plot_and_save()

### LINKED ISSUE #806: Add @pytest.mark.mpl_image_compare() fixture to test_plottingfuncs.py::test_plot_and_save_no_axes_no_colorbar
Noticed this test doesn't compare an image, only the sum of the elements and its shape after reading it back in which I'm unsure whether it demonstrates that the image has been saved without axes or colorbar. Would therefore be useful to add a fixture that uses the pytest-matplotlib extension to save and compare images as is done with others.

```python
@pytest.mark.mpl_image_compare(baseline_dir="resources/img/")
def test_plot_and_save_no_axes_no_colorbar(load_scan_data: LoadScans, plotting_config: dict, tmp_path: Path) -> None:
    """Test plotting without axes and without the colourbar."""
    plotting_config["axes"] = False
    plotting_config["colorbar"] = False
    fig, _ = Images(
        data=load_scan_data.image,
        output_dir=tmp_path,
        filename="01-raw_heightmap",
        title="Raw Height",
        **plotting_config,
    ).plot_and_save()
    return fig
```

ns-rse:
Closes #806

Whilst #806 was meant to address why no Matplotlib image was generated and tested/compared to a target image when plotting without axes or colorbar I discovered this was because the image is saved with `Images.save_array_figure()`.

This in turn uses [`matplotlib.pyplot.imsave()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imsave.html) to save the image and the test was using [`skimage.io.imread()`](https://scikit-image.org/docs/stable/api/skimage.io.html#skimage.io.imread) to read the image back as an array and test the `np.sum()` and `img.shape`. The other method available is `Images.save_fig()` which in turn calls Matplotlibs `matplot.pylig.savefig()`.

We no longer need to save arrays nor images as readable arrays (see #804 / #802) as the arrays are saved in the HDF5 `.topostats` files and so the need to save the image in this manner seemed redundant. The `Images.save_figure()` already has logic which excludes the axes and scale bars (see lines 321 of current changeset) although these have been further tweaked to use `tight_layout`.

Further the logic for deciding what to save within `Image.plot_and_save()` seemed overly complicated and if `Images.save_array_figure()` were being called `Images.plot_and_save()` did not return a `fig` (Matplotlib Image) that could be saved and tested using `pytest-mpl` extension, another reason for switching and in turn removal.

To this end I have...

1. Removed `Images.save_array_figure()` if people need the arrays they can use the HDF5 files).
2. Tweaked the plotting options under `Images.savefig()` starting at line 321 to use a [tight
 layout](https://matplotlib.org/2.0.2/users/tight_layout_guide.html) which ensures there is no border (see note below).
3. Simplified the logic in `Images.plot_and_save()` controlling whether images are saved so that `ValueError` are raised if `Images.savefig == False` or `if Images.image_set in ["all", "core"] or self.core_set:` evaluates to `False`. Ensures images are explicitly closed to reduce memory usage.
4. Added tests for the raising of `ValueError` exceptions in `Images.plot_and_save()`
5. Tidied up some test names to be consistent (resulted in image name change).
6. Tidied up passing of `plotting_config` into these tests.
7. Added a `pytest-mpl` test for `test_plot_and_save_no_axes_no_colorbar()`, adding the target image as required by the issue that started this.
8. Added an explanation of what `image_set` and `core_set` are to the `Images()` class definition as this confused me having not touched plotting in a number of months, hopefully useful for others who look at this code.

Note on Borders

The image generated by `pytest-mpl` against which comparisons are made _does_ have a border because it is saving the returned `fig` object itself. The actual generated image doesn't have a border (manual checks have been made).

image

**NB** It also appears that an old image was included in the repository for this test using the full `minicircle` image rather than the cropped area that @SylviaWhittle slimmed the test suite down to in order to speed up tests.

An unexpected consequence of this change was that some target directories were not being created. I couldn't see why as I hadn't touched the code that did this. In fixing it I found that some other tests broke where images that were meant to be in the `all` `image_set` were apparently being created when the `image_set == "core"`. See specific commit message for more details.


MaxGamill-Sheffield:
Hey @ns-rse before I review this I just want to check with the experimentalists that this is the desired intent of the tool and aligns with their goals as users. I'll mention in the group meeting tomorrow.

It is my belief that people will want to use the images generated by TS that appear in the directories, and having an output that saved the image pixel-for-pixel (without interpolation and dpi limitations) is useful for publications / presentations (and has helped me save on storage), all without having to run more code to save individual arrays from the `.topostats` files which I don't think is available yet without prior coding knowledge.

Regardless, the test logic and fixes using the cropped region, and the updated info in the doc strings are helpful too so I'll get back to you on this :)

topostats/processing.py
```diff
@@ -79,17 +79,21 @@ def run_filters(
...
                 if plot_name not in ["scan_raw"]:
                     if plot_name == "extracted_channel":
                         array = np.flipud(array.pixels)
-                    plotting_config["plot_dict"][plot_name]["output_dir"] = filter_out_path
+                    plotting_config["plot_dict"][plot_name]["output_dir"] = (
+                        core_out_path if plotting_config["plot_dict"][plot_name]["core_set"] else filter_out_path
+                    )
                     try:
                         Images(array, **plotting_config["plot_dict"][plot_name]).plot_and_save()
                         Images(array, **plotting_config["plot_dict"][plot_name]).plot_histogram_and_save()
```
MaxGamill-Sheffield, line 99: The only other thing is that you've got an `if image_set=='all'` statement which is then used again within the `plot_and_save()` function, and I'm thinking would it make sense just to use it the once in processing?

We could pop it and re-add it like the `plotting.run` param and keep all the non-core images within this `if` statement, and the core image outside like it is? What do you think?

MaxGamill-Sheffield (changes requested):

topostats/processing.py
ns-rse: The logic here though is to create the sub-directory `filter` as held in `filter_out_path`. The same is true for `grains`. We then need to set the parameter in the current images `plot_dict` to whatever that is, hence the `if/else`.

When I simplified the logic within `plot_and_save()` for some reason I couldn't understand I got error messages about target directories not being present so I figured I needed to create them.

ns-rse: Also having `plot_and_save()` know about `image_set` might be useful should anyone use the class/method interactively e.g. in a Notebook.

MaxGamill-Sheffield: Ok makes sense, thanks for explaining :)

MaxGamill-Sheffield (approved):

Issue #805, opened 2024-02-20
[feature] : Remove smallest_grain_size_nm2 as a parameter from config (superfluous)


SylviaWhittle:
### Is your feature request related to a problem?

The `smallest_grain_size_nm2` parameter of the grains section of the config is superfluous, as `absolute_area_threshold` covers this and has understandably recently caused confusion within a group of users. 

It previously hadn't been removed since it was doing no harm, however we now have reason to action this given the aforementioned confusion.

### Describe the solution you would like.

Remove `smallest_grain_size_nm2` and allow the area thresholding step to take on the responsibility of removing very small objects from images, as it would already do if the noise wasn't already removed by the remove noise function.

### Describe the alternatives you have considered.

Keeping it in would likely cause continued confusion in the user-base.

ns-rse:
Remove from the following files...

+ `topostats/default_dict.yaml`
+ `topostats/validation.py`
+ `docs/configuration.md`
+ `topostats/grains.py`

The main work required is in `topostats/grains.py` where `smallest_grain_size_nm2` is used in a few places...

## `Grains.remove_noise()`

Defined on line [151](https://github.com/AFM-SPM/TopoStats/blob/8798977820ddbf165c258a27ac5ba533f6decbad/topostats/grains.py#L151) this method uses the value directly as an argument to `skimage.morphology.remove_small_objects()`.

The method is only ever defined and is not used anywhere so can be safely removed.

## Lower threshold for area thresholding.

Line [324](https://github.com/AFM-SPM/TopoStats/blob/8798977820ddbf165c258a27ac5ba533f6decbad/topostats/grains.py#L324)

```python
            self.directions[direction]["removed_noise"] = self.area_thresholding(
                self.directions[direction]["tidied_border"],
                [self.smallest_grain_size_nm2, None],
            )
```

I propose replacing the lower area threshold (`self.smallest_grain_size_nm2`) with `self.minimum_grain_size / (self.pixel_to_nm_scaling) ** 2`.

ns-rse:
Removed in 0c602ea1e8894c60f1e3c502037e13262a9570ea which was merged via #1123

PR #816, opened 2024-03-05
Adds numpydoc-validation hook to pre-commit

ns-rse:
Adds the [numpydoc validation hook](https://numpydoc.readthedocs.io/en/latest/validation.html) to `.pre-commit-config.yaml` and configures via `pyproject.toml`.

`conftest` and `test_` functions are excluded from checks of docstrings.

For details of the validation checks see [Numpydoc validation | Built-in validation
checks](https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks).

I've gone through all files under `topostats/` and corrected all validation errors.

Each file has its own commit to make it clear, succinct and easy to identify where the changes came from (this is known as "atomic commits").

There were a few places where more changes were made but a lot of it was just starting new lines, putting spaces around semi-colons.

However, because this is mostly just formatting/linting these commits have been added to `.git-blame-ignore-revs` so that the "blame" for these changes does not get attributed to me but remains with the original author until they are changed.

I've taken the liberty to also address many of the typehints with a view to one day sitting down and applying [`mypy`](https://www.mypy-lang.org/) to the code base and introducing that as a `pre-commit` hook too (maybe one day we could be even stricter and use [`pydantic`](https://docs.pydantic.dev/latest/)!).

It also mean that subsequent commits will have their doc strings validated too but this is only a good thing, it improves the quality of the documentation by ensuring there are considerably less errors.

EDIT : I missed the files under `topostats/tracing/` out. Whilst I've sorted out `dnatracing.py` / `dnacurvature.py` / `skeletonize.py` I have _not_ completed adding docstrings to `tracingfuncs.py` as it is so poorly documented doing so is not adding any additional information at this point, the docstrings added so far are mostly meaningless and do not describe what the functions are doing. Given the Zhang method is implemented elsewhere (skimage) and its not clear how this deviates from that, plus that there is on-going work on this front anyway the file has been excluded from `numpydoc-validation` (as has `docs/conf.py` which is a Sphinx file for building documentation and not part of our code base).


topostats/filters.py
```diff
@@ -231,20 +267,45 @@ def remove_nonlinear_polynomial(self, image: np.ndarray, mask: np.ndarray | None
...
-        image: np.ndarray
-            2D numpy heightmap array of floats with a polynomial trend to remove.
-        mask: np.ndarray
-            2D numpy boolean array used to mask out any points in the image that are deemed not to be part of the
-            heightmap's background data. This argument is optional.
+        image : npt.NDArray
+            2-D numpy height-map array of floats with a polynomial trend to remove.
+        mask : npt.NDArray, optional
+            2-D Numpy boolean array used to mask any points in the image that are deemed not to be part of the
+            heightmap's background data.
```
llwiggins, line 274: Really minor comment but this was changed to "height-map" above so could also do the same here to ensure consistency.

topostats/filters.py
```diff
@@ -373,19 +461,21 @@ def average_background(self, image: np.ndarray, mask: np.ndarray = None) -> np.n
...
+        **kwargs
+            Keyword arguments passed on to the skimage.filters.gaussian() function.
 
         Returns
         -------
-        np.array
-            Numpy array of gaussian blurred image.
-
+        npt.NDArray
+            Keyword arguments passed on to the skimage.filters.gaussian() function.
```
llwiggins, line 478: This doesn't seem to describe the correct return value, as I think it has been copied from the **kwargs definition. This should instead be a numpy array that represents the image following Gaussian filtering.

topostats/grains.py
```diff
@@ -148,20 +198,23 @@ def calc_minimum_grain_size(self, image: np.ndarray) -> float:
         else:
             self.minimum_grain_size = -1
 
-    def remove_noise(self, image: np.ndarray, **kwargs) -> np.ndarray:
-        """Remove noise which are objects smaller than the 'smallest_grain_size_nm2'.
+    def remove_noise(self, image: npt.NDArray, **kwargs) -> npt.NDArray:
+        """
+        Remove noise which are objects smaller than the 'smallest_grain_size_nm2'.
```
llwiggins, line 203: Would it be better to write something like "Removes noise, where noise is defined as objects smaller than the 'smallest_grain_size_nm2"?

topostats/plotting.py
```diff
@@ -45,53 +87,51 @@ def __init__(
...
-        var_to_label: dict
+        var_to_label :  dict
             Variable to label dictionary for automatically adding titles to plots.
-        hue: str
+        hue :  str
             Dataframe column to group plots by.
-
-        Returns
-        =======
         """
```
llwiggins, line 135: No "Returns" heading provided here.

ns-rse: That is what I've removed :wink:

Typically old is in red and new is green.

You might find it easier (I do) to not use the "Unified diff view" on pull requests where things are shown line by line and instead use the "Split diff view" which shows the side by side. There should be a little book symbol right above the very first file in this PR which toggles this, see also [Diff view options](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#diff-view-options).

I also use the [Refined GitHub](https://github.com/refined-github/refined-github) browser plugin which seamlessly improves the GitHub User Experience. Many of their features have already been incorporated into GitHub (and are no longer part of the extension).

topostats/filters.py
ns-rse: Good spot thanks.

You can always use the "_Suggest Changes_" button to the comment box (see item 7 of [Reviewing proposed changes in a pull request - GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request)).

Can be quicker to just make a quick edit and submit it as a suggestion.

topostats/grains.py
ns-rse: I was pretty lazy on this as the `smallest_grain_size_nm2` is due to be removed and the `remove_noise()` function isn't actually called anywhere at present (I started work on removing this the other day too see #805 )

PR #823, opened 2024-04-15
refactor(grainstats): GrainStats.get_max_min_feret() > measure.feret

ns-rse:
Closes #798

Replaces the functionality for calculating minimum and maximum feret distances within the `GrainStats`  class with the new `topostats.measure.feret.min_max_feret()`.

+ Reorder dictionary returned by `min_max_feret()`.
+ Dictionary is included in the `stats` dictionary built by `GrainStats` and saved as CSV.
+ Updates `tests/test_grainstats_minicircle.py::test_grainstats_regression` to include the co-ordinates.
+ Updates `tests/test_grainstats.py::test_process_scan*` to include co-ordinates.
+ Removes `GrainStats.get_max_min_feret()` and associated methods that are called as well as their tests.
+ Co-ordinates for min/max feret are included by default in HDF5 output now that they are part of the returned dataframe.

Will have to rethink/change how results are returned by `GraintStats` class if height profiles are to be extracted. Probably most sensible to have profiling as a separate step conducted after GrainStats.

PR #824, opened 2024-04-17
test(grainstats): Removes min/max feret co-ordinates from output

ns-rse:
Further to #823 this reverts the GrainStats output to its previous form and removes the feret co-ordinates from the output as discussed in this mornings meeting.


MaxGamill-Sheffield (approved):
Yep looks good to go!

PR #825, opened 2024-04-17
chore: Linting dnatracing; setup skeletons for remaining tests; split pruning to submodule

ns-rse:
+ Linting docstrings of `dnatracing.py`.
+ Skeletons for remaining testing.
+ Pruning is moved to its own submodule to reduce the size of `skeletonize.py` which was failing `pylint`.


MaxGamill-Sheffield (approved):
Looks good and works as expected. Happy to be merged :)

PR #864, opened 2024-06-26
Bound aspect ratio between 0 and 1

### LINKED ISSUE #861: [feature] : Bound aspect ratio between 0 and 1

Aspect ratio currently bounded between 1 -> inf, which feels like it holds less meaning than 0 -> 1.

Change grainstats code:
```
# Enforce >= 1 aspect ratio
if aspect_ratio < 1.0:
    aspect_ratio = 1 / aspect_ratio
```
to
```
# Enforce <= 1 aspect ratio
if aspect_ratio > 1.0:
    aspect_ratio = 1 / aspect_ratio
```

_No response_

### Additional context

_No response_

llwiggins:
Resolves #861 

This PR consists of a small edit to code in `grainstats.py` to resolve issue #861 and bound the aspect ratio values between 0 and 1.


topostats/grainstats.py
```diff
@@ -834,8 +834,8 @@ def calculate_aspect_ratio(
                     (extremes["x_max"] - extremes["x_min"]),
                     (extremes["y_max"] - extremes["y_min"]),
                 )
-                # Enforce >= 1 aspect ratio
-                if aspect_ratio < 1.0:
+                # Enforce <= 1 aspect ratio
```
ns-rse, line 837: I think we don't need this, instead we can remove line `828` where `aspect_ratio` is calculated and instead calculated `aspect_ratio` as...

```suggestion
                aspect_ratio = smallest_bounding_width / smallest_bounding_length
```

ns-rse (changes requested):
As discussed I think there is a smarter way of ensuring the ratio is always <= 1, suggestion in line.

ns-rse:
Hi @llwiggins 

Looks like the regression tests will need updating too in light of these changes as three of the `above` boxes had ratios `> 1.0` and this obviously changes that.

The regression tests are generated using the [`pytest-regtest`](https://pytest-regtest.readthedocs.io/en/latest/) extension which writes output to files for comparison each time tests run.

The tests fail and exit on `tests/test_grainstats_minicircle.py` and so to update the regression test output you would...

```bash
pytest --regtest-reset tests/test_grainstats_minicircle.py
```

I think you will probably need to reset the `tests_processing.py` regression test output too...

```bash
pytest --regtest-reset tests/test_processing.py
```

Then the updated output (which resides under `tests/_regtest_outputs` will need staging and committing.

```bash
git commit -u
git commit -m "tests: Update regression tests for aspect_ratio < 1.0"
```

Give me a shout if you have any questions.

ns-rse:
@llwiggins I took the liberty of applying the simpler solution (and at the same time updating the regression tests) so we can get this tweak merged.

ns-rse (approved):

PR #874, opened 2024-07-16
Fixing broken tests (Part I)

ns-rse:
Work on getting the existing tests to pass in light of all the changes in `maxgamill-sheffield/800-better-tracing`, still a fair bit to go but small Pull Requests are considerably easier and faster to undertake.

## `tests/test_grainstats_minicircle.py`

+ reset regression tests output `molecule_number` > `grain_number` in `statistics` output.
+ Reset `minicircle_grainstats.cropped_size` as commit 53fd579 set the default `grainstats.cropped_size` to `-1` which
  broke the tests which were developed using a size of 40.0nm rather than bounding to the image (which is what `-1` does).

## `tests/tracing/test_dnatracing_multigrain.py`

+ `npt.NDArray` typehints.
+ `pytest.param(id="")` added.
+ `skeletonisation_method="topostats"` no longer a valid option to `dnaTrace()` class, so updated, in some places this uses the `**dnatracing_config` keyword approach.

## `tests/tracing/test_dnatracing_methods.py`

+ `pytest.param(id="")` added.
+ `grain` > `mask`
+ Removed `print()`

## `tests/tracing/test_dnatracing_single_grain.py`

+ `dnatrac.smoothed_grain` > `dnatrace.smoothed_mask`

## `topostats/tracing/dnatracing.py`

+ `grain` > `mask` in a number of places.


MaxGamill-Sheffield:
With merge of Sylvia's NN there is some issues, I'm not sure if it just needs a pull of her branch and changes?

MaxGamill-Sheffield:
Fixed above - mainly just conflicts between variable names and typehints in geometry.py in which Sylvias' were more explicit.
Changes in nodestats were lengthening abbreviated comments

MaxGamill-Sheffield:
This looks like a great start to keep things small. 

I was running the tests but realise they still fail because they have not been addressed here, just terminology and "id's" added which sound like a good idea for diagnosing why.

Happy to merge to the main 800 branch

PR #876, opened 2024-07-18
Separated Disordered Tracing into own module

MaxGamill-Sheffield:
This is the first part of the mega PR to refactor the 800-better-tracing branch.

This PR in particular does:
- Creates a smoke test for me to check the refactored workflow works (once it's finished)
- Adds to one of the tests to allow the comparison between `np.nan` and `np.nan` which previously failed
- Moves smoothing, skeletonisation and pruning workflow into it's own module. This works on the image but processes each grain separately, producing a dictionary of the grain's crop data (skeletonisation stages and bounding boxes). The individual crops are then compiled into a "full_image" to plot as the processing / diagnostic images. This differs from previously where the trace coordinates were stored / passed into dnatracing (this step will move into dnatracing as nodestats prefers skeletons)
- Moves the smoothing, skeletonisation and pruning parameters into it's own section in the config (and accommodates this in validation.py
- Adds the "disordered_trace" step in process_scan.
- Works along-side the current dnatracing pipeline and is NOT connected to anything on the output.

Some docstrings may not be right but I've ran (some) pre-commit checks so they at-least should be present (but I was going to check these together after the pipeline works as i/o's etc may change)

*when making changes to this, and other stacked branches, will need to rebase <branch_tip> --update-refs  and that should update 800-better-traces-refactor too


tests/test_nodestats_smoke.py
ns-rse, line 1: I'm not a great fan of the term "smoke" for regression tests. Its not very descriptive in comparison to "regression".

tests/test_nodestats_smoke.py
```diff
@@ -0,0 +1,59 @@
...
+from topostats.io import LoadScans, hdf5_to_dict, read_yaml
+from topostats.processing import (
+    process_scan,
+)
+
+BASE_DIR = Path.cwd()
+SMOKE = BASE_DIR / "tests/smoke"
+
+
+def test_smoke(regtest, tmp_path, load_scan_spm: LoadScans) -> None:
```
ns-rse, line 20: `test_smoke()` doesn't tell the reader what is being tested, its not smoke that is being tested.

```suggestion
def test_regression_dnatracing_with_nodetsats(regtest, tmp_path, load_scan_spm: LoadScans) -> None:
```

topostats/default_config.yaml
```diff
@@ -1,5 +1,5 @@
 base_dir: ./ # Directory in which to search for data files
-output_dir: ./ # Directory to output results to
```
ns-rse, line 2: Not sure why this is `./` its always been `./output/` on `main` but good to have it reverted. :+1:

topostats/plotting_dictionary.yaml
```diff
@@ -192,61 +192,62 @@ grain_mask_image:
...
+  mask_cmap: "blu"
   core_set: false
   savefig_dpi: 600
+# Nodestats troubleshooting images
 nodes:
   filename: "24-nodes"
   title: "Image with Skeletons and Highlighted Nodes"
   image_type: "non-binary"
-  mask_cmap: "green_black"
+  mask_cmap: "blu_purp_green"
```
ns-rse, line 227: I don't think its any great burden to avoid abbreviations and be explicit and suggest using `blue_purple_green`.

IDE's all support Tab-completion and its simple to refactor across a whole project with the language servers that sit behind them. If you use VSCode its [F2](https://code.visualstudio.com/docs/editor/refactoring#_rename-symbol).

tests/test_nodestats_smoke.py
MaxGamill-Sheffield: renamed all smoke references to regression

topostats/plotting_dictionary.yaml
MaxGamill-Sheffield: done!

MaxGamill-Sheffield:
nodestats tracing .topostats test fails in the comparison because previously the keys in the .topostats file for the new changes aren't saved out yet as the pipeline is not complete.

topostats/default_config.yaml
MaxGamill-Sheffield: reverted

ns-rse: This PR _does_ revert it :wink: 

If you want different output directories when developing/testing then you could use the command line options/flags or generate and customise a configuration file and invoke with that.

Although perhaps this doesn't fit with a Notebook based workflow, but if using notebooks load the config and change the value in the dictionary might be easiest, then you don't have to remember to revert changes on the feature branch. :shrug:

tests/smoke/minicircle.topostats.out
ns-rse, line 1: I think this file shouldn't be under `_regtest_outputs/` as this is a directory that `pytest-regtest` writes the output of the `print()` statement to.

Its also not clear how or why it differs from `tests/resources/processed/minicircle.topostats`.

I would expect there to be one target HDF5 file in the tests that are compared to what is produced by a full run. If there is need to have different versions then I would place them under `tests/resources/` with a name that makes it clear what aspect they relate to (in this instance `nodestats` so something like `tests/resources/minicircle_nodestats.topostats`).

tests/test_nodestats_regression.py
```diff
@@ -0,0 +1,57 @@
...
+from test_io import dict_almost_equal
+
+import topostats
+from topostats.io import LoadScans, hdf5_to_dict, read_yaml
+from topostats.processing import (
+    process_scan,
+)
+
+BASE_DIR = Path.cwd()
+REGRESSION = BASE_DIR / "tests/nodestats-regression"
```
ns-rse, line 17: In light of above comment on `minicircle.topostats.out` this should be...

```suggestion
RESOURCES = BASE_DIR / "tests/resources"
```

tests/test_nodestats_regression.py
```diff
@@ -0,0 +1,57 @@
...
+    results.drop(["basename"], axis=1, inplace=True)
+    print(img_stats.to_string(float_format="{:.4e}".format), file=regtest)  # noqa: T201
+    print(results.to_string(float_format="{:.4e}".format), file=regtest)  # noqa: T201
+
+    # Regtest for the topostats file
+    assert Path.exists(tmp_path / "tests/resources/processed/minicircle.topostats")
+    with h5py.File(tmp_path / "tests/resources/processed/minicircle.topostats", "r") as f:
+        saved_topostats = hdf5_to_dict(f, group_path="/")
+
+    with h5py.File(REGRESSION / "minicircle.topostats.out", "r") as f:
```
ns-rse, line 47:
```suggestion
    with h5py.File(RESOURCES / "minicircle_nodestats.topostats", "r") as f:
```

tests/resources/processed/minicircle.topostats
ns-rse, line 1: Not sure this file needs to be here :thinking: 

My understanding is that the `tests/resources/ministats.topostats` file is one that is compared to the output that is written to the `tmp_path/` directory when running TopoStats. I can't see that this is read in anywhere (`minicircle.topostats.out` is but I've suggested moving and renaming that to reflect the test it relates to).

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+from topostats.tracing.skeletonize import getSkeleton
+
+LOGGER = logging.getLogger(LOGGER_NAME)
+
+
+class disorderedTrace:
+    """
+    Calculate disordered traces for a DNA molecule and calculates statistics from those traces.
+
+    2023-06-09 : This class has undergone some refactoring so that it works with a single grain. The `trace_grain()`
```
ns-rse, line 26: This comment is more appropriate to commit messages. Users who are reading the API are unlikely to be interested in this level of detail. Developers who are would read it in the `git log`.

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+
+    def re_add_holes(
+        self, orig_mask: npt.NDArray, new_mask: npt.NDArray, holearea_min_max: list = (4, np.inf)
+    ) -> npt.NDArray:
+        """
+        Restore holes in masks that were occluded by dilation.
+
+        As Gaussian dilation smoothing methods can close holes in the original mask, this function obtains those holes
+        (based on the general background being the first due to padding) and adds them back into the smoothed mask. When
+        paired with smooth_mask, this essentially just smooths the outer edge of the mask.
```
ns-rse, line 169: Markup for inline code...

```suggestion
        paired with ``smooth_mask``, this essentially just smooths the outer edge of the mask.
```

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+        self.pruned_skeleton = self.remove_touching_edge(self.pruned_skeleton)
+        self.disordered_trace = np.argwhere(self.pruned_skeleton == 1)
+
+        if self.disordered_trace is None:
+            LOGGER.info(f"[{self.filename}] : Grain failed to Skeletonise")
+        elif len(self.disordered_trace) < self.min_skeleton_size:
+            self.disordered_trace = None
+
+    def re_add_holes(
+        self, orig_mask: npt.NDArray, new_mask: npt.NDArray, holearea_min_max: list = (4, np.inf)
```
ns-rse, line 162: I wonder if it would be more descriptive to use `smoothed_mask` here than `new_mask` as the method is applied to a smoother mask.

```suggestion
        self, orig_mask: npt.NDArray, smoothed_mask: npt.NDArray, holearea_min_max: list = (4, np.inf)
```

You should be able to easily change this across the code base using the [Rename symbol](https://code.visualstudio.com/docs/editor/refactoring#_rename-symbol) although in my experience it _doesn't_ make the necessary changes to the doc-string so you would have to manually update that field.

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+        # compare num holes in each mask
+        holey_smooth = new_mask.copy()
+        holey_smooth[holes == 1] = 0
+
+        return holey_smooth
+
+    @staticmethod
+    def remove_touching_edge(skeleton: npt.NDArray) -> npt.NDArray:
+        """
+        Remove any skeleton points touching the border (to prevent errors later).
```
ns-rse, line 206: What sort of errors does this throw?

It might be useful to use these as a basis for tests to ensure that this function does indeed remove them.

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+            Gaussian sigma value to smooth the grains after an Otsu threshold. If None, defaults to
+            max(grain.shape) / 256.
+
+        Returns
+        -------
+        npt.NDArray
+            Numpy array of smmoothed image.
+        """
+        gaussian_sigma = max(grain.shape) / 256 if gaussian_sigma is None else gaussian_sigma
+        print("-------", dilation_iterations, type(dilation_iterations))
```
ns-rse, line 248: We should remove all `print()` statements or convert to `LOGGER.info()` with informative messages as to the event being logged.

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+        dilation = ndimage.binary_dilation(grain, iterations=dilation_iterations).astype(np.int32)
+        gauss = gaussian(grain, sigma=gaussian_sigma)
+        gauss[gauss > threshold_otsu(gauss) * 1.3] = 1
+        gauss[gauss != 1] = 0
+        gauss = gauss.astype(np.int32)
+        # gauss
```
ns-rse, line 254: Its a little subtler than what this comment suggests since the Gaussian and Dilation smoothing has already been undertaken.

```suggestion
        # Now add hole to the smooth mask conditional on whether the difference in the dilation is greater than zero.
```

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+        if dilation.sum() - grain.sum() > gauss.sum() - grain.sum():
```
ns-rse, line 255: The utility of subtracting `grain.sum() here isn't clear to me, both dilations will have started from the same `grain` and one may be larger than the other.

Mathematically this inequality could be simplified to...

```suggestion
        if dilation.sum() > gauss.sum():
```

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+            bbox = bboxs[cropped_image_index]
+            full_image[bbox[0] : bbox[2], bbox[1] : bbox[3]] += crop[pad_width:-pad_width, pad_width:-pad_width]
+
+        disordered_trace_crop_data[f"grain_{cropped_image_index}"] = disordered_trace_images
+        disordered_trace_crop_data[f"grain_{cropped_image_index}"]["bbox"] = bboxs[cropped_image_index]
+
+    return disordered_trace_crop_data, all_images
+
+
+# not used
```
ns-rse, line 352: Is there a need to keep unused functions? They will add a burden to writing of tests so unless there is a need to keep them around I'd suggest removing them.

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+def round_splined_traces(splined_traces: dict) -> dict:
+    """
+    Round a Dict of floating point coordinates to integer floating point coordinates.
```
ns-rse, line 355:
```suggestion
    Round a dictionary of floating point coordinates to integer floating point coordinates.
```

...if it remains!

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+    Returns
+    -------
+    list(Tuple)
+        A list of tuples of the min_row, min_col of each bounding box.
+    """
+    bounding_coordinates = pad_bounding_box(array_shape, bounding_box, pad_width)
+    return (bounding_coordinates[0], bounding_coordinates[1])
+
+
+def trace_grain(
```
ns-rse, line 539: Was moving this intentional?

`trace_grain()` and related functions are something that I thought was to be done _after_ getting a disordered trace (by either of the two methods). I would have thought this should have remained in `dnatracing.py`. :thinking:

topostats/tracing/disordered_tracing.py
```diff
@@ -0,0 +1,708 @@
...
+
+    return cropped_images
+
+
+def crop_array(array: npt.NDArray, bounding_box: tuple, pad_width: int = 0) -> npt.NDArray:
+    """
+    Crop an array.
+
+    Ideally we pad the array that is being cropped so that we have heights outside of the grains bounding box. However,
+    in some cases, if an grain is near the edge of the image scan this results in requesting indexes outside of the
```
ns-rse, line 619:
```suggestion
    in some cases, if a grain is near the edge of the image scan this results in requesting indexes outside of the
```

ns-rse (changes requested):
First pass over the code, some suggestions some questions.

A lot of this looks as though its shifted code from `topostats/tracing/dnatracing.py` and we should be careful that we don't lose the `git blame` associated with the original comments made when writing them (this can be done using [`.git-blame-ignore-revs`](https://blog.nshephard.dev/posts/whos_to_blame/).

tests/test_nodestats_smoke.py
```diff
@@ -0,0 +1,59 @@
...
+    """Regression test for checking the process_scan functions correctly."""
+    load_scan_spm.get_data()
+    img_dic = load_scan_spm.img_dict
+    # Ensure there are below grains
+    process_scan_config = read_yaml(BASE_DIR / "topostats" / "default_config.yaml")
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Done

tests/smoke/minicircle.topostats.out
MaxGamill-Sheffield: Hey Neil, so this test was just to ensure the pipeline works after refactoring and as I did not want to edit the current tests, I created this standalone one.

It's not currently finished and to be permanent because the nodeStats `fwhm2()` func has an output of different length tuples so it can't be saved in the .topostats keys.

Would it be worth putting this into its own PR?

MaxGamill-Sheffield: Resolved by moving the test into it's own PR so issues with that can be addressed there.

tests/test_nodestats_regression.py
MaxGamill-Sheffield: Resolved by moving the test into it's own PR so issues with that can be addressed there.

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Was a remenant from dnatracing. Will remove.

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: The idea was to keep it as general as possible but if it's more readable this way then that's better!

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Didn't realise you could use markup in docstrings.

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: No idea as this is legacy tracing code and the error doesn't occur in this new module so we might have to test when the pipline is complete. My intuition tells me an indexError when trying to order a linear segment as the `get_node_count_and_area()` func attempts to find the 3x3 area around the node which would not be possible.

In addition, the above would only happend when the `pad_width` is 0, otherwise this function is kind of redundant (unless looking at large webbed structures but you'd usually want to keep these).

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Done

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Done

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: It was originally for playing with absolutes and deciding which smoothing route to take. But now changed

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Removed. I think it was an old diagnostic image when things wern't working and anyway it should be in splining. Good find!

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Another good spot! Removed

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: Yes as it's similar to how this was initially done in `dnatracing.py`, i.e. each grain is individually processed to get the disordered traces. Then the whole image is built up from the individual crops and this is the result. Then the next step will take the full pruned image and individually prcess each disordered trace to produce nodestats/ordered traces with it.

There was an initial thought to do this inside the loop of a single grain but we decided against this when we spoke in your office.

ns-rse (commented):
Pre-commit indicated there were still some linting errors with `topostats/tracing/disordered_tracing.py` (see
[log](openstreetmap.org) for full details of all issues).

I've addressed `RET504` in commit `ab8c7690` made directly on this branch.

In doing so `pylint` which is disabled in `pre-commit.ci` was run locally and threw the errors below which have also
been addressed in  `ab8c7690` and so are now fixed in this PR (mostly disabling the check but some line-wrapping, importing modules slightly different and
removing unused arguments)...

```
Pylint...................................................................Failed
- hook id: pylint
- exit code: 30

************* Module topostats.tracing.disordered_tracing
topostats/tracing/disordered_tracing.py:156:0: C0301: Line too long (138/120) (line-too-long) DONE
topostats/tracing/disordered_tracing.py:12:0: E0611: No name 'gaussian' in module 'skimage.filters' (no-name-in-module) DONE
topostats/tracing/disordered_tracing.py:12:0: E0611: No name 'threshold_otsu' in module 'skimage.filters' (no-name-in-module) DONE
topostats/tracing/disordered_tracing.py:22:0: R0902: Too many instance attributes (13/7) (too-many-instance-attributes) DISABLED
topostats/tracing/disordered_tracing.py:53:4: R0913: Too many arguments (11/7) (too-many-arguments) DISABLED
topostats/tracing/disordered_tracing.py:239:0: R0913: Too many arguments (10/7) (too-many-arguments) DISABLED
topostats/tracing/disordered_tracing.py:239:0: R0914: Too many local variables (26/15) (too-many-locals) DISABLED
topostats/tracing/disordered_tracing.py:327:15: W0718: Catching too general exception Exception (broad-exception-caught)
topostats/tracing/disordered_tracing.py:249:4: W0613: Unused argument 'cores' (unused-argument) DONE
topostats/tracing/disordered_tracing.py:402:0: R0913: Too many arguments (9/7) (too-many-arguments) DISABLED

-----------------------------------
Your code has been rated at 8.51/10
```

topostats/processing.py
```diff
@@ -628,6 +703,18 @@ def process_scan(
             grain_out_path=grain_out_path,
         )
 
+        # Disordered Tracing
+        disordered_traces = run_disorderedTrace(
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Ahhhh nightmare, this has been done but I assume in the nodestats PR because I wasn't yat the time sure how the two would flow together.

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: This has been renamed to "disordered_trace_grain()"

MaxGamill-Sheffield:
@ns-rse Thank you for amending the pre-commit complaints, some of these I don't think can be helped and when I tried to add the ignoring statements it didn't like them but that's probably a user error :)

The general exception was from older code and might be useful to look into when writing the tests?

topostats/tracing/disordered_tracing.py
ns-rse: Yep, docstrings are written in [reStructured Text](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html).

MaxGamill-Sheffield:
@ns-rse regarding the git-blame fixes, I've read your blog and it seems my commits were not atomic enough - the commit [Moved disordered traces out from dnatracing](https://github.com/AFM-SPM/TopoStats/pull/876/commits/d6d86c068daaf42a50467b2f67d9733aee9f9a12) includes:
• The disordered tracing functions moved out from dnatracing into `disordered_tracing.py`
• The updates to the config + validation
• My own functions to run the disordered tracing pipeline
With my reason being to ensure the pipeline worked so I wasn't making a load of tiny fix commits

So my question is, if I do put this commit in the `.git-blame-ignore-revs` file, how will it handle the new lines of code in the `disordered_tracing.py` file that wasn't there before? As it's a new filename, git wont associate those moved functions with their counterparts in the dnatracing file so there will be no blame references right, does this still make it worth doing?

MaxGamill-Sheffield:
This has been added in #889 (sorry)

PR #877, opened 2024-07-18
Separates out the NodeStats statistics from dnatracing

MaxGamill-Sheffield:
This is the second PR of the big 800-better-tracing-refactor which seperates and creates an additional "nodestats" step to run the crossing analyses and add extra grain statistics to the grainstats file.

This PR builds on the first by:
- separating out the nodestats config parameters (and associated params in the validation)
- renames trace_image to dnatrace_image to avoid confusion (this one might be split between this PR and the last)
- Modifies disordered trace output to accomodate nodestats input
- Adds nodestats process_scan and associated diagnostic plots 
- Adds grain statistics to the grainstats csv file
- *Does not append to the .topostats file -> the tuple produced by the `fwhm2` function doesn't like to be hdf5'd because they are unequal lengths

Again this one has even less tidying so will keep as a draft

*when making changes to this, and other stacked branches, will need to `rebase <branch_tip> --update-refs ` and that should update 800-better-traces-refactor too


MaxGamill-Sheffield:
Messed up branch refs as this is from 800-btr, not 800-btr_disordered-tracing

PR #878, opened 2024-07-19
Refactor and test connect_extended_nodes_nearest

SylviaWhittle:
This PR refactors and tests `nodestats`'s `connect_extended_nodes_nearest` function, adding sub-functions in `measure/geometry.py` and are tested with a single test case. I plan on adding more test cases in the future but will need to converse with @MaxGamill-Sheffield and @llwiggins about these.

This PR adheres to `mypy` too with proper typing as far as I can tell. No warnings are thrown from `mypy` on my system.

I have also endeavoured to improve the documentation and understanding about variables, though in this effort, I have turned the variable names into sentences... 😅 I can change the names if people prefer, but personally, as code is read far more than it is written / edited, I find that these overly verbose names can help prevent misunderstanding.


SylviaWhittle:
Errors are thrown during pre-commit, but as far as I can tell, they are all not pertaining to the `connect_extended_nodes_nearest` function that I've been refactoring.

SylviaWhittle:
Here is the latest pre-commit checks
image

I think these are all outside of the function I have been refactoring:
- `topostats/tracing/nodestats.py` Lines 368 - 426
- `topostats/measure/geometry.py` All lines

topostats/tracing/nodestats.py
```diff
@@ -642,11 +642,11 @@ def analyse_nodes(self, max_branch_length: float = 20e-9) -> None:
                     for i, angle in enumerate(angles):
                         matched_branches[i]["angles"] = angle
 
-                    """
-                    except ValueError:
-                        LOGGER.error(f"Node {node_no} too complex, see images for details.")
-                        error = True
-                    """
+                    # """
```
ns-rse, line 645: Could just remove this if its not used?

SylviaWhittle:
I think I've broken a lot of stuff - trying to rebase onto Max's `800-better-tracing` but it's pulling over lots of code I didn't add. I'll fix it somehow 👍

ns-rse:
If there are a number of commits that have been merged since you branched then you might want to use the `rerere` option when rebasing to avoid having to resolve the same error multiple times.

SylviaWhittle:
PR obsolete, code added

PR #879, opened 2024-07-19
Re-add changes for connect_nodes_nearest

SylviaWhittle:
Let's try this again 😅


SylviaWhittle:
image

pre-commit checks

tests/measure/conftest.py
```diff
@@ -0,0 +1,87 @@
...
+import pytest
+from skimage.morphology import label
+
+from topostats.tracing.nodestats import nodeStats
+
+# pylint: disable=redefined-outer-name
+
+
+@pytest.fixture()
+def network_array_representation_figure_8():
```
ns-rse, line 13: Very minor but could add...

```suggestion
def network_array_representation_figure_8() -> npt.NDArray:
```

..assuming that `import numpy.typing as npt` is added to the imports.

topostats/tracing/nodestats.py
```diff
@@ -3,17 +3,21 @@
 from __future__ import annotations
 
 import logging
-import math
 
 import networkx as nx
 import numpy as np
 import numpy.typing as npt
+from numpy.typing import NDArray
```
ns-rse, line 10: Essentially redundant given line 9.

I'm gradually favouring and getting in the habit of importing at the module level as it makes it easier to know what package modules belong to. In this instance not much of an issue but there may be other cases where there is the same function from different packages and by importing at the module level it makes it easier to distinguish.

A Find and Replace across the file for `NDArray` > `npt.NDArray` and removing this line should solve things.

SylviaWhittle:
Code now runs properly with `topostats process` using a generated default config on `minicircle.spm`

topostats/tracing/nodestats.py
```diff
@@ -140,6 +143,7 @@ def get_node_stats(self) -> tuple:
         LOGGER.info(f"Node Stats - Processing Grain: {self.n_grain}")
         self.conv_skelly = convolve_skeleton(self.skeleton)
         if len(self.conv_skelly[self.conv_skelly == 3]) != 0:  # check if any nodes
+            LOGGER.info(f"[{self.filename}] : Nodestats - {self.n_grain} contains crossings.")
```
ns-rse, line 146: :+1: Great to see more `LOGGER.info()` across everything.

topostats/tracing/nodestats.py
```diff
@@ -682,11 +634,11 @@ def analyse_nodes(self, max_branch_length: float = 20e-9) -> None:
                     for i, angle in enumerate(angles):
                         matched_branches[i]["angles"] = angle
 
-                    """
-                    except ValueError:
-                        LOGGER.error(f"Node {node_no} too complex, see images for details.")
-                        error = True
-                    """
+                    # """
```
ns-rse, line 637: Could this be removed?

topostats/tracing/nodestats.py
```diff
@@ -1053,7 +1008,7 @@ def gaussian(x: npt.NDArray, h: float, mean: float, sigma: float):
 
     def fwhm2(self, heights: npt.NDArray, distances: npt.NDArray, hm: float | None = None) -> tuple:
         """
-        Caculate the FWHM value.
+        Calculate the FWHM value. TODO: Dictionary-ify this one to help saving.
```
ns-rse, line 1011: Recommend capturing this in an issue so we can keep track of it (if not already done).

SylviaWhittle: That's a remnant I accidentally left in from Max's notes. removing it.

topostats/tracing/nodestats.py
SylviaWhittle: Done now, AFAIK I've found all of them.

topostats/tracing/nodestats.py
SylviaWhittle: 👍

PR #885, opened 2024-08-12
tests(ordered_tracing): Tests and fixes for ordered_trace_mask()

ns-rse:
The function `ordered_trace_mask()` looked strange on review...

- Docstring said it took `npt.NDArray` as argument but logic only returned mask if `list` was supplied.
- Slicing caused `TypeError: list indices must be integers or slices, not tuple`.
- After correcting that `np.arange(len(mol_coords))` raised a `ValueError: setting an array element with a sequence.`

Test written to demonstrate these problems and alternative `trace_mask()` function offered (along with tests that pass).

### DIFF
--- a/topostats/tracing/ordered_tracing.py
+++ b/topostats/tracing/ordered_tracing.py
@@ -16,7 +16,10 @@ from topostats.utils import convolve_skeleton, coords_2_img
 LOGGER = logging.getLogger(LOGGER_NAME)
 
 
-class OrderedTraceNodestats:
+# pylint: disable=too-many-locals
+
+
+class OrderedTraceNodestats:  # pylint: disable=too-many-instance-attributes
     """
     Order single pixel thick skeleton coordinates via NodeStats results.
 
@@ -77,7 +80,7 @@ class OrderedTraceNodestats:
 
         self.ordered_coordinates = []
 
-    def compile_trace(self) -> tuple:
+    def compile_trace(self) -> tuple:  # noqa: C901
         """
         Pipeline to obtain the trace and crossing trace image.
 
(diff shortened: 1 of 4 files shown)


ns-rse:
Noticed there are still pre-commit errors raised in CI on files I've not touched in this PR that should be addressed before the target branch is merged into `main`. Details [here](https://results.pre-commit.ci/run/github/168173540/1723479526.C8HXLAODRaabdNXZyGKrBQ) for reference.

ns-rse:
It is worth noting though that there is currently a `pylint` error for `unused-variable` pertaining to the returned
`graint_trace_data`...

```
************* Module topostats.processing
topostats/processing.py:999:20: W0612: Unused variable 'grain_trace_data' (unused-variable)
```

...which will need dealing with.

topostats/tracing/ordered_tracing.py
```diff
@@ -637,12 +640,39 @@ def ordered_trace_mask(ordered_coordinates: npt.NDArray, shape: tuple) -> npt.ND
...
+    npt.NDArray
+        NxM image with each pixel in the ordered trace labeled in ascending order.
+    """
+    mask = np.zeros(shape)
+    if isinstance(coordinates, np.ndarray):
+        mask[coordinates[:, 0], coordinates[:, 1]] = 1
+        return mask
+    if isinstance(coordinates, list):
+        for mol_coords in coordinates:
+            mask[mol_coords[0], mol_coords[1]] = 1
```
MaxGamill-Sheffield, line 666: Same here (and also this fails upon parsing a list of 2D coordinate arrays)

```suggestion
        mask[mol_coords[:, 0], mol_coords[:, 1]] = np.arange(len(coordinates))
```

MaxGamill-Sheffield (changes requested):
Hey Neil, great work thanks for looking at this! As in the comments, I'm happy for your function to be used instead of `ordered_trace_mask` (although I think it should keep this name).

However, the reason for handing the function lists was to parse a list of different length Nx2 coordinate arrays, and for these arrays to be plotted in ascending integer values to see the direction of tracing, and so I've suggested some changes to your function (but not the tests). 

Let me know if that makes sense :)

Binary mask:
Screenshot 2024-08-13 at 17 09 47
Ordered trace mask:
Screenshot 2024-08-13 at 17 11 27

PR #886, opened 2024-08-14
Seperate out Splining from dnaTracing

MaxGamill-Sheffield:
This PR splits out the dnatrace smoothing functions into the "splining.py" module to TopoStats which does the sole function of smoothing the disordered traces either as TopoStats normally does (average of splines), or Sylvia's rolling window method.

The fitted traces step has been removed as this could cause issues.
No new diagnostic images are produced - as this only works of the coordinates.
1 new core image is added -> The individual splined coordinates mapped onto the image.
(Currently) 2 new stats are added to grainstats -> total_contour_length and average_end_to_end_distance.
Saves molecule smoothed traces, bbox and contour length and e2e distances to the .topostats file.

### DIFF
--- a/topostats/tracing/disordered_tracing.py
+++ b/topostats/tracing/disordered_tracing.py
@@ -336,8 +336,8 @@ def trace_image_disordered(  # pylint: disable=too-many-arguments,too-many-local
             grainstats_additions[cropped_image_index] = {
                 "image": filename,
                 "grain_number": cropped_image_index,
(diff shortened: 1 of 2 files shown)


MaxGamill-Sheffield:
FYI @SylviaWhittle spent a day rebasing her analyse_nodes() (nodestats function) from this branch by accident instead of `800-btr-nodestats` so the work will remain here but make better us of her time.

PR #889, opened 2024-08-16
Maxgamill sheffield/800 btr splining

MaxGamill-Sheffield:


llwiggins:
Doing some detective work and noticed some unwanted behaviour in the way that pruned_skeletons are compiled from crops within `disordered_tracing/trace_image_disordered()` , see below 3x pruned skeletons for grains (1), (2) and (3) but the little nib within the pruned_trace mask for grain (2) is added to the final compiled mask, making it look as though grain (3) has a spurious branch (images below).

I think this is because that little fragment of grain (3) falls within the bounding box for grain (2)!

image
image
image
image

PR #890, opened 2024-08-21
Add U-Net grain segmentation option

SylviaWhittle:
This PR adds barebones functionality for re-segmenting grains detected in the `find_grains()` function using a pre-trained U-Net provided by the user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ topostats/_version.py
 
 # default output directory, often common from testing
 output/
+
+# Debugging
+pytest-debug.ini
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@
 
 import importlib.resources as pkg_resources
 from pathlib import Path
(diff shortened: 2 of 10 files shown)


SylviaWhittle:
Tests pass, will update docs now and tidy up then ready for review

llwiggins:
@SylviaWhittle I'm running into errors with grain finding when `threshold_method: unet` in the config. As there are no unet options in `get_thresholds()`, this causes problems when `self.thresholds[direction]` is called on line 401 in grains.py! 

Does it make sense to have this as a threshold value that then gets treated the same as otsu, std_dev and absolute - or to have this as a separate key/value in the config so that it doesn't go through the same functions as threshold values do?

llwiggins:
Could be better to make the default `grain_crop_padding` in `default_config.yaml` > 0 as currently this causes a long tail on molecules that may be tricky to get to the bottom off if not familiar with the code. See the top left molecule below as an example of this. 
Mask
Trace

PR #892, opened 2024-08-23
Add trace smoothing tests

SylviaWhittle:
Adds tests for the linear and circular trace smoothing methods

Also fix a couple bugs in the linear smoothing function


tests/test_splining.py
ns-rse, line 1: I can see the different methods (`pool_trace_circular()` and `pool_trace_linear()`) are tested but the `pixel_trace` for both are identical in that its always a 4x4 box.

Do we need to have a parameter set that is a truly linear?

Threes sides would give a linear molecule...

```
np.array([[0, 0], [0, 1], [0, 2], [0, 3], [1, 3], [2, 3], [3, 3], [3, 2], [3, 1], [3, 0]]).astype(np.int32)
```

topostats/tracing/splining.py
```diff
@@ -355,15 +357,15 @@ def pool_trace_linear(
         pooled_trace = [pixel_trace[0]]  # Add first coord as to not cut it off
 
         # Get average point for trace in rolling window
-        for i in range(1, len(pixel_trace) - 1):
+        for i in range(0, len(pixel_trace)):
```
ns-rse, line 360: :+1:

ns-rse (commented):
Just one question as to whether we should test actual linear pixel traces.

I'd kind of expect the two methods being tested to return slightly different traced splines which they do for the circular examples but it feels like there should be test cases/parameters that reflect the stated purpose of the methods.

tests/test_splining.py
SylviaWhittle: > I can see the different methods (`pool_trace_circular()` and `pool_trace_linear()`) are tested but the `pixel_trace` for both are identical in that its always a 4x4 box.
> 
> Do we need to have a parameter set that is a truly linear?
> 
> Threes sides would give a linear molecule...
> 
> ```
> np.array([[0, 0], [0, 1], [0, 2], [0, 3], [1, 3], [2, 3], [3, 3], [3, 2], [3, 1], [3, 0]]).astype(np.int32)
> ```

(Could very well be wrong about this, I might be misinterpreting your question)

So these two traces are both the same yes, but the tests are for how the function treats the trace. For example you could treat three (say [0, 0], [0, 1], [1, 1]) points as either:

Linear
Where there are two lines, starting at [0, 0], then to [0, 1], then to [1, 1]

Or

Circular
Where there are three lines, starting at [0, 0], then to [0, 1], then to [1, 1], then back to [0, 0]

The code for the circular pooling allows connecting the first and last points, whereas the linear one does not.

You would expect that if it was a circular trace, then the origin point would be repeated, but actually if this is the case, then that presents an issue with the pooling, as the point would be double-counted and therefore skew the average in that region.

IMG_4867

All this is to prevent (for linear) or allow (for circular) the averaging to loop around from the end of the trace to the start of the trace.
IMG_4867

ns-rse (changes requested):
Thanks for taking the time to explain the tests @SylviaWhittle makes sense and I did some plotting to clarify things in my mind.

Couple of minor suggestions, one to remove duplication of code and one to keep things tidy but will be ready to go after that.

ns-rse (approved):
Hope you don't mind but I took the liberty of moving the definition of `pixel_trace` under each set of parameters to be done once.

I'd say this is therefore good to go. :+1:

PR #893, opened 2024-08-31
Add multi-class deep learning segmentation support

SylviaWhittle:
This makes some fairly major edits to how TopoStats works, and merely introduces a new standard in a half-hearted manner.

Namely, that masks are no longer NxN binary / integer mask matrices, but NxNxC tensors, where C is the dimension for the class (channel). I have set a standard where C >= 2 even for binary masks, as it simplifies things in the downstream logic to have a boolean background mask, allowing many later functions to not have repeated checks such as:

```python
if len(mask.shape) == 2:
    do_single_channel_logic(mask)
elif len(mask.shape) == 3:
    do_single_channel_logic(mask[:, :, preferred_channel])
```

This naturally would force an overhaul of how processing modules such as `grainstats.py` and `dnatracing.py` would work, probably iterating over each class and making a bit of a mess of the output data frames.

So I have been lazy and simply hardcoded the class passed in to the `grainstats` and `dnatracing` sections of the code, to be class index `1`. This allows the processing to run exactly as before, but also allows any additional downstream pipelines to access multiple channel masks.


SylviaWhittle:
I have hardcoded the value `1`, and not added it into config as I am not currently entirely sure this is the right way to go, so want to discuss these changes before putting time into finalising things and adding config options.

SylviaWhittle:
I am so sorry to everyone in future that has to dig through 3D tensors to update regtests

SylviaWhittle:
Tests pass and pre-commit isn't angry, so I think it's ready?

The code sadly isn't type safe because of the dictionaries used as arguments but I think it would take too much effort & extra code to cement types properly.

llwiggins:
I kept running into an `ValueError: attempt to get argmax of an empty sequence` within `keep_largest_labelled_region()` in grains.py as sometimes a mask channel can have no labelled regions, for example if a grain is detected where only DNA is present and no protein.

I added a check and handle for cases when the mask is empty and the code nows runs through to completion. Here's my fix:

```
    @staticmethod
    def keep_largest_labelled_region(
            labelled_image: npt.NDArray[np.int32],
        ) -> npt.NDArray[np.bool_]:
        """
        Keep only the largest region in a labelled image.

        Parameters
        ----------
        labelled_image : npt.NDArray
            2-D Numpy array of labelled regions.

        Returns
        -------
        npt.NDArray
            2-D Numpy boolean array of labelled regions with only the largest region.
        """
        # Check if there are any labelled regions
        if labelled_image.max() == 0:
            return np.zeros_like(labelled_image, dtype=bool)

        # Get the sizes of the regions
        sizes = np.array([(labelled_image == label).sum() for label in range(1, labelled_image.max() + 1)])
        
        # Keep only the largest region
        return np.where(labelled_image == sizes.argmax() + 1, labelled_image, 0).astype(bool)
```

SylviaWhittle:
> I kept running into an `ValueError: attempt to get argmax of an empty sequence` within `keep_largest_labelled_region()` in grains.py as sometimes a mask channel can have no labelled regions, for example if a grain is detected where only DNA is present and no protein.
> 
> I added a check and handle for cases when the mask is empty and the code now runs through to completion. Here's my fix:
> 
> ```
>     @staticmethod
>     def keep_largest_labelled_region(
>             labelled_image: npt.NDArray[np.int32],
>         ) -> npt.NDArray[np.bool_]:
>         """
>         Keep only the largest region in a labelled image.
> 
>         Parameters
>         ----------
>         labelled_image : npt.NDArray
>             2-D Numpy array of labelled regions.
> 
>         Returns
>         -------
>         npt.NDArray
>             2-D Numpy boolean array of labelled regions with only the largest region.
>         """
>         # Check if there are any labelled regions
>         if labelled_image.max() == 0:
>             return np.zeros_like(labelled_image, dtype=bool)
> 
>         # Get the sizes of the regions
>         sizes = np.array([(labelled_image == label).sum() for label in range(1, labelled_image.max() + 1)])
>         
>         # Keep only the largest region
>         return np.where(labelled_image == sizes.argmax() + 1, labelled_image, 0).astype(bool)
> ```

Thank you very much, that's great. If I had made better tests this would have been identified... I'll see about adding a test case for no grains too to trigger this.

SylviaWhittle:
Thanks @llwiggins, fixed and added test cases.

llwiggins (approved):
Thank you @SylviaWhittle for your work on this PR! All worked well during my testing and I was able to modify which channel is accessed from the tensor to analyse DNA/protein separately.

DNA is assumed to be class 1 here and so the 1st channel is accessed via hard-coded 
`[:, :, 1] ` within processing.py. I think this is fine for now and we can look to generalise this and make it more configurable in future PRs. Happy for this to be pushed to main as is!

PR #894, opened 2024-09-02
Fix height profile tests for macOS

SylviaWhittle:
Previously, the tests `test_process_scan_above_height_profiles` and `test_process_scan_below_height_profiles` needed to be skipped on macOS (presumably arm64).

I think I have a simple solution for this, however it's entirely possible I am missing something and that pytest-regtest is required for this test.

My solution uses the `dict_almost_equal` method to directly compare the result and expected dictionaries with a tolerance of 0.01 nanometres.


llwiggins (approved):
Thanks @SylviaWhittle, I tried running `tests/test_processing.py` before/after your changes and your code enables the tests to run successfully without needing to be skipped. Good spot and nice fix, I'm happy for this to be merged!

Issue #887, opened 2024-08-15
[feature] : message in terminal that says which image causes an error when processing multiple files at the same time


TFirth2:
### Is your feature request related to a problem?

Currently, to find an image or images that cause an error in TopoStats you have to run 1 core which stops on the image that causes the error and this is inefficient. If there are multiple error causing images that can be very time consuming.

### Describe the solution you would like.

 It would be good to run the entire folder and then get a message saying that X image or X image caused the error so these can be dealt with more effectively.

### Describe the alternatives you have considered.

I currently batch process image with grain stats off to just view the images and manually remove images that are cut off, blanks or otherwise not good. I then re-run TopoStats for grain stats but then often get stuck on images that won't process to give grain stats.

ns-rse:
Thanks for highlighting this @TFirth2 

I can fully appreciate how much of a pain this is.

Its like this because of [technical debt](https://en.wikipedia.org/wiki/Technical_debt) where we were rushing to get something working quickly and implemented a broad and crude solution and haven't gone back and fixed it yet.

When we write code sometimes errors arise, but we want to handle those errors. In Python that can be done with the `try: ... except: ...` construct where some code is tried and if an exception (also known as an error) arises it does what is defined under the `except:` clause.

We have a broad `try: ... except: ...` around grain finding (see lines [166-184](https://github.com/AFM-SPM/TopoStats/blob/70d02bd811285efcb6c72bdef854b148f414b951/topostats/processing.py#L161) if interested) although I notice its currently commented out on the `main` branch which is possibly a hangover from someone debugging (currently checking with @SylviaWhittle ).

The challenge is knowing/anticipating what the possible errors that can occur might be and with only `minicircle.spm` to work with on developing tests there wasn't much scope for knowing what might crop up.

If you, or anyone else in the lab, have some sample images which cause errors it would be really useful for the developers to have them so we can see what they cause and write finer-grained code to capture these and, hopefully, allow processing of other images to continue.

ns-rse:
Has the reduction in logging information addressed this I wonder?

Are there batches of files we could test with please?

Feels like it might be possible to close this out for the release.

ns-rse:
@TFirth2 We think this is now handled as desired as we have reduced the amount of logging information output and report warnings when specific images fail.

Are you able to test against the current `main` branch on a set of files with at least one (ideally a few) that are known to fail and a few that are known to pass please?

[...]

PR #898, opened 2024-09-11
Adds topological features into better tracing

MaxGamill-Sheffield:
Adds:
- branch idx and associated image and funcs
- topological classifications using the Topoly library
- writhe classification and associated functions
*possibly more which I'll check and update this with

### DIFF
--- a/topostats/plotting_dictionary.yaml
+++ b/topostats/plotting_dictionary.yaml
@@ -125,18 +125,21 @@ mask_grains:
   filename: "17-mask_grains"
   title: "Mask for Grains"
   image_type: "binary"
+  mask_cmap: "binary"
   savefig_dpi: 100
   core_set: false
 labelled_regions_01:
   filename: "18-labelled_regions"
   title: "Labelled Regions"
   image_type: "binary"
+  mask_cmap: "rainbow"
   savefig_dpi: 100
   core_set: false
 tidied_border:
   filename: "19-tidy_borders"
   title: "Tidied Borders"
   image_type: "binary"
+  mask_cmap: "rainbow"
   savefig_dpi: 100
   core_set: false
 removed_noise:
@@ -178,6 +181,7 @@ coloured_boxes:
   filename: "25-labelled_image_bboxes"
   title: "Labelled Image with Bounding Boxes"
   image_type: "binary"
+  mask_cmap: "rainbow"
   savefig_dpi: 100
   core_set: false
 grain_image:
--- a/topostats/plotting_dictionary.yaml
+++ b/topostats/plotting_dictionary.yaml
@@ -218,23 +218,30 @@ pruned_skeleton:
   mask_cmap: "blue"
   core_set: false
   savefig_dpi: 600
+branch_indexes:
+  filename: "23-segment_indexes"
+  title: "Skeleton Segment Indexes"
+  image_type: "non-binary"
+  mask_cmap: "viridis"
+  core_set: false
+  savefig_dpi: 600
 branch_types:
-  filename: "23-branch_types"
-  title: "Skeleton Branch Types"
+  filename: "24-segment_types"
+  title: "Skeleton Segment Types"
   image_type: "non-binary"
   mask_cmap: "viridis"
   core_set: false
   savefig_dpi: 600
 # Nodestats troubleshooting images
 convolved_skeletons:
-  filename: "24-convolved_skeleton"
+  filename: "25-convolved_skeleton"
   title: "Skeletons and Nodes"
   image_type: "non-binary"
   mask_cmap: "blue_purple_green"
   core_set: false
   savefig_dpi: 600
 node_centres:
-  filename: "25-node_centres"
+  filename: "26-node_centres"
   title: "Skeletons and Highlighted Nodes"
   image_type: "non-binary"
   mask_cmap: "blue_purple_green"
@@ -276,21 +283,21 @@ ordered_traces:
   core_set: false
   savefig_dpi: 600
 trace_segments:
-  filename: "26-trace_segments"
+  filename: "27-trace_segments"
   title: "Trace Segments"
   image_type: "non-binary"
   mask_cmap: "gist_rainbow"
   savefig_dpi: 600
   core_set: false
 over_under:
-  filename: "27-molecule_crossings"
+  filename: "28-molecule_crossings"
   title: "Visualised Molecule Crossings"
   image_type: "non-binary"
   mask_cmap: "blue_purple_green"
   core_set: false
   savefig_dpi: 600
 all_molecules:
-  filename: "28-all_molecules"
+  filename: "29-all_molecules"
   title: "Individual Molecules"
   image_type: "non-binary"
   mask_cmap: "blue_purple_green"
@@ -298,7 +305,7 @@ all_molecules:
   core_set: false
 # Splining
 fitted_trace:
-  filename: "27-fitted-traces"
+  filename: "30-fitted-traces"
   title: "Fitted Trace"
   image_type: "non-binary"
   mask_cmap: "blue_purple_green"
--- a/topostats/tracing/disordered_tracing.py
+++ b/topostats/tracing/disordered_tracing.py
@@ -315,6 +315,7 @@ def trace_image_disordered(  # pylint: disable=too-many-arguments,too-many-local
         "smoothed_grain": img_base.copy(),
         "skeleton": img_base.copy(),
         "pruned_skeleton": img_base.copy(),
+        "branch_indexes": img_base.copy(),
         "branch_types": img_base.copy(),
     }
 
@@ -526,11 +527,12 @@ def disordered_trace_grain(  # pylint: disable=too-many-arguments
         "smoothed_grain": disorderedtrace.smoothed_mask,
         "skeleton": disorderedtrace.skeleton,
         "pruned_skeleton": disorderedtrace.pruned_skeleton,
-        "branch_types": get_branch_type_image(cropped_image, disorderedtrace.pruned_skeleton),
+        "branch_types": get_skan_image(cropped_image, disorderedtrace.pruned_skeleton, "branch-type"),
+        "branch_indexes": get_skan_image(cropped_image, disorderedtrace.pruned_skeleton, "node-id-src"),
     }
 
 
-def get_branch_type_image(original_image: npt.NDArray, pruned_skeleton: npt.NDArray) -> npt.NDArray:
+def get_skan_image(original_image: npt.NDArray, pruned_skeleton: npt.NDArray, skan_column: str) -> npt.NDArray:
     """
     Label each branch with it's Skan branch type label.
 
@@ -552,16 +554,18 @@ def get_branch_type_image(original_image: npt.NDArray, pruned_skeleton: npt.NDAr
     npt.NDArray
         2D array where the background is 0, and skeleton branches label as their Skan branch type.
     """
-    branch_type_image = np.zeros_like(original_image)
+    branch_field_image = np.zeros_like(original_image)
     skeleton_image = np.where(pruned_skeleton == 1, original_image, 0)
     skan_skeleton = skan.Skeleton(skeleton_image, spacing=1e-9, value_is_height=True)
     res = skan.summarize(skan_skeleton)
 
-    for i, branch_type in enumerate(res["branch-type"]):
+    for i, branch_field in enumerate(res[skan_column]):
         path_coords = skan_skeleton.path_coordinates(i)
-        branch_type_image[path_coords[:, 0], path_coords[:, 1]] = branch_type + 1
+        if skan_column == "node-id-src":
+            branch_field = i
+        branch_field_image[path_coords[:, 0], path_coords[:, 1]] = branch_field + 1
 
-    return branch_type_image
+    return branch_field_image
 
 
 def crop_array(array: npt.NDArray, bounding_box: tuple, pad_width: int = 0) -> npt.NDArray:
--- a/topostats/tracing/disordered_tracing.py
+++ b/topostats/tracing/disordered_tracing.py
@@ -128,7 +128,9 @@ class disorderedTrace:  # pylint: disable=too-many-instance-attributes
             method=self.skeletonisation_params["method"],
             height_bias=self.skeletonisation_params["height_bias"],
         ).get_skeleton()
-        self.pruned_skeleton = prune_skeleton(self.image, self.skeleton, self.pixel_to_nm_scaling, **self.pruning_params.copy())
+        self.pruned_skeleton = prune_skeleton(
+            self.image, self.skeleton, self.pixel_to_nm_scaling, **self.pruning_params.copy()
+        )
         self.pruned_skeleton = self.remove_touching_edge(self.pruned_skeleton)
         self.disordered_trace = np.argwhere(self.pruned_skeleton == 1)
 
--- a/topostats/tracing/disordered_tracing.py
+++ b/topostats/tracing/disordered_tracing.py
@@ -338,29 +338,13 @@ def trace_image_disordered(  # pylint: disable=too-many-arguments,too-many-local
             LOGGER.info(f"[{filename}] : Disordered Traced grain {cropped_image_index + 1} of {n_grains}")
 
             # obtain segment stats
-            res = skan.summarize(
-                skan.Skeleton(
-                    np.where(disordered_trace_images["pruned_skeleton"] == 1, cropped_image, 0),
-                    spacing=pixel_to_nm_scaling,
-                )
+            skan_skeleton = skan.Skeleton(
+                np.where(disordered_trace_images["pruned_skeleton"] == 1, cropped_image, 0),
+                spacing=pixel_to_nm_scaling,
             )
-            res["image"] = filename
-            res["grain_number"] = cropped_image_index
-            disordered_tracing_stats = pd.concat(
-                (
-                    disordered_tracing_stats,
-                    res[
-                        [
-                            "image",
-                            "grain_number",
-                            "branch-distance",
-                            "branch-type",
-                            "mean-pixel-value",
-                            "stdev-pixel-value",
-                        ]
-                    ],
-                )
-            )
+            skan_df = skan.summarize(skan_skeleton)
+            skan_df = compile_skan_stats(skan_df, skan_skeleton, cropped_image, filename, cropped_image_index)
+            disordered_tracing_stats = pd.concat((disordered_tracing_stats, skan_df))
 
             # obtain stats
             conv_pruned_skeleton = convolve_skeleton(disordered_trace_images["pruned_skeleton"])
@@ -369,7 +353,7 @@ def trace_image_disordered(  # pylint: disable=too-many-arguments,too-many-local
                 "grain_number": cropped_image_index,
                 "grain_endpoints": any(conv_pruned_skeleton[conv_pruned_skeleton == 2]),
                 "grain_crossings": any(conv_pruned_skeleton[conv_pruned_skeleton == 3]),
-                "total_branch_lengths": res["branch-distance"].sum(),
+                "total_branch_lengths": skan_df["branch-distance"].sum(),
             }
 
             # remap the cropped images back onto the original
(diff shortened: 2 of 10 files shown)


ns-rse (commented):
It might be worth merging the #897 first so that the `maxgamill-sheffield/topology` branch contains the tests and changes before merging into `maxgamill-sheffield/800-better-tracing`.

`pre-commit` fails on files that haven't been touched in this change set.

## `docs/configuration.md`

```
Linting: 14 file(s)
Summary: 7 error(s)
docs/configuration.md:89:97 MD033/no-inline-html Inline HTML [Element: br]
docs/configuration.md:89:124 MD033/no-inline-html Inline HTML [Element: br]
docs/configuration.md:90:97 MD033/no-inline-html Inline HTML [Element: br]
docs/configuration.md:91:97 MD033/no-inline-html Inline HTML [Element: br]
docs/configuration.md:91:116 MD033/no-inline-html Inline HTML [Element: br]
docs/configuration.md:91:143 MD033/no-inline-html Inline HTML [Element: br]
docs/configuration.md:91:166 MD033/no-inline-html Inline HTML [Element: br]
```

I think these changes come from [bd333b4](https://github.com/AFM-SPM/TopoStats/commit/bd333b4cdb205a2552a1cf3386cd71079e4415c9) and the desire to show what the default values for dictionaries are it looks like there is a conflict between `markdownlint-cli2` and `prettier`.

Two solutions as far as I can tell...

1) Split the dictionary over multiple lines with blank cells in rows.
2) Add the `br` element to the list of `allowed_elements` in `.markdownlint-cli2.yaml`.

```yaml
config:
  ...
  html:
    allowed_elements:
      - div
      - br
```

## `ruff`

The linting with `ruff` should be straight-forward to resolve.

```
tests/test_processing.py:285: expected a comma-separated list of codes (e.g., `# noqa: F401, F841`).
```
Copy & pasta error, need to use the code rather than the `pylint` error name.

```
tests/test_processing.py:509:39: F821 Undefined name `run_dnatracing`
```

The `processing.run_dnatracing()` function has been removed for some reason which is why this is complaining (can't yet find the commit). Will need an equivalent for the refactored code.

```
tests/tracing/test_dnatracing_multigrain.py:174:9: F821 Undefined name `trace_image`
```

Perhaps renamed to `dnatrace_image()` ? Not found the commit that changed that yet.

```
tests/tracing/test_dnatracing_multigrain.py:425:15: F821 Undefined name `trace_image`
```

As above.

```
topostats/tracing/nodestats.py:1485:9: C901 `average_height_trace` is too complex (11 > 10)
Found 4 errors.
```
Could disable this check with

```
# noqa: C901
```

## `numpydoc`

These too should be straight-forward to resolve.

```
+---------------------------+------------------------------------------+---------+----------------------------------------------------+
| file                      | item                                     | check   | description                                        |
+===========================+==========================================+=========+====================================================+
| topostats/plotting.py:480 | plotting.plot_crossing_linetrace_halfmax | GL01    | Docstring text (summary) should start in the line  |
|                           |                                          |         | immediately after the opening quotes (not in the   |
|                           |                                          |         | same line, or leaving a blank line in between)     |
+---------------------------+------------------------------------------+---------+----------------------------------------------------+
| topostats/plotting.py:480 | plotting.plot_crossing_linetrace_halfmax | SS05    | Summary must start with infinitive verb, not third |
|                           |                                          |         | person (e.g. use "Generate" instead of             |
|                           |                                          |         | "Generates")                                       |
+---------------------------+------------------------------------------+---------+----------------------------------------------------+
| topostats/plotting.py:480 | plotting.plot_crossing_linetrace_halfmax | PR01    | Parameters {'branch_stats_dict', 'mask_cmap',      |
|                           |                                          |         | 'title'} not documented                            |
+---------------------------+------------------------------------------+---------+----------------------------------------------------+
| topostats/plotting.py:480 | plotting.plot_crossing_linetrace_halfmax | RT01    | No Returns section found                           |
+---------------------------+------------------------------------------+---------+----------------------------------------------------+
```
These come from two commits

+ [Add crossing linetrace plotting func](https://github.com/AFM-SPM/TopoStats/commit/344ebe42800558bf04333ca061eea586b85b0040)
+ [fwhm2 -&gt; fwhm and dictionary-ified it to input into .topostats](https://github.com/AFM-SPM/TopoStats/commit/8bd499ccab56a06200707d7c63f77a4679cca8ba)

I think the following might solve these (can't make suggestion as not in this PR)...

```
    """
    Plot the heightmap lines traces of the branches found in the 'branch_stats' dictionary, and their meetings.

    Parameters
    ----------
    branch_stats_dict: dict
        Dictionary containing branch height, distance and fwhm info.
    mask_cmap: matplotlib.colors.Colormap
        Colormap for plotting.
    title: str
        Title for the plot.

    Returns
    -------
    fig, ax
       Matplotlib fig and ax objects.
    """

**Comment by ns-rse** [comment]
I've addressed most of the above in a separate branch which I will merge into `maxgamill-sheffield/800-better-tracing`.  The renaming/refactoring of `dnatracing.trace_image()` will take a while though so will be addressed as a separate task.

**Comment by ns-rse** [comment]
> I've addressed most of the above in a separate branch which I will merge into maxgamill-sheffield/800-better-tracing. The renaming/refactoring of dnatracing.trace_image() will take a while though so will be addressed as a separate task.

Further work required this is more involved, see #899

**Comment by MaxGamill-Sheffield** [comment]
Hey @ns-rse , just seen your comments and want to double check what is required for this to merge:
1. Add Sylvias testing suite from #897 
2. Resolve pre-commit errors: (Done in #900 ?)
- markdown <br>'s
- ruff complaints
- plot_crossing_linetrace_halfmax docstrings
3. Fix merge conflicts (which I will do now)

**Comment by ns-rse** [comment]
1. Not sure if you need to merge those into this although there may be issues if you've changed things and it breaks the tests @SylviaWhittle has written.
2. Yes I've addressed those.
3. Sorry for the merge conflicts, it's almost unavoidable when three people are working on the same areas of code.

Feel free to go with more meaningful names, I just didn't feel `grainstats_additions_df` gave much information as to what was being added.

File: topostats/tracing/pruning.py [hunk 1]
```diff
@@ -11,13 +11,13 @@
...
 from topostats.tracing.skeletonize import getSkeleton
-from topostats.tracing.tracingfuncs import genTracingFuncs
+from topostats.tracing.tracingfuncs import coord_dist, genTracingFuncs, order_branch
 from topostats.utils import convolve_skeleton
 
 LOGGER = logging.getLogger(LOGGER_NAME)
 
 
-def prune_skeleton(image: npt.NDArray, skeleton: npt.NDArray, **kwargs) -> npt.NDArray:
+def prune_skeleton(image: npt.NDArray, skeleton: npt.NDArray, px_2_nm: float, **kwargs) -> npt.NDArray:
```
ns-rse, line 20: I'd be inclined to keep the argument names the same as used elsewhere so its consistent across functions and there is no mental overhead in equating `px_2_nm` to `pixel_to_nm_scaling`. Variables are scoped correctly do we don't need to come up with endless variations to avoid conflicts. There is also a definite advantage to keeping names consistent too as it can make using `kwargs` very easy if dictionaries contain settings.

```suggestion
def prune_skeleton(image: npt.NDArray, skeleton: npt.NDArray, pixel_to_nm_scaling: float, **kwargs) -> npt.NDArray:
```

**NB** - This is a good case where you can use the Language Server of VSCode to replace all instances. In my experience it works very well, the one thing it _doesn't_ do is update the renaming within docstrings so that has to be changed manually.

topostats/tracing/pruning.py
```diff
@@ -39,10 +41,10 @@ def prune_skeleton(image: npt.NDArray, skeleton: npt.NDArray, **kwargs) -> npt.N
     """
     if image.shape != skeleton.shape:
         raise AttributeError("Error image and skeleton are not the same size.")
-    return _prune_method(image, skeleton, **kwargs)
+    return _prune_method(image, skeleton, px_2_nm, **kwargs)
 
 
-def _prune_method(image: npt.NDArray, skeleton: npt.NDArray, **kwargs) -> Callable:
+def _prune_method(image: npt.NDArray, skeleton: npt.NDArray, px_2_nm: float, **kwargs) -> Callable:
```
ns-rse, line 47: Same as above comment (and anywhere else that `px_2_nm` has been used).

topostats/tracing/pruning.py
```diff
@@ -321,62 +334,29 @@ def _prune_by_length(  # pylint: disable=too-many-locals  # noqa: C901
...
-                    for x, y in branch_coordinates:
-                        single_skeleton[x, y] = 0
-
-            if n_branches == 0:
-                pruning = False
-
-        return single_skeleton
+        coordinates = np.argwhere(single_skeleton == 1).tolist()
+        # The branches are typically short so if a branch is longer than
+        #  0.15 * total points, its assumed to be part of the real data
```
ns-rse, line 339: We should try and make our functions/methods as flexible as possible and avoid hard-coding variables. The default can be set as the default argument to the option but it then makes it considerably easier in the future when someone decides they want to use a threshold of say 10% of total branch length.

This would require adding another argument such as `max_length_proportion: float = 0.15`.

topostats/tracing/splining.py
```diff
@@ -587,13 +591,14 @@ def splining_image(
...
                 molstats[grain_no.split("_")[-1] + "_" + mol_no.split("_")[-1]] = {
                     "image": filename,
-                    "grain_number": grain_no.split("_")[-1],
+                    "grain_number": int(grain_no.rsplit("_", maxsplit=1)[-1]),
+                    "molecule_number": int(mol_no.rsplit("_", maxsplit=1)[-1]),
                 }
                 molstats[grain_no.split("_")[-1] + "_" + mol_no.split("_")[-1]].update(tracing_stats)
                 LOGGER.info(f"[{filename}] : Finished splining {grain_no} - {mol_no}")
 
             except Exception as e:  # pylint: disable=broad-exception-caught
```
ns-rse, line 600: Broad exceptions should be avoided if possible.

Are there any specific errors that have arisen during processing that we could explicitly list here?

topostats/tracing/tracingfuncs.py
```diff
@@ -531,3 +531,26 @@ def coord_dist(coords: npt.NDArray, px_2_nm: float = 1) -> npt.NDArray:
...
+
+    Parameters
+    ----------
+    list_of_values : list
+        List of numbers.
+
+    Returns
+    -------
+    list
+        A list of combinatorics pairs of values, retaining the initial order.
```
ns-rse, line 550:
```suggestion
        A list of combinatoric pairs of values, retaining the initial order.
```

topostats/tracing/nodestats.py
```diff
@@ -853,7 +853,7 @@ def analyse_node_branches(
         if len(crossing_quants) <= 1:
             conf = None
         else:
-            combs = nodeStats.get_two_combinations(crossing_quants)
+            combs = get_two_combinations(crossing_quants)
```
ns-rse, line 856: Perhaps obvious but we should be explicit and generally try and avoid abbreviations (`combs` are also things we style our hair with!).

```suggestion
            combinations = get_two_combinations(crossing_quants)
```

topostats/tracing/tracingfuncs.py
```diff
@@ -531,3 +531,26 @@ def coord_dist(coords: npt.NDArray, px_2_nm: float = 1) -> npt.NDArray:
...
+
+def get_two_combinations(list_of_values) -> list:
+    """
+    Obtain all paired combinations of values in the list.
+
+    Example: [1,2] -> [[1,2]], [1,2,3] -> [[1,2],[1,3],[2,3]]
+
+    Parameters
+    ----------
+    list_of_values : list
```
ns-rse, line 544: Will it always be a `list` or can/will Numpy Arrays also be be passed in?

topostats/tracing/tracingfuncs.py
```diff
@@ -531,3 +531,26 @@ def coord_dist(coords: npt.NDArray, px_2_nm: float = 1) -> npt.NDArray:
             dist += 1
         dist_list.append(dist)
     return np.asarray(dist_list) * px_2_nm
+
+
+def get_two_combinations(list_of_values) -> list:
```
ns-rse, line 536:
```suggestion
def get_two_combinations(list_of_values: list) -> list:
```

topostats/tracing/tracingfuncs.py
```diff
@@ -531,3 +531,26 @@ def coord_dist(coords: npt.NDArray, px_2_nm: float = 1) -> npt.NDArray:
...
+    """
+    Obtain all paired combinations of values in the list.
+
+    Example: [1,2] -> [[1,2]], [1,2,3] -> [[1,2],[1,3],[2,3]]
```
ns-rse, line 540: Perhaps split the two examples onto separate lines, we can also use in-line examples as shown in [Numpydoc Style](https://numpydoc.readthedocs.io/en/latest/format.html#examples).

Suggest moving this to after the `Return` section as...

```suggestion
    Examples
    --------
    >>> get_two_combinations([1,2])
    array([[1,2]])
    >>> get_two_combinations([1,2,3])
    array([[1,2],[1,3],[2,3]])
```

Examples can also be used as [doctests](https://docs.pytest.org/en/8.3.x/how-to/doctest.html) so I'm keen to have more of these throughout the code base.

topostats/tracing/tracingfuncs.py
```diff
@@ -531,3 +531,26 @@ def coord_dist(coords: npt.NDArray, px_2_nm: float = 1) -> npt.NDArray:
...
+    list_of_values : list
+        List of numbers.
+
+    Returns
+    -------
+    list
+        A list of combinatorics pairs of values, retaining the initial order.
+    """
+    combs = []
+    for i in range(len(list_of_values) - 1):
```
ns-rse, line 553: Probably faster to leverage [`itertools.combinations()`](https://docs.python.org/3/library/itertools.html#itertools.combinations).

Could then return

```python
def get_combinations(list_of_values, r)
return list(itertools.combinations(list_of_values, r))
```

Documentation states it retains the sort order of the input values.

This may mean the whole function is redundant and instead the calls to `get_two_combinations()` could be replaced with `combinations(<some_list>, <size)`.

topostats/validation.py
```diff
@@ -983,6 +987,20 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
...
             "savefig_dpi": int,
         },
+        "branch_indexes": {
+            "filename": str,
+            "title": str,
+            "image_type": Or(
+                "binary",
+                "non-binary",
+                error=(
+                    "Invalid value in config 'coloured_boxes.image_type', valid values " "are 'binary' or 'non-binary'"
```
ns-rse, line 997:
```suggestion
                    "Invalid value in config 'branch_indexes.image_type', valid values " "are 'binary' or 'non-binary'"
```

topostats/tracing/splining.py
```diff
@@ -587,13 +591,14 @@ def splining_image(
                 }
                 molstats[grain_no.split("_")[-1] + "_" + mol_no.split("_")[-1]] = {
                     "image": filename,
-                    "grain_number": grain_no.split("_")[-1],
+                    "grain_number": int(grain_no.rsplit("_", maxsplit=1)[-1]),
```
ns-rse, line 594: `maxsplit=1` looks redundant if you're only ever interested in the last element of the returned list as seems to be the case here and on the next line. There could be five `_` but the `[-1]` always picks up the last value.

```python
In [15]: "1_2_3_4_5_6".rsplit("_")[-1]
Out[15]: '6'

In [16]: "1_2_3_4_5_6".rsplit("_", maxsplit=1)[-1]
Out[16]: '6'

In [17]: "1_2_3_4_5_6".split("_")[-1]
Out[17]: '6'
```
```suggestion
                    "grain_number": int(grain_no.split("_")[-1]),
```

To which end you could also use `split` as is the case on line 597 as `rsplit()` is only really required when you do need to use `maxsplit`.

Same for next line too.

topostats/tracing/disordered_tracing.py
```diff
@@ -384,6 +373,128 @@ def trace_image_disordered(  # pylint: disable=too-many-arguments,too-many-local
...
+    skan_df["min_value"] = skan_df.apply(lambda x: segment_heights(x, skan_skeleton, image).min(), axis=1)
+    skan_df["median_value"] = skan_df.apply(lambda x: np.median(segment_heights(x, skan_skeleton, image)), axis=1)
+    skan_df["mid_value"] = skan_df.apply(segment_middles, skan_skeleton=skan_skeleton, image=image, axis=1)
+
+    # remove unused skan columns
+    return skan_df[
+        [
+            "image",
+            "grain_number",
+            "branch-distance",
```
ns-rse, line 413: Suggest using `_` in the variable names being added.

You can use `-` in Pandas variable names but it then prevents using the `.` (dot) notation to refer to an individual series of a dataframe.

I pointed this out on Skan and my pull request to convert it has been [merged](https://github.com/jni/skan/pull/215) but is awaiting a new release.

ns-rse (changes requested):
Finally got round to reviewing, hope the suggestions are useful.

topostats/tracing/splining.py
ns-rse: Is it [C0207](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/use-maxsplit-arg.html)?

If so the comment suggests we may be able to disable it as there aren't going to be a huge number of splits in this case (just one to separate `grain_mol` I think).

> Be aware that the performance improvement from not splitting the string so many times will only be realized in cases presenting more instances of the splitting character than the minimal example here.

topostats/tracing/pruning.py
MaxGamill-Sheffield: cool, done! Also mine seems to have renamed docstrings - or at least my process of `cmd`+`shift`+`h` did :)

topostats/tracing/pruning.py
MaxGamill-Sheffield: Honestly I may just fully remove the %-based functionality for favour of nm length-based pruning as it has led to some confusion among users and the iterative pruning can cause some possibly unexpected behaviour as the % length changes.

It's just that it was the "old topostats" default so wanted to keep it for tests etc. Do you know if any of your (@ns-rse) modified tests or @SylviaWhittle smoke tests use the `pruning_params.max_length` parameter value of "-1"?

If not this should be an easy remove.

topostats/tracing/splining.py
MaxGamill-Sheffield: Not that I know of, but it was meant to be a catch-all to not stop all the TS processing. 

I know it's bad practice but I think it's more detrimental if an unknown-to-us exception of prevents the processing of a users whole dataset and they don't have the coding knowledge to read the error and add the specific exception themselves. This way, if they have an issue with the processing not capturing the molecule(s), the error is still output in the logs should they want to bring it to the TS forums to help diagnose the issue, or decide that the one file/molecule isn't something to be worried about.

So a user experience vs best practice debate. What do you think?

topostats/tracing/nodestats.py
MaxGamill-Sheffield: "combs" -> "combinations" and "comb" -> "pair" (when looping through "combinations") across all of the newly added files

topostats/tracing/tracingfuncs.py
MaxGamill-Sheffield: Cool! I've subbed that in but it also needed converting to a list

topostats/tracing/tracingfuncs.py
MaxGamill-Sheffield: get_two_combs removed for itertools.combinations :)

topostats/tracing/tracingfuncs.py
MaxGamill-Sheffield: get_two_combs removed for itertools.combinations :)

topostats/tracing/tracingfuncs.py
MaxGamill-Sheffield: get_two_combs removed for itertools.combinations :)

topostats/tracing/splining.py
MaxGamill-Sheffield: Yep and cool, I've reverted and added the pylint: disable comment so it passes

topostats/tracing/splining.py
ns-rse: > I know it's bad practice but I think it's more detrimental if an unknown-to-us exception of prevents the processing of a users whole dataset and they don't have the coding knowledge to read the error and add the specific exception themselves. This way, if they have an issue with the processing not capturing the molecule(s), the error is still output in the logs **should they want to bring it to the TS forums to help diagnose the issue, or decide that the one file/molecule isn't something to be worried about.**


Emphasis on what I think the key point is. We don't see a huge amount of reports about errors really. @TFirth2 highlighted one of the problems with having a broad `except` elsewhere in the codebase in #887 although in this instance the offending file is returned in the log. Beyond that we rarely see reports of problematic images stopping processing.

I'm not sure how many actual reports of such errors we would see if processing doesn't halt. I get the impression there are usually quite a few pictures being processed (batch processing is after all one of the original purposes of TopoStats) and so one not processing will likely just be ignored and we won't see a report of the problem logged. In turn we're less likely to find out what errors can arise and then never be able to improve and handle them.

> So a user experience vs best practice debate. What do you think?

More an issue of how to improve the software and find out about images that are problematic from my perspective.

topostats/tracing/disordered_tracing.py
MaxGamill-Sheffield: cool I've renamed the columns but we'll need to remove that section when Skan updates :)

SylviaWhittle:
> @SylviaWhittle I think the merge conflicts in `tests_splining` are to merge both, or one is outdated, as they both consider he whole file. Please could you take a look at which one is to be merged and resolve this?

Yeah defo merge both I think

topostats/plotting_dictionary.yaml
```diff
@@ -125,18 +125,21 @@ mask_grains:
...
   title: "Mask for Grains"
   image_type: "binary"
+  mask_cmap: "binary"
   savefig_dpi: 100
   core_set: false
 labelled_regions_01:
   filename: "18-labelled_regions"
   title: "Labelled Regions"
   image_type: "binary"
+  mask_cmap: "rainbow"
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Examples
New:
Screenshot 2024-10-01 at 21 19 51

Old (found from an old TopoStats):
Screenshot 2024-10-01 at 21 27 13

topostats/tracing/splining.py
MaxGamill-Sheffield: Interestingly #887 was created _because_ it stopped processing the dataset and the want was to know which image caused the error so it could be removed and the dataset run until the end.

Although I do see your point and concerns about not knowing about flaws in the code and employing the ostrich algorithm.

Could instead, we go through and change logging levels for `error` / `debug` statements for example, and by default use `info` to produce a reduced logging output and thus be more obvious to users that things have gone wrong and where?
I know when I first started I just used `logger.info` for everything as anything else was at one point broken but seems like "warning" and "error" and "debug" now work. 

I'll sleep on it tonight and have a chat with others tomorrow and gather their opinions too.

topostats/plotting_dictionary.yaml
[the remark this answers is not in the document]
ns-rse: That is not a problem at all, I've just never used the `rainbow` cmap before and had to go and check if it was defined within `topostats/themes.py` (it wasn't) so then went hunting to find where it was from and wanted to check it was using that and not a custom one that was missing from this Pull Request.

ns-rse:
For the `pyproject.toml` conflict would it be possible to order the dependencies alphabetically please (both now and in the future rather than appending to the list). It makes it much easier to read and find what dependencies there are (I corrected this the other day in something).

`AFMReader` will also need bumping to the top of the list.

topostats/tracing/pruning.py
SylviaWhittle: Yeah lots of my tests use `pruning_params.max_length = -1` but this can be changed easily. Any ideas on good values for the cats & rep ints?

topostats/tracing/splining.py
SylviaWhittle: The deadline is tight here. I in principle agree that we should be letting things crash, which will lead to less tech debt and more stable software later on, however we need this merging ASAP for the publication to go ahead.

I personally would like to overhaul error handling a little in topostats, and would like to make an issue about it. Would it be okay if this be handled there instead of here?

SylviaWhittle:
- Alphabetised dependencies
- Resolved merge conflict

topostats/tracing/pruning.py
MaxGamill-Sheffield: Group says to remove the % length as it doesn't make sense for mixed samples / conformations and requires back calculating to nm anyway for unknown structures. So I'll do that.

@SylviaWhittle ~10nm should be perfect (double the tip-convolved DNA width)

MaxGamill-Sheffield: Removed the -1 dependancy in [d0751ad](https://github.com/AFM-SPM/TopoStats/pull/898/commits/d0751adab8139465ada0551e8515544c24a3878b)

topostats/tracing/splining.py
MaxGamill-Sheffield: Added these in commit [9a73090](https://github.com/AFM-SPM/TopoStats/pull/898/commits/9a7309013f1d114f105c0b09a7aad687179d9d40) and issue

MaxGamill-Sheffield:
Comments have been addressed so is this ready to go into `[maxgamill-sheffield/800-better-tracing]`?

SylviaWhittle:
Just fixed all the tracing tests 👍

llwiggins:
This all looks good to me and I'd be happy for it to be merged, I tested functionality out on a set of varied molecules and all were handled correctly! 👍

ns-rse:
> I tested functionality out on a set of varied molecules and all were handled correctly! 👍

Could these perhaps be used as part of the test suite? It could be as simple as additional regression tests that check the Matplotlib images I suspect are being manually reviewed in a notebook to make the decision that they are being handled correctly, but it would give us some reassurance in the future if/when any of the code is refactored that things aren't broken, or if they do change that we take time to consider if the change is intended.

PR #900, opened 2024-09-12
chore: fix linting errors

ns-rse:
Fixes a bunch of linting errors, some notes left in place for tasks to be completed (e.g. see #899).

PR #903, opened 2024-09-13
tests(processing): Improves logic and tests for check_run_steps()

### LINKED ISSUE #899: Extend tests  of check_run_steps() to test nodestats_run and splining_run options
With the splitting of configuration into separate stages for nodestats and splining we need to check that configuration options are consistent, i.e. that previous stages that are required are configured to run.

Currently that ~~isn't implemented or tested~~ is implemented but isn't tested (although I have setup the parameterisation of `tests/test_processing.py::test_check_run_steps` to do so because linting was failing).

The logic to make these checks and tests will need implementing and merging to `maxgamill-sheffield/800-better-tracing`.

ns-rse:
Closes #899

Also updates `test_process_stages()` although this includes disabling a one set of parameters for the time being until I work out what is now happening.

**NB** This should be merged **BEFORE** #900 so that that Pull Request/branch gets updated with these changes and everything is then merged into `maxgamill-sheffield/800-better-tracing`.

### DIFF
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -355,7 +355,7 @@ def run_grainstats(
         return create_empty_dataframe()
 
 
-def run_disorderedTrace(
+def run_disordered_trace(
     image: npt.NDArray,
     grain_masks: dict,
     pixel_to_nm_scaling: float,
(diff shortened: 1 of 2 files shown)


topostats/processing.py
```diff
@@ -1094,17 +1094,53 @@ def check_run_steps(  # noqa: C901
     splining_run : bool
         Flag for running DNA Tracing.
     """
+    LOGGER.debug(f"{filter_run=}")
```
SylviaWhittle, line 1097: Very good idea!!

SylviaWhittle (approved):
This is great, thank you for your apparently infinite patience

PR #906, opened 2024-09-16
fix: reinstate grains try except

ns-rse:
Commit `de517ff6` inadvertently disabled the broad `try: ... except:` within `run_grains()`. This is being reinstated and will be addressed when #887 is worked on.


SylviaWhittle (approved):
Apologies for causing this issue 😅

ns-rse:
Not a problem, its not like it was deliberate!

We'll get this addressed with #887

PR #907, opened 2024-09-17
test(utils): Fixing broken tests

ns-rse:
+ `test_create_empty_dataframe()` didn't have `molecule_number` reverted to `grain_number`.
+ `topostats.utils.update_plotting_config()` had a clause which only checked if an image was in the `core_set` and failed if it then didn't have a `mask_cmap` entry.


SylviaWhittle (approved):
🙏

PR #918, opened 2024-09-25
tests: Remove restriction on pytest<8.0.0 (cherry pick)

ns-rse:
Cherry pick cecf60fbabe17b12927284b2b5099e5c4c092c4d which removes the restriction of `pytest<8.0.0` and switches to usine `request.getfixturevalue()` for parameterised tests that use fixtures

This is a prelude to fixing these failing tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ requires-python = ">=3.9"
 dependencies = [
   "h5py",
   "igor2",
+  "keras",
   "matplotlib",
   "numpy",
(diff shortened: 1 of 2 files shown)


MaxGamill-Sheffield (approved):
Am I correct in thinking that the `.toml` changes are the ones you wanted in this PR and not the dnatracing tests?

Either way the dnatracing tests will soon be removed so it's not a problem - this is ready to go when you are :)

Issue #928, opened 2024-10-03
Check Python 3.13 compatibility


ns-rse:
Python 3.13 release is [imminent](https://devguide.python.org/versions/) (due next week) and features the option to disable the "Global Interprator Lock" (GIL) which controls parallel processes.

The [pytest-freethreaded](https://pypi.org/project/pytest-freethreaded/) extension allows checking of the impact of running testes with the GIL disabled.

We should test our code against this, may not require formal tests to be written (which would require additional images to be added to the test suite) but it will be important to know what, if any, this impact has.

PR #931, opened 2024-10-07
ci: restrict supported python version to <3.12

### LINKED ISSUE #930: Remove support for Python 3.12
In light of #927 where we have discovered a performance regression of the dependency [topoly](https://topoly.cent.uw.edu.pl/documentation.html) with Python 3.12 we should, for the better tracing work and release remove support for Python 3.12 from TopoStats until this has been resolved.

## `pyproject.toml`

```toml
requires-python = ">=3.9, <3.12"
```

## `.github/workflows/tests.yaml`

```yaml
        python-version: ["3.9", "3.10", "3.11"]
```

ns-rse:
Closes #930

Once Topoly performance regression has been resolved we should unpin this upper bound and ideally support upto Python 3.13.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11"]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python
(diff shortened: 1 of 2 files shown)


SylviaWhittle (approved):
Very much approve. An on the note of 3.13, we should be careful as arm64 support (for Macs) is slow in many packages and so might not be able to support it until they do. I currently run 3.11 in many environments because of some issues with packages not yet supporting arm64 in 3.12.5

ns-rse:
> An on the note of 3.13, we should be careful as arm64 support (for Macs) is slow in many packages and so might not be able to support it until they do. I currently run 3.11 in many environments because of some issues with packages not yet supporting arm64 in 3.12.5

Cool, thanks for the heads up, I have #928 to specifically keep track of it.

PR #934, opened 2024-10-08
Maxgamill sheffield/btr docs

ns-rse:


ns-rse:
Closing, wrong to merge into `main` as it branched from `maxgamill-sheffield/800-better-tracing` and doing so would pull in all that work (minus the changes that are on-going).

PR #936, opened 2024-10-09
docs: Document refactored tracing

ns-rse:
Adds documentation, written by @MaxGamill-Sheffield, for...

- Disordered tracing
- Nodestats
- Ordered tracing
- Splining

I squashed the history of all the incremental commits that wrote this documentation but have left the changes I've made recently as individual commits.

Locally the `advanced` Table of Contentes (`toctree`) is not rendered, I need to check if this is the case on the GitHub CI rendered pages too though (which this PR should trigger).

**NB** Ideally this should be merged to `maxgamill-sheffield/800-better-tracing` before it is merged to `main` so the documentation and features arrive on `main` at the same time.


MaxGamill-Sheffield:
@ns-rse you've requested a review so are you happy with the toctrees?
> Locally the advanced Table of Contentes (toctree) is not rendered, I need to check if this is the case on the GitHub CI rendered pages too though (which this PR should trigger).

And shall I re-examine the docs to ensure everything is rendered ok? (Also, how do I do this for CLI and not local?)

ns-rse:
Evening @MaxGamill-Sheffield 

Not completely happy with `toctrees` but for some reason this hasn't triggered the build of the pages on CI which I was hoping for to help me see if its something with my local setup.

Its slightly obfuscated but the commands to replicate what CI does using `sphinx-multidocs` are in `.github/workflows/sphinx_docs_to_gh_pages.yaml`.

```bash
sphinx-multiversion . _build/html "../*setup*" "../*tests*" "../*.ipynb" ../demo.py ../make_baseline.py ../jupyter_notebook_config.py ../demo_ftrs.py
```

Weirdly using `sphinx-build` includes the new `toctree` but using `sphinx-multiversion` doesn't (it's essentially a wrapper which checks out each tag and runs `sphinx-build` on each tagged commit and `maind` and placing the output under `_build/html/<tag|main>` so that there are multiple versions of the documentation).

I add reviewers to drafts when creating them (so I don't forget when I hit the "Ready for review") but in this instance will see if making it "Ready for review" triggers the build.

If not and for expedience I will rejig how the advanced docs are linked from the front page.

SylviaWhittle:
Not sure I'm understanding this. 

Is the issue that we aren't able to check that they're rendering okay on gh-pages? If so, if we are happy with the contents could we try just pushing to main which would presumably force a build?

ns-rse:
> we aren't able to check that they're rendering okay on gh-pages?

Correct.

> If so, if we are happy with the contents could we try just pushing to main which would presumably force a build?

This would result in the [`main` documentation]() having documentation which isn't quite yet on that branch.

I'm going to look at this today and may just ditch having an additional `toctree` in favour of an extra `Advanced` entry in the existing one and link articles from there (as I can envisage this section growing considerably) but go sidetracked trying to work out why tests are failing on Windows and possible solutions.

ns-rse:
Ready for review, its not triggering the GitHub Action to build the documentation here on GitHub because that is configured to only run on Pull Requests that target the `main` branch. Once this is merged into `maxgamill-sheffield/800-better-tracing` it will trigger a build.

I couldn't work out how to get the second `.. toctree::` so have instead added a `docs/advanced.md` which serves as an index for Advanced topics which reside under `docs/advanced/`. Hopefully over time the pages detailing the inner workings will grow.

SylviaWhittle (approved):

PR #938, opened 2024-10-09
Fixes failing pruning tests on 800-better-tracing

MaxGamill-Sheffield:
Fixes the failing pruning tests by:
• adding to the pruner's` topostatsPrune` functions the `pixel_to_nanometer_scaling` (of value 1).
• adjusted the default `-1` length flag to values which give roughly the correct skeletons in the `prune_by_length` tests.
• un-skipping some tests
• adjusted any skeletons to match the expected i.e. only endpoint branches should be pruned as the convolution of the skeleton is not re-done so pruning is not iterative. Also, the nodes remain after pruning all branches of a star-like molecule
• also removes `-1` flag test as we do not support this anymore


topostats/tracing/pruning.py
```diff
@@ -649,7 +649,6 @@ def filter_segments(self, segments: npt.NDArray) -> npt.NDArray:
             height_values = self._get_branch_medians(segments)
         elif self.method_values == "mid":
             height_values = self._get_branch_middles(segments)
-        print(f"{height_values=}")
```
ns-rse, line 652: :+1:

tests/tracing/test_pruning.py
```diff
@@ -1098,50 +1137,6 @@ def test_prune_by_length(
             id="Pruning by length and height disabled",
             # marks=pytest.mark.skip(),
         ),
-        pytest.param(
```
ns-rse, line 1101: Why has this set of parameters been removed? Is it because the ability to prune by a fixed (or user specified) length has (or will be) removed?

If so has that been removed from the code base?

ns-rse (commented):
One question in-line and based on the Pull Request comment also...

> • un-skipping some tests

I couldn't see any changes where `marks_pytest.mark.sip()` has been commented out (most tests already have this commented out as they were put in to skip the tests until they were fixed). Maybe the comments were added prior to making the branch so they're not showing in this Pull Request? :thinking: That isn't a problem just a guess and I'm happy with leaving them in commented.

tests/tracing/test_pruning.py
SylviaWhittle: Default -1 length flag has been changed to 10. Also the ability to use -1 has been removed

Hence this parameterised test is no longer needed as it checked the -1 functionality (15% of length)

MaxGamill-Sheffield:
> One question in-line and based on the Pull Request comment also...
> 
> > • un-skipping some tests
> 
> I couldn't see any changes where `marks_pytest.mark.skip()` has been commented out (most tests already have this commented out as they were put in to skip the tests until they were fixed). Maybe the comments were added prior to making the branch so they're not showing in this Pull Request? 🤔 That isn't a problem just a guess and I'm happy with leaving them in commented.

@ns-rse
Lines 1363 -> 1358 the pytes.mark skip has been commented out and Sylvia beat me to the -1 flags and tests

tests/tracing/test_pruning.py
ns-rse: Is this change pending?

Currently on `maxgamill-sheffield/800-better-tracing` the default value for `max_length` is `None` when the [class is initalised](https://github.com/AFM-SPM/TopoStats/blob/33274043df8c6beedad0b89f60e9e97f378c27ef/topostats/tracing/pruning.py#L240). I recall there was at some point something along the lines of...

```python
self.max_length = max_length if max_length is None else length * 0.15
```

...and I suggested that it be made configurable, which it now is, but the default isn't 10%, its currently `None`.

ns-rse:
> @ns-rse
> Lines 1363 -> 1358 the pytes.mark skip has been commented out and Sylvia beat me to the -1 flags and tests

And that is why I shouldn't work late at night. Thanks for directing me to the correct line (searching for `# marks=pytest.mark.skip()` was never going to work!).

tests/tracing/test_pruning.py
MaxGamill-Sheffield: Ahh right, yes it no longer works as a percentage, that is the "feature" that was dropped in favour of a physical value (nm). So `-1` which meant 15% of total length is now `10` for 10nm

The line you reference shows it's default initialisation value can be `None` which is allowed as it will just skip the `prune_by_length` function, I'll re-add the default `None` test then :).

MaxGamill-Sheffield: Tell a lie - The test before it tests `None` for height and length pruning so I still don't think we need the test.

SylviaWhittle (approved):
Have had a look at the `test_prune_skeleton` and `pruning by length and height disabled` parameterisation and it covers the initialisation default values so seems appropriate. It also tests variable lengths in nanometres so covers the kinds of values that users would change it to.

PR #932, opened 2024-10-07
Adding the Better Tracing Suite into Main

Max-Gamill:
Closes #800 

This aims to add features which split out the DNATracing pipeline into smaller parts (disordered tracing, ordered tracing and splining) while adding individual analyses in each of these to be more modular. It also adds topological functions and analyses into ordered tracing, facilitating the processing of catenated molecules as separate objects, and includes a new module to handle and analyse crossings of DNA segments.

Issue #921, opened 2024-10-01
Improve automatic testing strategy


ns-rse:
The use of [GitFlow Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) workflow to address the work that has been done on improving tracing and untangling molecules means that because the test suite we have in place to ensure changes do not break existing functionality and expected output do not run on pull requests to the target development branch (`maxgamill-sheffield/800-better-tracing`) since currently the workflow is setup to only run on pull requests that target the `main` branch...

### `.github/workflows/tests.yaml`

```yaml
name: Tests (pytest)

on:
  pull_request:
    branches:
      - main
```

This has led to problems as the changes introduced in this branch have broken a number of tests (see #850).

We should be catching these much earlier in the development cycle and there are a couple of approaches which I would like to consider.

- Run CI on _all_ pull requests not just `main`.
- Run tests with `pre-commit` on only the changed/affected files using [pytest-testmon](https://github.com/tarpas/pytest-testmon/).

I would also advocate against the overly complicated [GitFlow Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) which uses development branches and instead think we should use either [GitHub or GitLab Flow](https://www.eisquare.co.uk/blogs/how-to-choose-your-branching-strategy) instead. I would personally favour the former ([GitHub Flow](https://docs.github.com/en/get-started/using-github/github-flow) as attempts so far to define clear sets of features and establish Milestones under which release branches can be developed in the [Gitlab Flow](https://about.gitlab.com/topics/version-control/what-is-gitlab-flow/) approach.

To date we have mostly used the [GitHub Flow](https://docs.github.com/en/get-started/using-github/github-flow) approach of merging all features to `main` and periodically tagging commits to trigger releases. This approach requires that tasks are broken down into smaller components which makes development and feedback considerably faster.

PR #939, opened 2024-10-10
Fix processing tests

SylviaWhittle:
This PR (hopefully) fixes the tests in `test_processing.py` for the #932


SylviaWhittle:
Ah heck, so this PR is just to address the processing tests, not the others... Any ideas on how to do a PR that fixes only some of the tests? @ns-rse

SylviaWhittle:
The processing tests pass on my machine, you might want to verify though 
image

ns-rse:
> Any ideas on how to do a PR that fixes only some of the tests? ￼@ns-rse

The Continuous Integration tests will continue to fail until all tests pass, no way around that I can think of without switching to using `pytest --testmon` in CI if that is even possible as it would be problematic if say a newer package were installed in CI than on a local machine where the `.testmondata`.

Its useful to run the whole test suite in CI though and ensuring it runs on _all_ pull requests was something I deliberately introduced the other day with #921 as had this been the case from the start (rather than just running on PRs that target `main`) we would have known about and addressed the failing tests earlier.

MaxGamill-Sheffield (approved):
Looks good and TS runs!

topostats/grains.py
```diff
@@ -295,6 +300,41 @@ def remove_small_objects(self, image: np.array, **kwargs) -> npt.NDArray:
             return small_objects_removed > 0.0
         return image
 
+    def remove_objects_too_small_to_process(
+        self, image: npt.NDArray, minimum_size_px: int, minimum_bbox_size_px: int
```
ns-rse, line 304: Do we expect this method to be used independently of the processing?

If not then I don't see any advantage have `minimum_size_px` or `minimum_bbox_size_px` as arguments to the method as they are set as attributes to the class when it is initialised and available as `self.minimum_grain_size_px` and `self.minimum_bbox_size_px`.

The comment above on hardcoding suggests its not something that will ever be changed.

PR #945, opened 2024-10-13
Add topostats file helper class

SylviaWhittle:
This PR would add a small helper class in `topostats.io` to assist users that want to explore & retrieve data contained in `.topostats` files, as we have had feedback from the experimentalists that navigating the `.hdf5` file structure is prohibitively complex / difficult to do manually.

Previously, to load the file in a notebook, one had to:
```python
from pathlib import Path

import h5py

from topostats.io import hdf5_to_dict


file = Path("./path/to/file.topostats")
with h5py.File(file, "r") as f:
    data_dict = hdf5_to_dict(f, "/")

# Then try to manually navigate the dictionary to find the specific item wanted
data = data_dict["ordered_trace_heights"]["0"]
# get the keys wrong
>>> ValueError
# manually print keys at each level, akin to doing lots of ls, cd
print(data_dict.keys())
data = data_dict["grain_trace_data"]
print(data.keys())
.
.
.
```

The TopoFileHelper class adds some methods to help with this:
- `pretty_print_structure()` will print the entire structure (but not messy dictionaries of arrays!):
```
[./tests/resources/file.topostats]
├ filename
│   └ minicircle
├ grain_masks
│   └ above
│       └ Numpy array, shape: (1024, 1024), dtype: int64
├ grain_trace_data
│   └ above
│       ├ cropped_images
│       │   └ 21 keys with numpy arrays as values
│       ├ ordered_trace_cumulative_distances
│       │   └ 21 keys with numpy arrays as values
│       ├ ordered_trace_heights
│       │   └ 21 keys with numpy arrays as values
│       ├ ordered_traces
│       │   └ 21 keys with numpy arrays as values
│       └ splined_traces
│           └ 21 keys with numpy arrays as values
├ image
│   └ Numpy array, shape: (1024, 1024), dtype: float64
├ image_original
│   └ Numpy array, shape: (1024, 1024), dtype: float64
├ img_path
│   └ /Users/sylvi/Documents/TopoStats/tests/resources/minicircle
├ pixel_to_nm_scaling
│   └ 0.4940029296875
└ topostats_file_version
    └ 0.2
```
- `find_data()` will perform a strict search for the keys (given in a list) and if no match is found, perform a partial search to find possible matches that the user intended. Eg:
```python
topofilehelper.find_data(["ordered_trace_heights", "0"])
```
```
 [ Searching for ['ordered_trace_heights', '0'] in ./tests/resources/file.topostats ]
 | [search] No direct match found.
 | [search] Searching for partial matches.
 | [search] !! [ 1 Partial matches found] !!
 | [search] └ grain_trace_data/above/ordered_trace_heights/0
 └ [End of search]
```
- `get_data()` Simply retries data when provided with a key string separated by "/"s:
```python
ordered_trace_heights = topofilehelper.get_data("grain_trace_data/above/ordered_trace_heights/0")
```
- `data_info()` Prints a little information about the value at a specific key:
```python
topofilehelper.data_info("grain_trace_data/above/ordered_trace_heights/0")
topofilehelper.data_info("grain_trace_data/above/ordered_trace_heights")
```
```
Data at grain_trace_data/above/ordered_trace_heights/0 is a numpy array with shape: (95,), dtype: float64
Data at grain_trace_data/above/ordered_trace_heights is a dictionary with 21 keys of types {<class 'str'>} and values of types {<class 'numpy.ndarray'>}
```

No tests yet, would want feedback first

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -1291,3 +1291,45 @@ class TopoFileHelper:
         self.topofile: Path = Path(topofile)
         with h5py.File(self.topofile, "r") as f:
             self.data: dict = hdf5_to_dict(open_hdf5_file=f, group_path="/")
+    def find_data(self, search_keys: list) -> None:
+        """
+        Find the data in the dictionary that matches the list of keys.
+
+        Parameters
+        ----------
+        search_keys : list
+            The list of keys to search for.
+
+        Returns
+        -------
+        None
+        """
+        # Find the best match for the list of keys
+        # First check if there is a direct match
+        LOGGER.info(f"[ Searching for {search_keys} in {self.topofile} ]")
+
+        try:
+            current_data = self.data
+            for key in search_keys:
+                current_data = current_data[key]
+
+            LOGGER.info("| [search] Direct match found")
+        except KeyError:
+            LOGGER.info("| [search] No direct match found.")
+
+        # If no direct match is found, try to find a partial match
+        LOGGER.info("| [search] Searching for partial matches.")
+        partial_matches = self.search_partial_matches(data=self.data, keys=search_keys)
+        if partial_matches:
+            LOGGER.info(f"| [search] !! [ {len(partial_matches)} Partial matches found] !!")
+            for index, match in enumerate(partial_matches):
+                match_str = "/".join(match)
+                if index == len(partial_matches) - 1:
+                    prefix = "| [search] └"
+                else:
+                    prefix = "| [search] ├"
+                LOGGER.info(f"{prefix} {match_str}")
+        else:
+            LOGGER.info("| [search] No partial matches found.")
+        LOGGER.info("└ [End of search]")
+        return
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -1333,3 +1333,44 @@ class TopoFileHelper:
             LOGGER.info("| [search] No partial matches found.")
         LOGGER.info("└ [End of search]")
         return
+    def data_info(self, location: str, verbose: bool = False) -> None:
+        """Get information about the data at a location.
+
+        Parameters
+        ----------
+        location : str
+            The location of the data in the dictionary, separated by '/'.
+        """
+        # If there's a trailing '/', remove it
+        if location[-1] == "/":
+            location = location[:-1]
+        keys = location.split("/")
+
+        try:
+            current_data = self.data
+            for key in keys:
+                current_data = current_data[key]
+        except KeyError as e:
+            LOGGER.error(f"[ Info ] Key not found: {e}, please check the location string.")
+            return
+
+        if isinstance(current_data, dict):
+            key_types = {type(k) for k in current_data.keys()}
+            value_types = {type(v) for v in current_data.values()}
+            LOGGER.info(
+                f"[ Info ] Data at {location} is a dictionary with {len(current_data)} "
+                f"keys of types {key_types} and values "
+                f"of types {value_types}"
+            )
+            if verbose:
+                for k, v in current_data.items():
+                    LOGGER.info(f"  {k}: {type(v)}")
+        elif isinstance(current_data, np.ndarray):
+            LOGGER.info(
+                f"[ Info ] Data at {location} is a numpy array with shape: {current_data.shape}, "
+                f"dtype: {current_data.dtype}"
+            )
+        else:
+            LOGGER.info(f"[ Info ] Data at {location} is {type(current_data)}")
+
+        return
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -1291,6 +1291,91 @@ class TopoFileHelper:
         self.topofile: Path = Path(topofile)
         with h5py.File(self.topofile, "r") as f:
             self.data: dict = hdf5_to_dict(open_hdf5_file=f, group_path="/")
+
+    def search_partial_matches(self, data: dict, keys: list, current_path: list | None = None):
+        """
+        Find partial matches to the keys in the dictionary.
+
+        Recursively search through nested dictionaries and keep only the paths that match the keys in the correct order,
+        allowing gaps between the keys.
+
+        Parameters
+        ----------
+        data : dict
+            The dictionary to search through.
+        keys : list
+            The list of keys to search for.
+        current_path : list, optional
+            The current path in the dictionary, by default [].
+
+        Returns
+        -------
+        list
+            A list of paths that match the keys in the correct order.
+        """
+        if current_path is None:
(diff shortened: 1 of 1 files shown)


ns-rse:
Not ignoring this, have had a scan through and it looks good but focusing on the various outstanding issues with the better tracing merger.

SylviaWhittle:
> Not ignoring this, have had a scan through and it looks good but focusing on the various outstanding issues with the better tracing merger.

All good, not wanting this to take time away from more important stuff, it can wait and I want user opinions first too :)

MaxGamill-Sheffield:
Can we work into this a notebook on opening and extracting data from the file too, to address the comments from the workshop day?

SylviaWhittle:
Added a notebook showing how to use the class in `./notebooks/`

topostats/io.py
```diff
@@ -1284,3 +1284,369 @@ def dict_to_json(data: dict, output_dir: str | Path, filename: str | Path, inden
...
+
+    Parameters
+    ----------
+    topofile : Path
+        Path to the .topostats file.
+
+    Examples
+    --------
+    Creating a helper object.
+    ```python
```
ns-rse, line 1301: Not sure about how this renders when Sphinx's Autoapi-doc parses it to generate API docs in the webpage as docstrings are Restructured text. Might be worth using the `.. code-block::` approach, see @MaxGamill-Sheffield solution in commit b953634e6c98156d6d26e35518302c2167b9739e .

SylviaWhittle: Very good point thank you, I'll do that

ns-rse:
I think the tests are under Python 3.9 failing because under that version we need the following import...

```python
from __future__ import annotations
```

notebooks/topostats_file_helper_example.ipynb
ns-rse, line 1: I think the Notebook would be easier to read if the comments were moved to Markdown sections to delineate the code examples. Otherwise may as well just include a codeblock in `docs/advanced/topostats_file_helper.md` and be a simpler solution to documentation as its a web-page people can go to, they wouldn't need to activate a virtual environment and then start a Jupyter. Code chunks could be copy and pasted (I'd have to work out how to enable a button to support that though).

ns-rse:
Just came across [h5glance](https://pypi.org/project/h5glance/) after the Skan developer mentioned it on Mastodon. I wonder if using this would be a simpler solution, it sounds as though it might work within Jupyter Notebooks too.

SylviaWhittle:
> Just came across [h5glance](https://pypi.org/project/h5glance/) after the Skan developer mentioned it on Mastodon. I wonder if using this would be a simpler solution, it sounds as though it might work within Jupyter Notebooks too.

Good find!

It does work Wonderfully in notebooks and it've even interactive! You can click on the items to expand / hide sub-fields.
image

**However** this is just for looking at `hdf5` files and not retrieving data from them (which would still require the `with h5py.File('testfile.hdf5', 'r') as f: <load stuff manually>)

I propose that I replace the code I wrote to display the contents of the file with `h5glance` but keep the methods I wrote to do data retrieval?

ns-rse:
Sounds like a plan.

From memory these are thin wrappers around being able to access dictionary items directly and I feel that its substituting learning how to work with dictionaries directly with learning how to use the wrappers. I'm of the opinion that the more general skill (working directly with dictionaries) has broader benefits to users in the long term.

:balance_scale:

SylviaWhittle:
There is an issue with `h5glance`, that it must be called from within the notebook. It cannot be called in a wrapper in a standard `.py` file. If this happens then this is the result: 

image

I tried this:
```python
    def pretty_print_structure(self) -> None:
        """
        Print the structure of the data in the data dictionary.

        The structure is printed with the keys indented to show the hierarchy of the data.
        """
        LOGGER.info(f"running h5glance")
        H5Glance(self.topofile)
```

and this:
```python
    def pretty_print_structure(self) -> None:
        """
        Print the structure of the data in the data dictionary.

        The structure is printed with the keys indented to show the hierarchy of the data.
        """
        LOGGER.info(f"running h5glance")
        result = H5Glance(self.topofile)
        print(result)
```

and neither produce the nested output needed.

If users want the interactive `h5glance` notebook UI I screenshotted earlier, they must call it explicitly in a notebook, `h5glance.H5Glance("./file.h5")`. 

So either they must remember `h5glance` as a separate tool to `TopoFileHelper` or we could provide my worse implementation as a built-in alternative in case they don't remember to use `h5glance`?

ns-rse:
I guess it depends how people are using the `.topostats` files? Are Notebooks widely used within the group and outside of it to explore `.topostats` files?

I'm not a great fan of re-inventing the :wheel: and if a tool exists I'll tend to advocate for its use over making something new which adds to our codebase and the overhead of maintenance.

As a general rule though, and perhaps I'm missing something, but `hdf5` files once loaded are essentially dictionaries. A wrapper to make exploring dictionaries is perhaps useful but it still requires learning how to use the wrapper. I'd personally advocate for helping people learn how to explore dictionaries as it gives them a transferable skill that can be used in lots of other scenarios.

I think I recall writing such when I wrote original Notebooks (see for example the line after `from topostats.io import read_yaml` in [this notebook](https://github.com/AFM-SPM/TopoStats/blob/main/notebooks/OUTDATED-00-Walkthrough-minicircle.ipynb)). We could point people to tutorials on how to use `.keys()` and `.values()` and iteration over dictionaries.

The [h5glance README](https://github.com/European-XFEL/h5glance) points to using the `h5py` package when wanting to work with such files within code but then that is what you are using here.

Having re-read the `Examples` you've written it looks like it is demonstrating to people how to use the helper to view files in a Notebook, which they could do with `h5glance` and its one less thing to maintain within TopoStats :balance_scale:

ns-rse:
Sorry to miss code clean, was embroiled in some bioinformatics work and didn't notice the time.

With regards to Notebook this might be an ideal opportunity to migrate to the newer [marimo](https://marimo.io/) which among other things has a major advantage of updating all dependent cells when an earlier one is re-run. 

For more on the problems marimo solves see the [faq](https://docs.marimo.io/faq#choosing-marimo).

I doubt there will be much of an overhead in migrating since its still a notebook running the cells so both Markdown and code cells could be copied over. There is a section on [migrating from Jupyter](https://docs.marimo.io/guides/coming_from/jupyter/).

ns-rse:
In light of the proposed restructuring (see #1102 ) would it be reasonable to close this and address loading of `.topostats` in the documentation once we know what that loopk like with the classes?

PR #946, opened 2024-10-14
ci: Adds tmate to tests.yaml workflow

### LINKED ISSUE #896: Incorporate tmate into CI to facilitate debugging
Discovered the [tmate GitHub action](https://github.com/marketplace/actions/debugging-with-tmate) which facilitates debugging of failed GitHub Actions.

In essence when a job fails it gives you the SSH connection details to the Virtual Machine that has been spun up, you can connect and play around, working out what has gone wrong. This shortens the feedback/development cycle.

It would be useful to incorporate this into the test and some other workflows TopoStats uses.

ns-rse:
Resolves #896

Adds [debugging with tmate](https://github.com/marketplace/actions/debugging-with-tmate) action to `tests.yaml` which
allows SSHing into GitHub Runners _if_ tests have failed.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -51,3 +51,8 @@ jobs:
           coverage xml
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v4
+
+      - name: Setup tmate session
+        if: ${{ failure() }}
+        uses: mxschmitt/action-tmate@v3
+        timeout-minutes: 5


SylviaWhittle (approved):
I assume this needs to be merged before it will work, but just in case, I thought I'd mention that for this PR, it doesn't seem to work?

image

ns-rse:
Thanks for the :white_check_mark: 

Nope that is by design, it only runs `if: ${{ failure() }}`occurs in earlier stages(line 56). 

That way we don't leave sessions open and accessible if they pass.

PR #954, opened 2024-10-15
Bump Python version to 3.12 in gh-pages workflow

ns-rse:
Checking to see if bumping version improves building of v2.0.0 release.

Annoyingly building this version on its own using

``` bash
git checkout v2.0.0
cd docs
make html
```

There is no error but when attempting to build with `sphinx-multiversion` which in the background switches to tagged
releases and runs `sphinx-build` (which is exactly what the `Makefile` does). There are errors from
`sphinx-autoapi-typehints` on this tagged release.


ns-rse:
Forked from the wrong branch :facepalm:

PR #956, opened 2024-10-15
800 better tracing angle out of range

### LINKED ISSUE #951: Nodestats test_nodestats_image raises runtime warning from calc_angles()
Running tests I noticed the following being returned...

```
tests/tracing/test_nodestats.py::test_nodestats_image[catenane]
tests/tracing/test_nodestats.py::test_nodestats_image[catenane]
tests/tracing/test_nodestats.py::test_nodestats_image[replication_intermediate, not pairing odd branches]
tests/tracing/test_nodestats.py::test_nodestats_image[replication_intermediate, pairing odd branches]
  /home/neil/work/git/hub/AFM-SPM/TopoStats/topostats/tracing/nodestats.py:1110: RuntimeWarning: invalid value encountered in arccos
    return abs(np.arccos(cos_angles) / np.pi * 180)  # angles in degrees

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

Relates to the `calc_angles()` static method (for which we don't currently have tests).

We should fix this, perhaps via a a `try: ... except:` depending on the nature of the problem, prior to #800 being completed via #932.

MaxGamill-Sheffield:
closes #951 

moves fill_diag() inside calc angles so there are no >1's caused by floating point errors.


SylviaWhittle (approved):
Yeah this makes a lot of sense 👍

PR #957, opened 2024-10-15
fix: docutils<0.21; unpin sphinx-autoapi

ns-rse:
Questing to solve the issues with pages failing to build I'm trying removing the `sphinx-autoapi` pinned version and at the same time ensuring `docutils<0.21` are installed as part of the docs dependencies. :crossed_fingers:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,10 +70,11 @@ tests = [
 ]
 docs = [
   "Sphinx",
+  "docutils<0.21",
   "myst_parser",
   "numpydoc",
   "pydata_sphinx_theme",
-  "sphinx-autoapi==3.1.0b0",
+  "sphinx-autoapi",
   "sphinx-autodoc-typehints",
   "sphinx-multiversion",
   "sphinx_markdown_tables",


SylviaWhittle:
You branched off of `main` but despite that the tests fail, which is really confusing. I am struggling to read the diff clearly in the GitHub view so can't tell exactly what the differences are.

image

SylviaWhittle:
In yet more confusion, I checked out this branch and ran the offending test on python 3.9 on my windows install and it passes??

(apologies for photo of a screen, I don't have a convenient way to transfer stuff from windows to macOS since it's so rare that I need to)

IMG_5132

ns-rse:
> I am struggling to read the diff clearly in the GitHub view so can't tell exactly what the differences are.

The test returned an `Empty DataFrame` when it should have had three rows of statistics.

<aside>
[Emacs Diff-mode](https://www.gnu.org/software/emacs/manual/html_node/emacs/Diff-Mode.html) helps (although requires a small amount of editing to remove some of the GitHub additional text). Although in essence all it is doing here is colouring the current (lines that begin with `-`) and expected (lines that begin with `+`), for once a screen shot from me...

image

</aside>

ns-rse:
Test passed on re-run, closing though as I've solved this locally (and checked the pages build correctly locally) and made a separate PR #961.

PR #973, opened 2024-10-18
Fixes None tracing outputs attempted to be concatenated

MaxGamill-Sheffield:
Apologies, the previous PR was a tad early, I thought the "merge when ready" would wait for until tests finished...

This should provide a hotfix to the errors and similar like them:
```
disordered_trace_results[str(img)] = disordered_trace_result.dropna(axis=1, how="all")
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'dropna'
```

This occurred because the disabled modules would return `None` instead of an empty dataframe and we expected to have the dataframe function `DataFrame.dropna()` which obvs didn't exist.

This fix should ensure that dataframes are returned by disabled modules.

It also adds another exception I found when forcing the disordered traces step to fail causing KeyErrors in NodeStats and Ordered Tracing as they couldn't find the image name keys of a skeleton to grab.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -7,7 +7,6 @@ import numpy.typing as npt
 import pytest
 
 from topostats.utils import (
-    ALL_STATISTICS_COLUMNS,
     bound_padded_coordinates_to_image,
     convert_path,
     convolve_skeleton,
@@ -184,7 +183,7 @@ def test_get_thresholds_value_error(image_random: np.ndarray) -> None:
 
 def test_create_empty_dataframe() -> None:
     """Test the empty dataframe is created correctly."""
-    empty_df = create_empty_dataframe(ALL_STATISTICS_COLUMNS)
+    empty_df = create_empty_dataframe()
 
     assert empty_df.index.name == "grain_number"
     assert "grain_number" not in empty_df.columns
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -629,6 +629,9 @@ def run_nodestats(  # noqa: C901
             # merge all image dictionaries
             return nodestats_whole_data, resultant_grainstats
 
+        except KeyError as e:
+            LOGGER.info(f"[{filename}] : NodeStats failed {e} - no skeletons found from the Disordered Tracing step.")
+
         except Exception as e:
             LOGGER.info(
                 f"[{filename}] : NodeStats failed - skipping. Consider raising an issue on GitHub. Error: ", exc_info=e
@@ -754,6 +757,16 @@ def run_ordered_tracing(
 
             # merge all image dictionaries
             return ordered_tracing_image_data, resultant_grainstats, ordered_tracing_molstats
+
+        except KeyError as e:
+            LOGGER.info(
+                f"[{filename}] : Ordered Tracing failed {e} - no skeletons found from the Disordered Tracing step."
+            )
+            return (
+                ordered_tracing_image_data,
+                grainstats_df,
+                create_empty_dataframe(column_set="mol_statistics", index="molecule_number"),
+            )
 
         except Exception as e:
             LOGGER.info(
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -492,10 +492,14 @@ def run_disordered_trace(
                 f"[{filename}] : Disordered tracing failed - skipping. Consider raising an issue on GitHub. Error: ",
                 exc_info=e,
             )
-            return {}, grainstats_df, None
+            return (
+                disordered_traces,
+                grainstats_df,
+                create_empty_dataframe(column_set="disordered_tracing_statistics", index="index"),
+            )
 
     LOGGER.info(f"[{filename}] Calculation of Disordered Tracing disabled, returning empty dictionary.")
-    return None, grainstats_df, None
+    return None, grainstats_df, create_empty_dataframe(column_set="disordered_tracing_statistics", index="index")
 
 
 def run_nodestats(  # noqa: C901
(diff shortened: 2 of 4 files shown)


SylviaWhittle (approved):
Very much approve of the different `column_set`s. This makes things really clear IMO. Also love the returning an empty dataframe that is created in the same return statement

ns-rse:
Just looking through this and will be adding a commit to parametrize the tests and have one question which I need to investigate further before asking.

topostats/utils.py
```diff
@@ -269,14 +296,14 @@ def get_thresholds(  # noqa: C901
     return thresholds
 
 
-def create_empty_dataframe(columns: set = ALL_STATISTICS_COLUMNS, index: str = "grain_number") -> pd.DataFrame:
+def create_empty_dataframe(column_set: str = "grainstats", index: str = "grain_number") -> pd.DataFrame:
```
ns-rse, line 299: I realised `index` is a common Python method and will change this to `index_col` to avoid any possible confusion in the work I'm doing to parametrise the tests.

tests/test_utils.py
```diff
@@ -184,7 +183,7 @@ def test_get_thresholds_value_error(image_random: np.ndarray) -> None:
 
 def test_create_empty_dataframe() -> None:
```
ns-rse, line 184: This only tests the `grainstats_df` dataframe is correctly created. I will parametrise the test to check all three possible options (also makes it easier to extend if/when newer empty dataframes are required).

topostats/processing.py
```diff
@@ -625,6 +629,9 @@ def run_nodestats(  # noqa: C901
             # merge all image dictionaries
             return nodestats_whole_data, resultant_grainstats
 
+        except KeyError as e:
```
ns-rse, line 632: :+1:

topostats/processing.py
```diff
@@ -816,7 +837,7 @@ def run_splining(
                         f"[{filename}] : No grains exist for the {direction} direction. Skipping disordered_tracing for {direction}."
                     )
                     splining_grainstats = create_empty_dataframe()
```
ns-rse, line 839: Perhaps worth being explicit here rather than relying on defaults which might change.

topostats/processing.py
```diff
@@ -877,7 +898,7 @@ def run_splining(
             )
             return splined_image_data, grainstats_df, splining_molstats
 
-    return None, grainstats_df, None
+    return None, grainstats_df, molstats_df
```
ns-rse, line 901: If this line is ever reached, which will be the case if `splining_config["run"] == False` then `molstats_df` is whatever has been passed into the function as an argument `molstats_df`. The default for this is `None` and as we've found having `None` is being deprecated in `pd.concat()`.

I've not dug deeply into this but will `molstats_df` always be non-`None` when processing?

Should this perhaps be...

```suggestion
    if molstats_df is not None:
        return None, grainstats_df, molstats_df
    return None, grainstats_df, create_empty_dataframe("splinning_statistics", index="molecule_number")
```

...and the `topostats.utils.COLUMN_SETS` needs another key/value pair for the expected columns?

topostats/processing.py
ns-rse: Now I've checked the context I don't see the point in creating these empty dataframes here either since they are within `if not ordered_tracing_direction_data:` which immediately after creation will `raises ValueError("No grains exist for the {direction} direction")`.

Could we perhaps create the two empty dataframes when they are first instantiated on lines 808 and 809 and move these two lines to there so instead of...

```python
    splining_grainstats = pd.DataFrame()
    splining_molstats = pd.DataFrame()
```

...we have...

```python
    splining_grainstats = create_empty_dataframe(column_set="grainstats", index_col="grain_number")
    splining_molstats = create_empty_dataframe(column_set="mol_statistics", index_col="molecule_number")
```

:thinking:

ns-rse: Tried that, it didn't turn out well because further down there are `merge()` and duplicate columns are introduced.

MaxGamill-Sheffield: Shall we leave them as is then?

topostats/processing.py
MaxGamill-Sheffield: I've stopped it from defaulting to None and instead have the default values similar to `create_empty_dataframe("molecule_statistics", index="molecule_number")`

MaxGamill-Sheffield: Good spot, I forgot about this. I've now added this in at the start so it's clear what happens to the None's

topostats/processing.py
ns-rse: Its a bit more work to change the logic and flow of dataframe generation and subsequent merging.

Welcome to tackle it now but I'd be fine with leaving it for future, its not detrimental.

MaxGamill-Sheffield: Think we'll leave it for the future and just get the fix out asap

ns-rse: Could you log it as an issue then please. I know I'll forget unless its written down.

llwiggins (approved):
This all seems to be working as expected, happy for this to be merged!

llwiggins:
Max already added it in the general tidying PR as he didn’t realise it was
included in this one, so it might already be there?

Dr. Laura Wiggins (she/her)
Post Doctoral Research Associate

Pyne Lab
Department of Materials Science and Engineering
University of Sheffield
Sir Robert Hadfield Building
Mappin Street
Sheffield, S1 3JD

ns-rse:
> Max already added it in the general tidying PR as he didn’t realise it was included in this one, so it might already be there?

I realised I was commenting on the wrong PR, sorry.

PR #974, opened 2024-10-20
Add basic curvature

### LINKED ISSUE #168: Merge curvature calculations into dev

**Additional context**

SylviaWhittle:
This PR adds just a basic angle-turn-per-nm metric, calculated for each molecule of each grain, linear and circular. It is super simple but extensible to meet future needs (I hope).

It will likely need tuning and maybe adding a configurable 1D gaussian filter (since curvature really changes depending on the length scale you're looking for. I've defaulted to angle per nm but maybe in future for different projects, we will want this configurable. Not adding it here yet though as I want feedback from different projects first. 

Works best on high-res data!

Closes #168


SylviaWhittle:
Oh also before merging, we need to discuss & solve the issue of the custom plots I've made. @MaxGamill-Sheffield will likely have a way to do this but currently it dumbly saves the images in the current directory and is just for experimentation.

SylviaWhittle:
Not completely completed but would like high-level feedback before putting effort towards finalising if that's okay 😄

llwiggins:
I'm wondering how much the sign of the angle difference matters, and whether it would be better to just consider the absolute magnitude for most applications? See example below, first plot is with the signed angle difference and the second is using the absolute value of the angle difference. To me, the absolute value makes it more intuitive e.g. dark red regions are highly curved, light regions are not.

Perhaps this could even be configurable if we anticipate different applications where one approach would be better than the other?
32-individual_grain_curvature_grain_19
32-individual_grain_curvature_grain_19

SylviaWhittle:
> I'm wondering how much the sign of the angle difference matters, and whether it would be better to just consider the absolute magnitude for most applications? See example below, first plot is with the signed angle difference and the second is using the absolute value of the angle difference. To me, the absolute value makes it more intuitive e.g. dark red regions are highly curved, light regions are not.
> 
> Perhaps this could even be configurable if we anticipate different applications where one approach would be better than the other? 32-individual_grain_curvature_grain_19 32-individual_grain_curvature_grain_19

Just asked Tom, Libby and Kinga, and they say there is no use in directional curvature. I'd be happy with adding a `np.abs()` for now and removing it in future if we need! I'll add that on Monday 😄

llwiggins:
That sounds good to me. I really like your approach to this, and although the method itself is simple I think it will be really valuable for future projects! I agree with your comment about making the angle per nm part configurable, as I think it may show regions of high/low curvature better if we consider larger neighbourhoods around each point?? 1D gaussian would also be a neat solution for this!

SylviaWhittle:
Will update regtest

tests/test_processing.py
```diff
@@ -558,54 +565,59 @@ def test_check_run_steps(
...
-            False,
-            False,
+            True,  # Filter
+            True,  # Grains
+            True,  # Grainstats
+            False,  # Disordered Tracing
+            False,  # Nodestats
+            False,  # Ordered tracing
+            False,  # Splining
+            False,  # Curvature
```
ns-rse, line 620: Do you think we should have a test that checks if the `Curvature` option is `True` which would require all prior steps?

topostats/measure/curvature.py
```diff
@@ -0,0 +1,175 @@
...
+    float
+        The signed angle difference in radians.
+    """
+    if v1.shape != (2,) or v2.shape != (2,):
+        raise ValueError("Vectors must be of shape (2,)")
+
+    angle = np.arctan2(v2[1], v2[0]) - np.arctan2(v1[1], v1[0])
+    if angle > np.pi:
+        angle -= 2 * np.pi
+    elif angle < -np.pi:
```
ns-rse, line 39: What if `angle == np.pi` (rare but you never know!)? I can see `angle` is then returned without subtracting/adding `2 * np.pi` but is that the desired behaviour?

ns-rse (commented):
Looks good, a couple of minor comments in-line.

I've added issues so I don't forget to include this and splinning into the entry points (#1009 and #1010) and have an issue to document how to do this when further modules are added.

llwiggins (approved):
Looks great to me! I think the change to using the absolute value rather than directional works really well and emphasises areas of high curvature better for users, particularly in the curvature plot that is output where it is harder to see local curvature changes.

Happy for this to be merged 👍

topostats/measure/curvature.py
SylviaWhittle: Very thorough reading there! But I believe this is intended. The domain for the angles is designed to be `[-np.pi, np.pi]` inclusive.

llwiggins (approved):

PR #977, opened 2024-10-21
tests(utils): parameterise test_create_empty_dataframe()

ns-rse:
- Parameterises `test_utils.py::test_create_empty_dataframe()` so that all three possible empty dataframes are now tested.
- Renames `create_empty_dataframe()` parameter `index` > `index_col` to avoid confusion with Python and Pandas `index` methods.
- Explicitly calls `create_empty_dataframe()` with the `column_set` and `index_col` in each instance rather than relying on defaults when an empty `grainstats_df` is required.
- Further at the end of `process_scan()` empty dataframes of all types were incorrectly generated as no types were specified for any, these now generate the correct type for each required data frame.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -181,14 +181,30 @@ def test_get_thresholds_value_error(image_random: np.ndarray) -> None:
         get_thresholds(image=image_random, threshold_method="mean", **THRESHOLD_OPTIONS)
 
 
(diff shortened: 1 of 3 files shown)


MaxGamill-Sheffield (approved):
Aye looks good and nice and simple 👍

PR #980, opened 2024-10-23
feature: Print 'art' above completion message

ns-rse:
Prints "TopoStast" using the [art]() package and the "twisted" font which I felt was appropriate for DNA!


SylviaWhittle:
This is the output now btw:
image

SylviaWhittle (approved):
Looks good 👍 runs 👍 can't see how this could break stuff 👍

PR #983, opened 2024-10-25
Optimized Area thresholding

AuroVarat:
This modification enhances the area thresholding function, significantly improving processing speed, especially for large images.

### DIFF
--- a/topostats/grains.py
+++ b/topostats/grains.py
@@ -360,22 +360,34 @@ class Grains:
         if lower_size_limit is None:
             lower_size_limit = 0
         # Get array of grain numbers (discounting zero)
-        uniq = np.delete(np.unique(image), 0)
-        grain_count = 0
+        # uniq = np.delete(np.unique(image), 0)
+        # grain_count = 0
         LOGGER.debug(
             f"[{self.filename}] : Area thresholding grains | Thresholds: L: {(lower_size_limit / self.pixel_to_nm_scaling**2):.2f},"
             f"U: {(upper_size_limit / self.pixel_to_nm_scaling**2):.2f} px^2, L: {lower_size_limit:.2f}, U: {upper_size_limit:.2f} nm^2."
         )
-        for grain_no in uniq:  # Calculate grian area in nm^2
-            grain_area = np.sum(image_cp == grain_no) * (self.pixel_to_nm_scaling**2)
-            # Compare area in nm^2 to area thresholds
-            if grain_area > upper_size_limit or grain_area < lower_size_limit:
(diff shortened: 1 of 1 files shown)


ns-rse:
Thanks for the Pull Request @AuroVarat I'll take a look through this on Monday.

In the mean time would you be able to address the failing tests please?

If you're not familiar with tests the [software development](https://afm-spm.github.io/TopoStats/main/contributing.html#software-development) section of our docs has some advice. You'll want to `pip install -e .[dev,tests.docs]` and ideally install the `pre-commit` hooks with `pre-commit install`. We use the `pytest-testmon` extension and so you'd have to run the tests once to check everything is working with `pytest --testmon` and then when you make commits only the tests that are affected will run.

Note also that [`pre-commit.ci`](https://pre-commit.ci) has modified this branch by fixing some minor linting errors so you may want to `git pull` to get those changes (although if you install pre-commit locally it will make them before you can make your next commit again).

AuroVarat:
Hi, I am not familiar with tests but I managed to go through the steps and do the "pytest --testmon". It fails 8 tests, but it fails 8 tests regardless of my changes ( I did the same test by running it on the current main branch). Maybe I am doing something wrong, I will look at it when I have more time. 

But if you see my changes, they are minimal and should not result in this failure. The code seems to be working fine when I am doing my analysis. 

Thanks.

AuroVarat:
@ns-rse Hi Neil, Hope you are well. Were you able to merge it?

ns-rse:
Hi @AuroVarat 

Apologies I've not had time to investigate why the tests are failing as I'm not full time on this project and have had various other tasks to work through.

The [logs](https://github.com/AFM-SPM/TopoStats/actions/runs/11817416180/job/32927571522?pr=983) show that there are two elements that are different in the revised code. How big an issue that might be isn't something I can really answer but we have tests in place to ensure that changes under the hood don't change the expected results.  You could get more information on where these have changed running the specific test with...

```bash
pytest -vv tests/test_grains.py::test_find_grains
```

...although there may be other places where tests fail as we use the `-x` flag which exits the test suite on the first failure that is encountered so it would be worth running the whole test suite (this can be done in parallel if you install all of the `dev` optional dependencies in your virtual environment with `pytest -n #` where `#` is the number of cores to use which will depend on your system and what else you want to do whilst its running).


If you are new to testing you may find the following tips on [Contributing](https://afm-spm.github.io/TopoStats/main/contributing.html#software-development) useful as they show how to get setup with a development environment and install the pre-commit hooks that we have in place which performs various linting.

I've got some failing tests of my own to work through but feel free to message if you've any questions.

PR #981, opened 2024-10-23
Add support for partial config files

SylviaWhittle:
As discussed in the recent TopoStats catchup, the config file is about to explode with parameters. This will result in a rather cumbersome config file for users which will be intimidating.

After fielding a few options to the experimentalists, it was decided that the least bad way forward might be to add support for partial config files, where any missing fields are added in from the `default_config.yaml`.

This PR is attempting to do just that. Though I'm sure I've forgotten something.

Oh also this PR is entirely **type safe** (I think?, at least according to mypy!) It brings some type safety to the beginning of `run_topostats.py` and marks my first foray into generic typing!

PR #985, opened 2024-10-28
refactor(entry_points): Removing legacy entry points

ns-rse:
As a step towards #517 this commit removes the legacy entry points and their associated tests.


SylviaWhittle (approved):
Always very happy to remove code!

ns-rse:
Thanks @SylviaWhittle merged so you're good to rebase onto `main` now for #981

PR #992, opened 2024-11-05
Hariborings vetting to main

SylviaWhittle:
- Adds vetting steps for grains in `find_grains` after the unet improvement.
  - Vet number of regions in classes (eg class 3 needs 2 regions)
  - Vet sizes of classes (eg class 2 needs to have 25 nm^2)
  - Convert classes to other classes if nearby (eg convert all but largest region of class 2 into class 3 for regions of class 2 that touch class 3)
  - Keep largest region only for particular classes (eg keep largest regions for classes 2 and 3)
  - Vet based on class connection points (eg keep only grains where class 1 is connected to class 2 by between 3 and 5 connection regions)
- Adds the ability to merge classes into an extra tensor class (appended to the end) (eg merge classes (2, 3, 4) and (5, 6) -> produces two extra classes, merged(2,3,4) and merged(5,6).
- Adds the ability to choose which class `grainstats` runs on. (eg run grainstats on class 2) currently only allows a single class.


SylviaWhittle:
Tests working it seems (except windows ofc)

tests/test_grains.py
```diff
@@ -423,6 +492,7 @@ def test_find_grains(
     smallest_grain_size_nm2: int,
     remove_edge_intersecting_grains: bool,
     expected_grain_mask: npt.NDArray[np.int32],
+    expected_labelled_regions: npt.NDArray[np.int32],
```
ns-rse, line 495: The types here are `np.int32` but in the above parametrisation the arrays are explicitly set to be `np.int64`.

tests/test_grains.py
```diff
@@ -480,7 +555,7 @@ def test_find_grains(
                         [
                             [1, 1, 1, 1, 1, 1, 1, 1, 1],
                             [1, 0, 0, 0, 1, 0, 1, 0, 1],
-                            [1, 0, 0, 0, 1, 0, 1, 1, 1],
+                            [1, 0, 1, 0, 1, 0, 1, 1, 1],
```
ns-rse, line 558: Any idea why there is a this single pixel change?

topostats/default_config.yaml
```diff
@@ -53,6 +53,13 @@ grains:
     grain_crop_padding: 2 # Padding to apply to the grain crop bounding box
     upper_norm_bound: 5.0 # Upper bound for normalisation of input data. This should be slightly higher than the maximum desired / expected height of grains.
     lower_norm_bound: -1.0 # Lower bound for normalisation of input data. This should be slightly lower than the minimum desired / expected height of the background.
+  vetting:
```
ns-rse, line 56: Would it be possible to include these descriptions in the `docs/config.md` please?

topostats/grains.py
```diff
@@ -610,7 +655,15 @@ def improve_grain_segmentation_unet(
         # You may also get an error referencing a "group_1" parameter, this is discussed in this issue:
         # https://github.com/keras-team/keras/issues/19441 which also has an experimental fix that we can try but
         # I haven't tested it yet.
-        unet_model = keras.models.load_model(unet_config["model_path"], custom_objects={"mean_iou": mean_iou})
+
+        # Important for when loading the model inievitably fails on users' machines.
```
ns-rse, line 659: How does this stack up against the recent changes which reduce logging output?

Could we perhaps have these reported as `LOGGER.error()` within a `try:... except: ...` when the failure occurs?

topostats/grains.py
```diff
@@ -682,7 +735,7 @@ def improve_grain_segmentation_unet(
                 # Grab the unet mask for the class
                 unet_predicted_mask_labelled = morphology.label(predicted_mask[:, :, class_index])
                 # Keep only the largest object in the grain crop (needs to be configurable in future)
-                unet_predicted_mask_labelled = Grains.keep_largest_labelled_region(unet_predicted_mask_labelled)
+                # unet_predicted_mask_labelled = Grains.keep_largest_labelled_region(unet_predicted_mask_labelled)
```
ns-rse, line 738: Can this be removed if no longer needed?

topostats/grains.py
```diff
@@ -742,3 +795,741 @@ def keep_largest_labelled_region(
...
+        grain_mask_tensor : npt.NDArray
+            Multi class grain mask tensor tensor of shape (N, N, C).
+
+        Returns
+        -------
+        npt.NDArray
+            Combined binary mask of all but the background class (:, :, 0).
+        """
+        assert len(grain_mask_tensor.shape) == 3, f"Tensor not 3D: {grain_mask_tensor.shape}"
+        # Convert to binary in case there are multiple hits in the same pixel. We dont want to have 2s, 3s etc
```
ns-rse, line 815: Feels like this could either be removed or moved to the doc-string to provide users with more information on the API documentation side of things.

topostats/grains.py
```diff
@@ -742,3 +795,741 @@ def keep_largest_labelled_region(
...
+        ----------
+        grain_mask_tensor : npt.NDArray
+            3-D Numpy array of grain mask tensor.
+
+        Returns
+        -------
+        dict
+            Dictionary of bounding boxes indexed by grain number.
+        """
+        # Get the flattened mask
```
ns-rse, line 837: I must be getting old (or slightly better at reading code) as I feel these comments are extraneous given we have expressive variable and method names which make it clear what is happening at each step.

topostats/grains.py
```diff
@@ -742,3 +795,741 @@ def keep_largest_labelled_region(
...
+        return grain_mask_tensor.astype(bool)
+
+    @staticmethod
+    def vet_class_sizes(
+        grain_mask_tensor: npt.NDArray,
+        pixel_to_nm_scaling: float,
+        class_size_thresholds: list[list[int, int, int]],
+    ) -> npt.NDArray:
+        """
+        Vet the sizes of the classes in an image tensor.
```
ns-rse, line 889: I think "Vet" is a poor description of what this method is doing.

```suggestion
        Remove grains from mask tensor below a given threshold.
```

topostats/grains.py
```diff
@@ -742,3 +795,741 @@ def keep_largest_labelled_region(
...
+        return grain_mask_tensor
+
+    @staticmethod
+    def vet_class_sizes_single_grain(
+        single_grain_mask_tensor: npt.NDArray,
+        pixel_to_nm_scaling: float,
+        class_size_thresholds: list[list[int, int, int]] | None,
+    ) -> tuple[npt.NDArray, bool]:
+        """
+        Vet the sizes of the classes in a single grain mask tensor.
```
ns-rse, line 957: Ditto, need to explain what "Vet" means here.

topostats/grains.py
```diff
@@ -742,3 +795,741 @@ def keep_largest_labelled_region(
...
+
+        return grain_crops, bounding_boxes, padding
+
+    @staticmethod
+    def vet_numbers_of_regions_single_grain(
+        grain_mask_tensor: npt.NDArray,
+        class_region_number_thresholds: list[list[int, int, int]] | None,
+    ) -> tuple[npt.NDArray, bool]:
+        """
+        Vet the number of regions in a grain mask tensor of a single grain, ignoring the background class.
```
ns-rse, line 1096: Ditto, what criteria is being checked/vetted.

topostats/grains.py
```diff
@@ -742,3 +795,741 @@ def keep_largest_labelled_region(
...
+        return grain_mask_tensor, True
+
+    @staticmethod
+    def convert_classes_to_nearby_classes(
+        grain_mask_tensor: npt.NDArray,
+        classes_to_convert: list[tuple[int, int]],
+        class_touching_threshold: int = 1,
+    ) -> npt.NDArray:
+        """
+        Convert all but the largest regions of class A into class B provided that the class A region touches class B.
```
ns-rse, line 1158: What are "class A" and "class B" and how do they differ?

topostats/processing.py
```diff
@@ -34,7 +34,6 @@
 # pylint: disable=too-many-locals
 # pylint: disable=too-many-statements
 # pylint: disable=too-many-nested-blocks
-# pylint: disable=too-many-positional-arguments
```
ns-rse, line 37: :grin:

topostats/default_config.yaml
```diff
@@ -53,6 +53,13 @@ grains:
...
+    class_region_number_thresholds: null # Class region number thresholds eg 1: [2, 4], 2: [1, 1]
```
ns-rse, line 57: `validation.py` indicates that these should be lists of lists of three integers, should probably have that reflected in the comment and the documentation.

ns-rse (changes requested):
First pass through with some general comments.

I've not used the libraries and have no idea what is actually being done so its mostly all on style/structure/documentation rather than a deeper appreciation and feedback on what the additions are doing.

tests/test_grains.py
SylviaWhittle: Yeah in my new code, I add an `update_background_class` call at the end of `find_grains` which ensures that the background mask is correct. Previously I did not do this, and the background mask erroneously did not add that background pixel back in. This change in the test reflects the updated behavior.

topostats/default_config.yaml
SylviaWhittle: Yep! Just wanted overall high-level feedback before adding finishing touches (to prevent doing the finishing touches twice if someone requested major changes)

topostats/grains.py
SylviaWhittle: Good idea, will do!

tests/test_grains.py
SylviaWhittle: Solved 😃

tests/test_grains.py
ns-rse: Ok cool, always good to improve things :+1:

topostats/grains.py
```diff
@@ -656,12 +656,13 @@ def improve_grain_segmentation_unet(
...
 
-        # Important for when loading the model inievitably fails on users' machines.
-        LOGGER.info(f"Python executable: {os.sys.executable}")
-        LOGGER.info(f"Keras version: {keras.__version__}")
-        LOGGER.info(f"Model path: {unet_config['model_path']}")
-
-        unet_model = keras.models.load_model(unet_config["model_path"], compile=False)
+        try:
+            unet_model = keras.models.load_model(unet_config["model_path"], compile=False)
+        except Exception as e:
```
ns-rse, line 661: This will likely have fallen foul of `pylint`'s complaint about broad exceptions.

Are there any examples of the types of exceptions that are raised that we could use?

topostats/grains.py
SylviaWhittle: I've attempted to remedy the obscure descriptions, see what you think now, happy to iterate again.

llwiggins:
Thank you for working on this @SylviaWhittle - I've really enjoyed playing around with the new vetting parameters in the config! A couple of comments below:

- The description next to the `class_region_number_thresholds` parameter doesn't make it clear that this needs to be a list of lists, with double use of square brackets e.g. [[1, 2, 4], [2, 1, 1]]. The other parameters do make this clear so I would just copy how they're written for consistency. 
- For the `class_size_thresholds`, it may be best to add to the comment next to it that users can enter just a lower or an upper bound by leaving one of the list entries empty e.g. [[2,300,]] if they only want to vet regions of class 2 that are < 300 without an upper threshold. 
- It could also be worth adding units to the `class_size_thresholds` parameter comment so that users know which values to enter for thresholding
- `vet_class_sizes_single_grain()` sets the whole mask to background if a class objects fall outside of their expected size, would it be better to have the option to set these regions to another class? E.g. if class 2 is smaller than expected, set this to class 1? This may help us retain a lot of grains that we're currently throwing away. (Apologies if this can already be done using a clever combination of the vetting steps).

I think all of this will require quite comprehensive documentation as it does require some playing around with but is very powerful once parameters are optimised! Visualisations of the different vetting steps and the problems they fix would be really useful. 😄

topostats/grains.py
```diff
@@ -1019,10 +1019,10 @@ def get_individual_grain_crops(
     @staticmethod
     def vet_numbers_of_regions_single_grain(
         grain_mask_tensor: npt.NDArray,
-        class_region_number_thresholds: list[list[int, int, int]] | None,
+        class_region_number_thresholds: list[tuple[int, int, int]] | None,
```
ns-rse, line 1022: I often feel referring to the type in variable names is excessive and unnecessary, in `vet_class_sizes_single_grain` the option is simply `class_size_threholds` and I think we should be consistent throughout as its easier to understand code (at least I find it easier) when things that have the same meaning have the same name.

```suggestion
        class_region_thresholds: list[tuple[int, int, int]] | None,
```

Obviously use within the function needs updating but simple with the language server functionality to replace (although in my experience it rarely updates the docstring so that will need to be done manually).

topostats/grains.py
```diff
@@ -1080,11 +1080,19 @@ def vet_numbers_of_regions_single_grain(
...
+        classes_to_convert: list[tuple[int, int]] | None,
         class_touching_threshold: int = 1,
     ) -> npt.NDArray:
         """
-        Convert all but the largest regions of class A into class B provided that the class A region touches class B.
+        Convert all but the largest regions of one class into another class provided the former touches the latter.
+
+        Specifically, it takes a list of tuples of two integers (dubbed class A and class B). For each class A, class B
+        pair, it will find the largest region of class A and flag it to be ignored. Then for each non-largest region of
+        class A, it will check if it touches any class B region (within the class_touching_threshold distance). If it
```
ns-rse, line 1091:
```suggestion
        class A, it will check if it touches any class B region (within the ``class_touching_threshold`` distance). If it
```

topostats/grains.py
SylviaWhittle: If I'm understanding you correctly, this variable name is intended to be the same style as `class_size_thresholds`, as in we are thresholding numbers of regions, hence `class_*region_number*_thresholds` but obviously this isn't very apparent.

I think `class_region_thresholds` would be too ambiguous, what are we thresholding about the class regions? Their size, height, number?

Any thoughts?

SylviaWhittle:
> Thank you for working on this @SylviaWhittle - I've really enjoyed playing around with the new vetting parameters in the config! A couple of comments below:
> 
> * The description next to the `class_region_number_thresholds` parameter doesn't make it clear that this needs to be a list of lists, with double use of square brackets e.g. [[1, 2, 4], [2, 1, 1]]. The other parameters do make this clear so I would just copy how they're written for consistency.
> * For the `class_size_thresholds`, it may be best to add to the comment next to it that users can enter just a lower or an upper bound by leaving one of the list entries empty e.g. [[2,300,]] if they only want to vet regions of class 2 that are < 300 without an upper threshold.
> * It could also be worth adding units to the `class_size_thresholds` parameter comment so that users know which values to enter for thresholding
> * `vet_class_sizes_single_grain()` sets the whole mask to background if a class object falls outside of their expected size, would it be better to have the option to set these regions to another class? E.g. if class 2 is smaller than expected, set this to class 1? This may help us retain a lot of grains that we're currently throwing away. (Apologies if this can already be done using a clever combination of the vetting steps).
> 
> I think all of this will require quite comprehensive documentation as it does require some playing around with but is very powerful once parameters are optimised! Visualisations of the different vetting steps and the problems they fix would be really useful. 😄

I believe I've addressed all comments except your last one. I think I'd prefer to defer until this feature is explicitly needed before implementing it? Unless you have an imminent use case?

llwiggins:
> > Thank you for working on this @SylviaWhittle - I've really enjoyed playing around with the new vetting parameters in the config! A couple of comments below:
> > 
> > * The description next to the `class_region_number_thresholds` parameter doesn't make it clear that this needs to be a list of lists, with double use of square brackets e.g. [[1, 2, 4], [2, 1, 1]]. The other parameters do make this clear so I would just copy how they're written for consistency.
> > * For the `class_size_thresholds`, it may be best to add to the comment next to it that users can enter just a lower or an upper bound by leaving one of the list entries empty e.g. [[2,300,]] if they only want to vet regions of class 2 that are < 300 without an upper threshold.
> > * It could also be worth adding units to the `class_size_thresholds` parameter comment so that users know which values to enter for thresholding
> > * `vet_class_sizes_single_grain()` sets the whole mask to background if a class object falls outside of their expected size, would it be better to have the option to set these regions to another class? E.g. if class 2 is smaller than expected, set this to class 1? This may help us retain a lot of grains that we're currently throwing away. (Apologies if this can already be done using a clever combination of the vetting steps).
> > 
> > I think all of this will require quite comprehensive documentation as it does require some playing around with but is very powerful once parameters are optimised! Visualisations of the different vetting steps and the problems they fix would be really useful. 😄
> 
> I believe I've addressed all comments except your last one. I think I'd prefer to defer until this feature is explicitly needed before implementing it? Unless you have an imminent use case?

Looks great, thanks for making those changes so quickly! An imminent use case for the final point would be crossings being frequently misclassified as protein. More often than not, the crossings have an area that is smaller than expected of the protein and so it would be useful to have the option to relabel all protein objects that do not meet a certain size threshold as DNA. Could there be another parameter in the config that takes a tuple of the form [class1, class2, threshold] to convert any class1 objects below a certain size threshold to class2?

The function to do this would then be similar to `vet_class_sizes_single_grain()` but setting small class1 regions to class2 right?

topostats/grains.py
SylviaWhittle: Personally I find that rather abstract - how about `class_region_quantity_thresholds`?

topostats/validation.py
```diff
@@ -187,6 +187,14 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 "lower_norm_bound": float,
             },
             "vetting": {
+                "class_conversion_size_thresholds": Or(
+                    None,
+                    # List of lists of 3 integers and 2 integers
+                    [[[int, int, int], [int, int]]],
+                    error="Invalid value in config for 'grainstats.vetting.class_conversion_size_thresholds', this"
```
ns-rse, line 194: Without a space at the end of the line the concatenation of strings will end up reading `thisneeds`...

```suggestion
                    error="Invalid value in config for 'grainstats.vetting.class_conversion_size_thresholds', this "
```

topostats/validation.py
```diff
@@ -187,6 +187,14 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
...
+                    "needs to be a list of lists of 3 integers and 2 integers - [class number, class to convert to"
```
ns-rse, line 195: Ditto...

```suggestion
                    "needs to be a list of lists of 3 integers and 2 integers - [class number, class to convert to "
```

ns-rse (commented):
Very minor formatting issue on the `class_conversion_size_thresholds` validation.

llwiggins:
Thanks @SylviaWhittle for adding `convert_classes_when_too_big_or_small()` to the set of vetting functions. I've tested this out and it really nicely resolves the crossings vs. protein segmentation issue that we keep running into and will be super useful for our current projects. Another shoutout for the comprehensive tests that you have written for these functions. 😄 

I'm happy for this PR to be merged! 👍

llwiggins (approved):

tests/test_grains.py
```diff
@@ -914,3 +989,2290 @@ def test_keep_largest_labelled_region(
...
+    result_grain_mask_tensor = Grains.merge_classes(grain_mask_tensor, classes_to_merge)
+
+    np.testing.assert_array_equal(result_grain_mask_tensor, expected_result_grain_mask_tensor)
+
+
+@pytest.mark.parametrize(
+    (
+        "grain_mask_tensor",
+        "pixel_to_nm_scaling",
+        "class_size_thresholds",
```
ns-rse, line 2802: Tests fail as there are only nine parameter names here but there are ten parameters in each `pytest.param()`.

L:eyes:king at the arguments for the function and the comments within the parameters themselves I think its this...

```suggestion
        "class_conversion_size_thresholds",
        "class_size_thresholds",
```

ns-rse (approved):
Ready to go! :rocket:

Issue #927, opened 2024-10-01
Tests hang on `maxgamill-sheffield/topology` branch


ns-rse:
After merging #897 I checked to see what fails on `maxgamill-sheffield/topology` as I'm working my way through #850 and this branch is due to be merged to `maxgamill-sheffield/800-better-tracing`.

I found though that whether I ran the tests on a single thread or multiple the test suite never actually starts, it sits there using 100% CPU of the number of cores specified and I have to manually kill it.

## `git bisect`

To investigate this I've employed a [`git bisect`](https://git-scm.com/docs/git-bisect) (using [Magit's bisect](https://magit.vc/manual/magit/Bisecting.html)) running `pytest -x` on each bisection.

### Bisect Log

```
Bisect Log (8)
git bisect start 'HEAD' '60c9d1636'
263f07cf bad Merge pull request #897 from AFM-SPM/SylviaWhittle/topology_tests
60c9d163 good re-add nodestats to config.md

git bisect bad c15f390d453eeedc4249bf9e17ab29255ff7afe6
c15f390d bad Fix pair_odd_branches not present in class constructor for test_connect_nodes_nearest

git bisect bad a34e5135a222de39d585deec89c074f21ffed9b8
a34e5135 bad removes double dilation away from nodes so unpaired branches finish away from the crossing branches for tracing

git bisect bad 5f279e1d279859ad1452f599c69c681f056a8ed1
5f279e1d bad got writhe calc working - i think

git bisect bad 4b7d25813e48073099d9b9e9dcd7b02255f13a19
4b7d2581 bad ordered traces adds topology & outputs molstats & splining takes molstats

git bisect good 435f92832708ae322d7d076c0b0022c327aea44d
435f9283 good tidy up new pruning method

git bisect good 560d0b19d86f38969277002af32cca08920c0746
560d0b19 good dis trace pruning re-structured

4b7d25813e48073099d9b9e9dcd7b02255f13a19 is the first bad commit
```

### Bad Commit

This identified the bad commit as...

```
4b7d25813e48073099d9b9e9dcd7b02255f13a19 is the first bad commit
commit 4b7d25813e48073099d9b9e9dcd7b02255f13a19
Author: Max Gamill <mcgamill1@sheffield.ac.uk>
Date:   2024-09-03 17:20:05 +0100
    ordered traces adds topology & outputs molstats & splining takes molstats
 topostats/processing.py              | 133 +++++++++--------
[...]

PR #995, opened 2024-11-11
package: Unpin topoly version

### LINKED ISSUE #994: Unpin dependency on topoly==1.0.2
We should soon be able to unpin the dependency on `topoly==1.0.2` as the authors have released a wheel of `topoly==1.0.4` for Windows and version for Apple ARM architecture is [in the offing](https://github.com/AFM-SPM/TopoStats/issues/927#issuecomment-2458194728).

ns-rse:
Wheels are now available for all OS's and architectures (thanks @prubach :+1: )

Closes #994


ns-rse:
Rebase this branch after #988 has been merged to resolve the issues with `pytest-regtest` under OSX.

ns-rse:
I've manually updated the necessary files so that the change in nomenclature used by `topoly` to describe the connection types (? I don't even know what they are describing!) has changed from `4^2_1` > `L4a1` and so forth.

Still feel we should be using `pytest-regtest` to address these regression tests as its another thing to have to explain/teach how to do this manually for people who come in the future to work on the code base.

MaxGamill-Sheffield:
Do we also need to pin numpy < 2 from Pawel's comment: "There seems to be some incompatibility with numpy 2.x. We built the 1.0.4 version before numpy 2.x was out and there are no issues with older numpy versions"

ns-rse:
> Do we also need to pin numpy < 2 from Pawel's comment: "_There seems to be some incompatibility with numpy 2.x. We built the 1.0.4 version before numpy 2.x was out and there are no issues with older numpy versions_"

Full [comment](https://github.com/AFM-SPM/TopoStats/issues/927#issuecomment-2468367367) for context (bold emphasis is mine)...

> BTW while I was investigating that matter I found another issue when using certain functions in Topoly (mostly GLN). There seems to be some incompatibility with numpy 2.x. We built the 1.0.4 version before numpy 2.x was out and there are no issues with older numpy versions but since I didn't limit the version, when you now create a new virtual environment numpy 2.x is by default installed on newer versions of python. I have no clue about the cause as it results in some segfaults **but only with certain input data**, even though we don't use numpy data types etc. in the compiled code. I'm considering restricting the numpy versions for now but need to investigate it further.

Until we get bitten by whatever the problem is (and we don't have enough details to know what that is) I'd say no, not at the moment[^1]. 

Numpy 2.0.0 was a significant release, hence the major semantic version change with breakages to be expected and was made 5 months ago. There have been minor revisions since (current stable version is 2.1.3 released 2024-11-02). Long term keeping up-to-date with revisions is important otherwise you end up in the situation when I first came to work on TopoStats with Docker containers running unsupported software.

For that reason I'm not a fan of pinning version dependencies for long periods, it kicks the can of fixing whatever is broken further down the line.

Further, and perhaps more fundamentally, if `topoly` do decide to pin the Numpy version then I would expect `pip` to resolve that for us and install the correct version of Numpy and other packages so that things "Just Work(TM)".

[^1]: Perhaps there is a hint as maybe explicit data types are required in some instances.

PR #996, opened 2024-11-12
refactor(run_modules): Move common setup to individual functions

### LINKED ISSUE #998: Remove references to run_topostats from documentation
#997 highlighted that in the `README.md` there is reference to `run_topostats` which was removed in #985. This needs removing but also need to remove all references under `docs/` to `run_topostats`.

```bash
❱ grep run_topostats docs/*.md   
```

...doesn't return any hits but will have to manually check.

ns-rse:
Each `run_<module>()` will still need to parse configuration and load files.

Rather than duplicate the code that does so I've broken out the functionality into a few small private functions (nothing is private in Python land though) so they can be re-used across the `run_<module>()` functions.

Introduces a simple test of logging level, must other functionality already has tests in place.

Aiming to keep changes and PRs small to aid reviewing.

Closes #998


ns-rse:
Realised the `docs/usage.md` needed updating and have done so (these documentation changes should have been in my last PR for this work :disappointed_relieved: ).

This should now be ready for review and merging.

SylviaWhittle:
> Realised the `docs/usage.md` needed updating and have done so (these documentation changes should have been in my last PR for this work 😥 ).
> 
> This should now be ready for review and merging.

Don't worry about it, I forget to update docs and configuration all the time. Would be good to have a little automated reminder when I create a PR to update the docs 😄

SylviaWhittle (approved):
This is great, and tidies things up immensely @ns-rse, thank you!

This clearly leads the way to the swiss-army-knife entry points in an elegant fashion.

One concern I have which is not to do with this PR exactly, but the refactor, is that talking to the experimentalists, it would seem that nobody wishes to / intends to use command line arguments to override config options. If so, the ability to override might be not needed, but this can be discussed later.

ns-rse:
> One concern I have which is not to do with this PR exactly, but the refactor, is that talking to the experimentalists, it would seem that nobody wishes to / intends to use command line arguments to override config options. If so, the ability to override might be not needed, but this can be discussed later.

I'd like to think that TopoStats will see wider adoption outside of the immediate research group and others might use these options :crossed_fingers: 

Maybe when there is a need to experiment with settings on the fly to find optimal parameters for unusual samples/scans that haven't been considered in the development pipeline this will be of more use.

The structure is starting to take shape though, just need to write all the `run_<module>` functions (one at a time!).

PR #999, opened 2024-11-14
Added average grain width measurement to disordered tracing

tcatley:
Added function `calculate_dna_width()` into class `disorderedTrace` which uses the smoothed mask and the pruned trace to calculate the average width of the mask using a distance transform. Outputs a float value of the width in metres to the `all_statistics.csv`.


topostats/tracing/disordered_tracing.py
```diff
@@ -254,6 +255,17 @@ def smooth_mask(
         LOGGER.debug(f"[{self.filename}] : smoothing done by dilation {dilation_iterations}")
         return self.re_add_holes(grain, dilation, holearea_min_max)
 
+    @staticmethod
+    def calculate_dna_width(smoothed_mask: npt.NDArray, pruned_skeleton: npt.NDArray, px2nm: float = 1) -> float:
+        """
+        Calculate the average width in metres of the DNA using the trace and mask.
```
ns-rse, line 261: Needs a more complete docstring (one of the reasons pre-commit checks are failing).

I've switched to the more explicit `mean` to avoid ambiguity with `mode` or `median`. More generally though this could be an option to have either the mean or the median width returned. :thinking: 

```suggestion
        Calculate the mean width in metres of the DNA using the trace and mask.
        
        Parameters
        ----------
        smoothed_mask : npt.NDArray
            Smoothed mask to be measured.
        pruned_skeleton : npt.NDArray
            Pruned skeleton.
        px2nm : float
            Scaling of pixels to nanometres.
            
        Returns
        -------
        float
            Width of DNA in metres.
```

ns-rse (changes requested):
Thanks for tackling this @tcatley appreciated.

The tests are failing because you've added in the `grain_width` column to the resulting output files.

In order to update the regression test files you will have to install the `tests` dependencies and whilst at it may as well install the `dev` dependencies then you'll have the ability to run `pre-commit` locally.

```bash
pip install -e .[tests,dev]
```

To install `pre-commit` you can then...

```bash
pre-commit install
```

To update the test files so they include the additional field you then...

```bash
pytest --regtest-reset tests/test_processing.py
```

Ideally we should have a test in place to check the method returns the correct values in isolation from the global regression test.

This involves adding a test to `tests/tracing/test_disordered_tracing.py` but if you look through you will see many are marked as skipped as they are awaiting being written.

I can write this up in the existing issue (or create a new one) if you don't fancy having a stab at it.

ns-rse:
Thanks for the quick update/fixes @tcatley almost there but your second commit #[09e10d1](https://github.com/AFM-SPM/TopoStats/pull/999/commits/09e10d13b7cf335f4048b5acae5bed6d506156e4) change the column name to `grain_width_mean` whilst your first commit updated the regression test files with `grain_width` as the new column.

Thus you need to run the regeneration of the regression test files again. We can avoid duplicating commit messages by ammending and force pushing...

```bash
pytest --regtest-reset tests/test_processing.py
git add -u
git commit --amend
git push --force
```

ns-rse:
Hi @tcatley,

Thanks for addressing the changes. Unfortunately the tests still fail, but in a different location (this is progress!!! :grin: ).

Here its `tests/tracing/test_disordered_tracing.py::test_trace_image_disordered` that fails because again we have a dataframe with new columns.

Unfortunately these checks are made outside of the `pytest-regtest` frame work so we can't just run `pytest --regtest-reset tests/tracing/test_disordered_tracing.py::test_trace_image_disordered`. Rather we have to uncomment a section of code that writes the target files to disk, run the test to update the target files and include those. This is a bit more involved and for that reason I'll undertake the changes on a separate branch which we can merge back into this branch once I've fixed them.

tcatley:
Great, thanks @ns-rse! Let me know if you need anything else from me

ns-rse:
It was a bit more involved than I thought @tcatley and involved another set of regression tests being updated and rebasing onto existing branches that are out for pull requests (#995) which fix the tests in light of a newer version of `topoly`.

#1001 looks big but that is a consequence of "rebasing" which brings the changes from the other branches into yours.

Once #1001 is merged into this we should wait for #995 to be merged to `main` first (in essence this PR now branches from #995 and my #1001 branches from this #999, for those interested an article I found useful on this is [Using stacked pull requests in GitHub - LogRocket Blog](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)).

ns-rse:
For other reviewers...

I split the failing test into two, one for the nested dictionaries that are loaded from pickles and one for comparing dataframes via `pytest-regtest`.

I did some more reading on [syrupy](https://syrupy-project.github.io/syrupy/) and think it might be a viable alternative to replace `pytest-regtest` as tests inherit/have the `snapshot` function passed in which is used to compare against the returned value of whatever is being tested. These are updated with the command line flag `--snapshot-update` (akin to `--regtest-reset`) but objects are stored as serialised objects rather than requiring to be `print()` and written to file.

MaxGamill-Sheffield:
@tcatley already had a simple test which I've quickly implemented and expanded slightly.
Might need a new reviewer as I've contributed now.

tests/tracing/test_disordered_tracing.py
```diff
@@ -1584,3 +1584,179 @@ def test_pad_bounding_box(array_shape: tuple, bounding_box: list, pad_width: int
...
+                    [0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0],
+                    [0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
+                    [0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                ],
+                dtype=np.bool_,
+            ),
+            1,
+            3.8666666666666667,
+            id="Jagged square",
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: So this might be seen if the pruning is poor and thought it a good example of something lower as the extended corners reach into the lower distance values, bringing the average down

topostats/tracing/disordered_tracing.py
```diff
@@ -254,6 +255,31 @@ def smooth_mask(
...
+        Returns
+        -------
+        float
+            Width of grain in metres.
+        """
+        # Code will go here
+        dist_trans = distance_transform_edt(smoothed_mask)
+        comb = np.where(pruned_skeleton == 1, dist_trans, 0)
+
+        return comb[comb != 0].mean() * 2 * px2nm
```
ns-rse, line 281: In my confusion I've now looked up what [`distance_transform_edt()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.distance_transform_edt.html) does and am curious why the mean value is multiplied by `* 2` here?

A dummy example which hopefully explains my concern.

```
import numpy as np
from skimage.morphology import skeletonize
from scipy import ndimage

mask = np.asarray([
    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
    [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
    [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
    [0, 1, 1, 0, 0, 0, 1, 1, 1, 0],
    [0, 1, 1, 0, 0, 0, 1, 1, 1, 0],
    [0, 1, 1, 0, 0, 0, 1, 1, 1, 0],
    [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
    [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
    [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
    [0, 1, 1, 1, 1, 1, 1, 1, 1, 0],
    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],])
skeleton = np.asarray(skeletonize(mask), dtype=int)
distance = ndimage.distance_transform_edt(mask)
comb = np.where(skeleton ==1, distance, 0)
```

The resulting arrays are...

## `skeleton`

```
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 1, 1, 1, 1, 1, 0, 0, 0],
       [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
       [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
       [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
       [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
       [0, 1, 0, 0, 0, 0, 0, 1, 0, 0],
       [0, 0, 1, 1, 1, 1, 1, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]])
```

## `distance`

```
array([[0. ,0. , 0.  , 0., 0.,  0., 0.  , 0., 0., 0.],
       [0. ,1. , 1.  , 1., 1.,  1., 1.  , 1., 1., 0.],
       [0. ,1. , 1.41, 1., 1.,  1., 1.41, 2., 1., 0.],
       [0. ,1. , 1.  , 0., 0.,  0., 1.  , 2., 1., 0.],
       [0. ,1. , 1.  , 0., 0.,  0., 1.  , 2., 1., 0.],
       [0. ,1. , 1.  , 0., 0.,  0., 1.  , 2., 1., 0.],
       [0. ,1. , 1.41, 1., 1.,  1., 1.41, 2., 1., 0.],
       [0. ,1. , 2.  , 2., 2.,  2., 2.23, 2., 1., 0.],
       [0. ,1. , 2.  , 2., 2.,  2., 2.  , 2., 1., 0.],
       [0. ,1. , 1.  , 1., 1.,  1., 1.  , 1., 1., 0.],
       [0. ,0. , 0.  , 0., 0.,  0., 0.  , 0., 0., 0.]])
```

## `comb`

```
array([[0., 0., 0., 0., 0., 0., 0.  , 0., 0., 0.],
       [0., 0., 1., 1., 1., 1., 1.  , 0., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0.  , 2., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0.  , 2., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0.  , 2., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0.  , 2., 0., 0.],
       [0., 1., 0., 0., 0., 0., 0.  , 2., 0., 0.],
       [0., 0., 2., 2., 2., 2., 2.23, 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0.  , 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0.  , 0., 0., 0.],
       [0., 0., 0., 0., 0., 0., 0.  , 0., 0., 0.]])

```


Looking at the right hand-side of this square the `comb` has a width of `2` where the mask width is `3` pixels. By multiplying by `2` it implies the width is `4` when actually it is `3`. This in turn inflates the mean width.

tests/tracing/test_disordered_tracing.py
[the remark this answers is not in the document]
ns-rse: Ok, useful to have edge case examples. Thought about the main function again and have a concern the width may not be calculated correctly, comment in place.

topostats/tracing/disordered_tracing.py
tcatley: I think this would apply to all odd pixel width objects, that we would overcompensate by one 1 pixel? This should only be a big problem for very low resolution scans with the error essentially just being the 1*px_nm_scaling. Having discussed with the group we have decided this is acceptable for now. When @llwiggins and @SylviaWhittle write the grainstats documentation I will let them know so that they can add a note to say that cation should be taken with this parameter for low resolution scans. There are plans for a more sophisticated function to calculate in the future, but this is lower priority.

ns-rse: Ok, I'm more than happy to have incremental improvements and caveats with existing methods noted, nothing is ever perfect and we move forward in small steps, not leaps and bounds.

Prior to merging I will...

- add a modified version of the above test with the true width being `3` (or another odd number) and mark it with `@pytest.mark.xfail()` which means that we will expect it to fail on masks with an odd-numbered of pixel widths.
- write an issue to address this in the future so it doesn't get forgotten about.

ns-rse (approved):

ns-rse (approved):

Issue #80, opened 2022-03-31
Create example Jupyter Notebook (myBinder, Colab)


bobturneruk:
To run on myBinder and on Google Colab. Running on these platforms will (almost certainly) require python 3 and no Gwyddion dependency.

PR #1031, opened 2024-12-02
refactor(io): Load files with AFMReader

ns-rse:
Refactors `topostats/io.py` to use [`AFMReader`](https://github.com/AFM-SPM/AFMReader/) to load all files.

Almost everything works as expected with the exception of `.ibw` where AFMReader raises an `UnboundLocalError` if a file is not as `image` is created as `None` rather than raising `FileNotFound` error. This is captured in [Consistently raise FileNotFound errors across each load_*() functions. #80](https://github.com/AFM-SPM/AFMReader/issues/80) and for now the on test that is affected has been marked as `pytest.mark.xfail()` as it is expected to fail.

I was perhaps a little hasty in making a release of [`AFMReader v0.0.2`](https://github.com/AFM-SPM/AFMReader/releases/tag/v0.0.2) which I realised early on as being required so that tests pass in Continuous Integration.

---

Items that are Not Applicable have been ~~struck through~~.

- [x] Existing tests pass.
- [x] ~~Documentation has been updated and builds. Remember to update as required...~~
  - [x] ~~`docs/configuration.md`~~
  - [x] ~~`docs/usage.md`~~
  - [x] ~~`docs/data_dictionary.md`~~
  - [x] ~~`docs/advanced.md` and new pages it should link to.~~
- [x] Pre-commit checks pass.
- [x] ~~New functions/methods have typehints and docstrings.~~
- [x] ~~New functions/methods have tests which check the intended behaviour is correct.~~


ns-rse:
I might have to think about/revise the use of `.pop()` in `run_modules.process()` as it might be more prudent to remove these conveniences that AFMReader returns when using `LoadScans.get_data()` so that we only have to do it once and don't have to do it for every single module in `run_modules` as the "swiss army knife" work will require loading `.topostats` files for subsequent processing/re-processing.

SylviaWhittle (approved):
Love a PR that reduces the codebase!

This is neat and tidy, agree that using `.pop` would be nice to replace with a more elegant solution in the future, perhaps by automatically removing them from the returned dictionary in `AFMReader` and passing the image, pixel scaling and file version as separate returns (image, scaling, version, data) so we don't have to pop them in each stage like you mentioned we will have to.

ns-rse:
Thanks for the review and approval.

> agree that using .pop would be nice to replace with a more elegant solution in the future, perhaps by automatically removing them from the returned dictionary in AFMReader and passing the image, pixel scaling and file version as separate returns (image, scaling, version, data) so we don't have to pop them in each stage like you mentioned we will have to.

I'm going to change this now, otherwise it will linger and get forgotten about.

ns-rse:
Also just noticed I van remove `_ibw_pixel_to_nm_scaling()` definition too as that has migrated to `AFMReader` and isn't used here.

ns-rse:
Change of heart, going to address this as part of the Swiss army knife work as I don't wish to hold up the pending version release, but have removed the redundant `_ibw_pixel_to_nm_scaling()` function (and kept history clean by using `git commit --fixup` to add it to the relevant commit then `git push --force` :innocent: )

PR #1049, opened 2024-12-11
chore: Remove unused method, test and package dependencies

### LINKED ISSUE #1036: Remove `io.LoadScan._spm_px_to_nm_scaling()`
See [AFMReader #83](https://github.com/AFM-SPM/AFMReader/issues/83) where this method needs to be part of that package so that scans are correctly loaded.

Ideally we should have tests that cover all of the pixel scaling units are correctly loaded in the TopoStats pipeline (although I would be happy to have them covered by unit tests in AFMReader which is probably where they belong).

ns-rse:
Closes #1036

With the delegation of much I/O (Input/Output) functionality to @SylviaWhittle `AFMReader` (see #1031) we have removed most of the methods to the `io.LoadScan()` class. One that slipped through was `_spm_pixel_to_nm_scaling()` and this commit removes it along with its associated test.

Further the delegation of reading AFM files using `AFMReader` means `TopoStats` no longer has explicit dependencies on the `igor2`, `pySPM` or `tifffile` packages and so these have been removed from the `dependencies` section of `pyproject.toml`. These are dependencies of `AFMReader` and will still be installed in virtual environments.

Moved `pytest-testmon` from `tests` to `dev` under optional dependencies as its not used when running the test suite, only when making commits locally.

---

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,13 +38,11 @@ dependencies = [
   "art",
   "AFMReader",
   "h5py",
-  "igor2",
   "keras",
   "matplotlib",
   "numpy",
   "numpyencoder",
   "pandas",
-  "pySPM",
   "pyyaml",
   "ruamel.yaml",
   "schema",
@@ -54,7 +52,6 @@ dependencies = [
   "skan~=0.11.0",
   "snoop",
   "tensorflow",
-  "tifffile",
   "topoly",
   "tqdm",
 ]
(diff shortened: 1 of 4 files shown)


MaxGamill-Sheffield (approved):
Looks good, clean and can confirm it runs from command line on some samples I've been testing! Thanks for the tidy up here Neil

PR #1052, opened 2024-12-16
refactor(disordered_tracing): Work with skan-0.12.2

### LINKED ISSUE #986: Use new skan.summarize(separator="_")
We added a line in [disrordered_tracing()](https://github.com/AFM-SPM/TopoStats/blob/5d01d437144455d179254e7d8eaa98edcc0ff58d/topostats/tracing/disordered_tracing.py#L425) to substitute the `-` in column names of dataframes produced by Skan.

A new version of [skan](https://github.com/jni/skan/releases/tag/v0.12.0) has been released (19hrs ago) which will include the minor contribution I made of converting all variable names from `-` to `_` (see [release notes and linked pull request](https://github.com/jni/skan/releases/tag/v0.12.0)).

We should, until `_` becomes the default, set `skan.summarize(..., separator="_")`.

ns-rse:
Closes #986

With the release of `skan-0.12.2` we can remove the code which converts `-` in the variable names of Pandas DataFrames to instead use `_` which allows the use of "dot-notation" when referring to the variables of data frames.

We therefore unpin the `skan` version in `pyproject.toml`, explicitly state the `separator="_"` argument to `skan.summarize()` and adapt the hard-coded variable names appropriately. As noted in the [release notes](https://github.com/jni/skan/releases/tag/v0.12.2) the default will be `_` from `v0.13.0` and the flag allows backwards compatibility but since we have gone with the use of `_` in variable names it will do no harm leaving these in place (although they could be removed when `skan-v0.13.0` is released).

In addressing #986 I tested in a fresh virtual environment. This pulled in [matplotlib-3.10.0](https://github.com/matplotlib/matplotlib/releases/tag/v3.10.0) which was released 2024-12-14 (two days ago!). Our dependency on [topoly](topoly.cent.uw.edu.pl/documentation.html) doesn't work with this new version because of the removal of `inset_location.InsetPosition` in favour of `inset_axes`.

This has been reported to the Topoly developers, see [ImportError: cannot import name 'InsetPosition' from
'mpl_toolkits.axes_grid1.inset_locator' · Issue #5 · ilbsm/topoly_tutorial](https://github.com/ilbsm/topoly_tutorial/issues/5) and hopefully won't take long to resolve. For now though we pin our dependency `matplotlib~=3.9.4` and all tests pass.

---

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
   "scikit-image",
   "scipy",
   "seaborn",
-  "skan~=0.11.0",
+  "skan",
   "snoop",
   "tensorflow",
   "topoly",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
   "AFMReader",
   "h5py",
   "keras",
-  "matplotlib",
+  "matplotlib~=3.9.4",
   "numpy",
   "numpyencoder",
   "pandas",
--- a/topostats/tracing/disordered_tracing.py
+++ b/topostats/tracing/disordered_tracing.py
@@ -387,7 +387,7 @@ def trace_image_disordered(  # pylint: disable=too-many-arguments,too-many-local
                         np.where(disordered_trace_images["pruned_skeleton"] == 1, cropped_image, 0),
                         spacing=pixel_to_nm_scaling,
                     )
-                    skan_df = skan.summarize(skan_skeleton)
+                    skan_df = skan.summarize(skan_skeleton, separator="_")
                     skan_df = compile_skan_stats(skan_df, skan_skeleton, cropped_image, filename, cropped_image_index)
                     total_branch_length = skan_df["branch_distance"].sum() * 1e-9
                 except ValueError:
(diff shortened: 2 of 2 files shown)


MaxGamill-Sheffield:
Hey Neil, thanks for following up on this issue with the skan and topoly people!

I've not had time to test but does this then change any of the disordered_stats column names? 
This might be an issue for the paper as we have a bunch of analysis scripts relying on certain names and it might be more of an overhead to change this in the `v2.3.0` release along with all the analysis scripts than to cite the version once released.

ns-rse:
> I've not had time to test but does this then change any of the disordered_stats column names?

## Too Long; Did not Read

No, the PR doesn't include any changes to test files where changes in column names would have an impact, _ergo_ nothing in the scripts needs changing because none of the column names have actually changed.

## Long; Did Read

Skan originally used `-` in column names to the dataframe it produces.

This means that you can't use the notation `df.<colname>` when referring to columns that are produced.

I made a PR to [skan](https://github.com/jni/skan/pull/215) to switch to using `_`.

When `disordered_tracing()` was being refactored and the results of `skan.summarize()` were used I mentioned that there was this pending change coming and that we should use `_` rather than `-` in column names.

This PR removes the line that made that change to the returned data frame (see [line itself and associated comment](https://github.com/AFM-SPM/TopoStats/blob/1b45bea9e9e35709ec3ca8bdc24616990f8cfee8/topostats/tracing/disordered_tracing.py#L473)) and also explicitly calls `skan.summarize()` with the `separator="_"` so that the returned data frame uses `_` in the column names and we don't need to manually make the switch. A few other changes were required where comparisons to keys or arguments are hard coded.

Thus the output doesn't actually change at all, which is why all the tests pass without any need to modify _any_ of the test files.

It also threw up the mentioned problem with Matplotlib 3.10.0 causing problems with Topoly which was pure chance but I've addressed it as otherwise the tests in Continuous Integration would pull in the new version and throw errors.

MaxGamill-Sheffield (approved):
Ace! In that case this looks clean and good to go, thanks Neil!

PR #1056, opened 2024-12-17
curvature images doesn't overwrite

MaxGamill-Sheffield:
Closes #1054

Updates curvature main image to have a filename so it doesn't overwrite itself. 
And the individual images to go into a `dnatracing/above/clustering` directory.

PR #1058, opened 2024-12-17
feature: Adds --image-set as option to topostats

### LINKED ISSUE #1057: Add `image_set` to `topostats` argparse
In reviewing #1056 I noticed that we do not have a command line option to modify the value of the `image_set` option.

This should be added as part of the work undertaken by #517 but is small and should probably be squeezed in as I
imagine it would be quite convenient to be able to toggle that option without creating a configuration file and
modifying the value.

ns-rse:
Closes #1057

In reviewing #1056 I wanted to look at the individual grain curvature plots that were generated and realised that there
was no command-line option to modify the configuration value loaded from the YAML (`topostats/default_config.yaml` or
user specified with `--config-file/-c`).

This means that currently if people want the `all` image set then they have to use `topostats create-config` to create a
configuration file, modify the value of `image_set` and run the analysis with `topostats -c new_config.yaml`.

It was very quick and easy to add this option so I could check the output and as its been done I felt it should be
incorporated into TopoStats.

-------

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

### DIFF
--- a/topostats/entry_point.py
+++ b/topostats/entry_point.py
@@ -106,6 +106,13 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Channel to extract.",
     )
+    parser.add_argument(
+        "--image-set",
+        dest="image_set",
+        type=str,
+        required=False,
+        help="Image set to generate, default is 'core' other option is 'all'.",
+    )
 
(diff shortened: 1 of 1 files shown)


MaxGamill-Sheffield (approved):
Looks good, can go into the curvature branch :)

PR #1091, opened 2025-02-19
fix: Updates io to work with AFMReader only returning data

### LINKED ISSUE #1086: Update loading of `.topostats` now just data is returned
In light of changing what [AFMReader returns](https://github.com/AFM-SPM/AFMReader/pull/112** so that _just** the data is
returned we will need to update TopoStats to handle this.

**IMPORTANT** This will require co-ordination with AFMReader releases with clear advice to update to newer versions of
AFMReader and `pip install --no-cache-dir --upgrade topostats`.

Points that need changing....

### `LoadScans.load_topostats()`

[Line 679](https://github.com/AFM-SPM/TopoStats/blob/be8d296d12a5e888227d59d41995eec5292067b8/topostats/io.py#L679)

``` python
            image, px_to_nm_scaling, data = topostats.load_topostats(self.img_path)
```

...will be...

``` python
            data = topostats.load_topostats(self.img_path)
```

We then have a section that was introduced to handle whether we wanted different components starting on [Line
683](https://github.com/AFM-SPM/TopoStats/blob/be8d296d12a5e888227d59d41995eec5292067b8/topostats/io.py#L683)

``` python
        try:
            # We want everything if performing any step beyond filtering (or explicitly ask for None/"all")
            if extract in [None, "all", "grains", "grainstats"]:
                return (image, px_to_nm_scaling, data)
            # Otherwise we want the raw/image_original
            if extract == "filter":
                return (image, px_to_nm_scaling, None)
            return (data[map_stage_to_image[extract]], px_to_nm_scaling, None)
        except KeyError as ke:
            raise KeyError(f"Can not extract array of type '{extract}' from .topostats objects.") from ke
```

...as we will only ever be returning `data` this can be replaced with

``` python
        return data
```

### `LoadScans.get_data()`

There are a logical checks here that may need changing to account for only receiving `data`. This section extracts
`self.image` `selt.pixel_to_nm_scaling` and optionally `data` depending on the type of processing being undertaken. This
is stored in the `self.extract` and represents `all` or the stage (e.g. `filter`, `grain`, `grainstats`).

A key point which I missed when working on this was that the `else:` after the `try: ... except: ...` as I forgot that
the `else:` will run if no exception is raised (i.e. `try:` is successful), and this step builds a dictionary of the
images/data.

[Line 782](https://github.com/AFM-SPM/TopoStats/blob/be8d296d12a5e888227d59d41995eec5292067b8/topostats/io.py#L782)

``` python
                try:
                    if suffix == ".topostats" and self.extract in (None, "all", "grains", "grainstats"):
                        self.image, self.pixel_to_nm_scaling, data = self.load_topostats()
                    elif suffix == ".topostats" and self.extract not in (None, "all"):
                        self.image, self.pixel_to_nm_scaling, _ = self.load_topostats(self.extract)
                    else:
                        self.image, self.pixel_to_nm_scaling = suffix_to_loader[suffix]()
                except Exception as e:
                    if "Channel" in str(e) and "not found" in str(e):
                        LOGGER.warning(e)  # log the specific error message
                        LOGGER.warning(f"[{self.filename}] Channel {self.channel} not found, skipping image.")
                    else:
                        raise
                else:
                    if suffix == ".asd":
                        for index, frame in enumerate(self.image):
                            self._check_image_size_and_add_to_dict(image=frame, filename=f"{self.filename}_{index}")
                    # If we have extracted the image dictionary (only possible with .topostats files) we add that to the
                    # dictionary
                    elif data is not None:
                        data["img_path"] = img_path.with_suffix("")
                        self.img_dict[self.filename] = self.clean_dict(img_dict=data)
                    # Otherwise check the size and add image to dictionary
                    else:
                        self._check_image_size_and_add_to_dict(image=self.image, filename=self.filename)
```

Could probably be

``` python
                try:
                    if suffix == ".topostats":
                        data = self.load_topostats()
                        self.image = data["image"]
                        self.pixel_to_nm_scaling = data["pixel_to_nm_scaling"]
                    else:
                        self.image, self.pixel_to_nm_scaling = suffix_to_loader[suffix]()
                except Exception as e:
                    if "Channel" in str(e) and "not found" in str(e):
                        LOGGER.warning(e)  # log the specific error message
                        LOGGER.warning(f"[{self.filename}] Channel {self.channel} not found, skipping image.")
                    else:
                        raise
                else:
                    if suffix == ".asd":
                        for index, frame in enumerate(self.image):
                            self._check_image_size_and_add_to_dict(image=frame, filename=f"{self.filename}_{index}")
                    # If we have extracted the image dictionary (only possible with .topostats files) we add that to the
                    # dictionary
                    elif data is not None:
                        data["img_path"] = img_path.with_suffix("")
                        self.img_dict[self.filename] = self.clean_dict(img_dict=data)
                    # Otherwise check the size and add image to dictionary
                    else:
                        self._check_image_size_and_add_to_dict(image=self.image, filename=self.filename)

```

## `run_modules.py`

The entry points/functions for running `topostats {filters|grains}` shouldn't need changing I don't think but worth
noting this is where the returned data is used. Something in the back of my mind says they will need changing but I
can't see it right now, perhaps it will be the `process_filters()` as this will have to extract the `original_image` from
the returned `data` as it will no longer be receiving a dictionary with this as the first item. Probably similar for the
`process()` function too, but I think the `process_grains()` function will be ok as its expecting to get a dictionary.

ns-rse:
Closes #1086

Updates the `LoadScans.load_topostats()` and `LoadScans.get_data()` to work with the new version of AFMReader.

~~This means the CI tests _will_ fail because we haven't released AFMReader yet (and I'm wary of doing so until we are ready to make another TopoStats release).~~

EDIT : Updated `pyproject.toml` to install AFMReader from GitHub so tests should pass on CI :crossed_fingers: 

But...tests all pass for me locally(!) with AFMReader installed from `HEAD` of `main` branch.

If you want to check locally you'll need a fresh Python3.11 environment (because `topoly`!) and to install AFMReader from your local git clone (you can install from git using `pip` but I can't remember the command and am not going to look it up now!).

-------

# TopoStats Pull Requests

- [x] Existing tests pass.
- [ ] ~~Documentation has been updated and builds. Remember to update as required...~~
  - [ ] ~~`docs/configuration.md`~~
  - [ ] ~~`docs/usage.md`~~
  - [ ] ~~`docs/data_dictionary.md`~~
  - [ ] ~~`docs/advanced.md` and new pages it should link to.~~
- [x] ~~Pre-commit checks pass.
- [ ] ~~New functions/methods have typehints and docstrings.~~
- [ ] ~~New functions/methods have tests which check the intended behaviour is correct.~~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ keywords = [
 requires-python = ">=3.9, <3.12"
 dependencies = [
   "art",
-  "AFMReader",
+  "AFMReader @ git+https://github.com/AFM-SPM/AFMReader@main",
   "h5py",
   "keras",
   "matplotlib~=3.9.4",
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -33,7 +33,6 @@ def test_median_flatten_no_mask(
     median_flattened = test_filters_random.median_flatten(
         image, mask=None, row_alignment_quantile=row_alignment_quantile
     )
-
     assert isinstance(median_flattened, np.ndarray)
     assert median_flattened.shape == (20, 20)
     np.testing.assert_allclose(median_flattened, expected, **TOLERANCE)
@@ -42,7 +41,6 @@ def test_median_flatten_no_mask(
 def test_remove_tilt_no_mask(test_filters_random: Filters, image_random_remove_x_y_tilt: np.array) -> None:
     """Test removal of x/y tilt."""
     tilt_removed = test_filters_random.remove_tilt(test_filters_random.images["pixels"], mask=None)
-
     assert isinstance(tilt_removed, np.ndarray)
     assert tilt_removed.shape == (1024, 1024)
     np.testing.assert_allclose(tilt_removed, image_random_remove_x_y_tilt, **TOLERANCE)
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -587,14 +587,14 @@ def test_load_scan_topostats_all(load_scan_topostats: LoadScans) -> None:
     """Test loading all data from a .topostats file."""
     load_scan_topostats.img_path = load_scan_topostats.img_paths[0]
     load_scan_topostats.filename = load_scan_topostats.img_paths[0].stem
-    image, px_to_nm_scaling, data = load_scan_topostats.load_topostats(extract="all")
+    data = load_scan_topostats.load_topostats(extract="all")
     above_grain_mask = data["grain_masks"]["above"]
     grain_trace_data = data["grain_trace_data"]
-    assert isinstance(image, np.ndarray)
-    assert image.shape == (1024, 1024)
-    assert image.sum() == 184140.8593819073
-    assert isinstance(px_to_nm_scaling, float)
-    assert px_to_nm_scaling == 0.4940029296875
+    assert isinstance(data["image"], np.ndarray)
+    assert data["image"].shape == (1024, 1024)
+    assert data["image"].sum() == 184140.8593819073
+    assert isinstance(data["pixel_to_nm_scaling"], float)
+    assert data["pixel_to_nm_scaling"] == 0.4940029296875
     # Check that the grain mask is loaded correctly
     assert isinstance(above_grain_mask, np.ndarray)
     assert above_grain_mask.sum() == 633746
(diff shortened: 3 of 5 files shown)


ns-rse:
Realised we can install AFMReader directly from GitHub which avoids the need to have a release. However, once a release is made we should going forward, change `pyproject.toml` to have a dependency of `  "AFMReader>=0.#.#"` so that older versions of AFMReader aren't pulled in.

SylviaWhittle (approved):
Works on my machine too and looks good 👍 thank you

ns-rse:
Thanks @SylviaWhittle wasn't quite the solution/work I'd sketched out in #1086 but was surprisingly less work than I expected.

Should make the remaining tools on the Swiss Army knife a bit quicker for me to burn through. :crossed_fingers:

PR #1092, opened 2025-02-20
style: using dictionaries for parameters and **kwargs

ns-rse:
- Spotted a few `print()` statements from debugging.
- Explicitly test the number of grains below that are returned.
- switching to a dictionary in the parameterisaion of `test_merge_classes()` instead of multiple individual
  options with comments/labels. The dictionaries are expressive about what the values are since the keys are the
  configuration options themselves. This in turn means we can just use `**vet_grains_conf` to unpack the dictionary of
  options when calling the `vet_grains()` function.

### DIFF
--- a/tests/test_grains.py
+++ b/tests/test_grains.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from typing import Any
 from unittest.mock import MagicMock, patch
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


SylviaWhittle (approved):
I haven't tested this locally but it's clear this is a definite improvement. So much neater! thank you very much 🙏

PR #1094, opened 2025-02-24
feature(grainstats): Adds entry point for grainstats

ns-rse:
Closes #743

Adds an entry point for running GrainStats `process grainstats --help`.

You can test it with...

```
❱ topostats -b tests/resources/test_image -f .topostats grainstats
❱ tree output
[4.0K Feb 24 14:53]  output
├── [2.7K Feb 24 14:53]  output/config.yaml
├── [2.7K Feb 24 14:53]  output/height_profiles.json
├── [1.7K Feb 24 14:53]  output/image_stats.csv
└── [4.0K Feb 24 14:53]  output/processed
    └── [109K Feb 24 14:53]  output/processed/minicircle_small.topostats

2 directories, 4 files
```

Unlike with #1076 only the required directory structure is created.

-------

Before submitting a Pull Request please check the following.

- [x] Existing tests pass.
- [x] Documentation has been updated and builds. Remember to update as required...
- [x] Pre-commit checks pass.
- [x] New functions/methods have typehints and docstrings.
- [x] New functions/methods have tests which check the intended behaviour is correct.

PR #1099, opened 2025-03-04
tests: disable test_run_modules::test_grainstats

ns-rse:
The refactoring to use classes for objects rather than dictionaries breaks the `topostats grainstats` entry point that
was introduced with #1094.

Previously the dictionary item `image=topostats_object["image"]` was passed into `processing.run_grainstats()` when called from
`processing.process_grainstats()`. The refactoring requires that `image_grain_crop=image_grain_crop`, an object of the new type
`ImageGrainCrops`, is now passed to `grainstats`.

This isn't currently possible though because `AFMReader` loads `.topostats` objects and returns a dictionary and whilst
the refactoring does save the new `image_grain_crop` (/`ImageGrainCrops`) the loading does _not_ currently re-create
these structures.

For now I have disabled the test of the entry point. Once this refactoring has been merged we will have to...

- Make `TopoStats` a dependency of `AFMReader` (somewhat wary of this as it may cause headaches further down the line
  but for now we'll go with it!).
- Modify `AFMReader.topostats.load_topostats()` to modify the `data["grain_tensors"]["above"]` and
  `data["grain_tensors"]["above"]` so that they are of class `ImageGrainCrops` (and the associated nested classes).
- Once that is done we can then pass the loaded `data["grain_tensors"]` to `processing.run_grainstats()` (this may
require reconstructing to be the same as `image_grain_crop`, not sure at the moment!)


ns-rse:
Ignore this, I got the wrong target branch (should have been `SylviaWhittle/grain_restructure` and not `SylivaWhitle/grain-refactor-tracing`).

PR #1104, opened 2025-03-12
feature(grains|io): Method to convert GrainCrop to dictionary

ns-rse:
Adds a method to convert a `grains.GrainCrop` object to a dictionary (`grains.GrainCrop.grain_crop_to_dict()`) which is
indexed by the attribute names with the attributes as the values. Includes a simple test using the `dummy_graincrop`
fixture.

As this will be needed to convert the nested dictionaries of `ImageGrainCrops` (which hold `GrainCrop` objects within
`GrainCropsDirection`) objects we include a check to see if the `item` (value) of a dictionary is of type `GrainCrop`
and if so recursively call `dict_to_hdf5` on the result of `item.grain_crop_to_dict()`. Again a simple test to ensure
this functions as expected is included.

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

### DIFF
--- a/tests/test_grains.py
+++ b/tests/test_grains.py
@@ -68,6 +68,19 @@ grain_array4 = np.array(
 )
 
 
+def test_grain_crop_to_dict(dummy_graincrop: GrainCrop):
+    """Test the GrainCrop.grain_crop_to_dict() method."""
+    expected = {
+        "image": dummy_graincrop.image,
+        "mask": dummy_graincrop.mask,
+        "padding": dummy_graincrop.padding,
+        "bbox": dummy_graincrop.bbox,
+        "pixel_to_nm_scaling": dummy_graincrop.pixel_to_nm_scaling,
+        "filename": dummy_graincrop.filename,
+    }
+    np.testing.assert_array_equal(dummy_graincrop.grain_crop_to_dict(), expected)
+
+
 @pytest.mark.parametrize(
     ("area_thresh_nm", "expected"),
     [([None, None], grain_array), ([None, 32], grain_array2), ([12, 24], grain_array3), ([32, 44], grain_array4)],
@@ -279,7 +292,6 @@ def test_remove_edge_intersecting_grains(
     assert number_of_grains == expected_number_of_grains
 
 
-# Find grains without unet
 @pytest.mark.parametrize(
     (
         "image",
@@ -621,7 +633,7 @@ def test_find_grains(
     expected_labelled_regions: npt.NDArray[np.int32],
     expected_imagegraincrops: ImageGrainCrops,
 ) -> None:
-    """Test the find_grains method of the Grains class."""
+    """Test the find_grains method of the Grains class without unet."""
     # Initialise the grains object
     grains_object = Grains(
         image=image,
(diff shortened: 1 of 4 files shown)


tests/test_io.py
```diff
@@ -959,6 +960,35 @@ def test_dict_to_hdf5_list(tmp_path: Path) -> None:
         np.testing.assert_array_equal(f["list"][()], expected["list"])
 
 
+def test_dict_to_hdf5_graincrop(dummy_graincrops_dict: grains.GrainCrop, tmp_path: Path) -> None:
+    """Test loading a GrainGrop object and writing to file."""
+    # Make a dictionary from dummy_graincrop
+    print(f"{dummy_graincrops_dict=}")
```
ns-rse, line 966:
```suggestion
```

Clocked I'd left this lingering in the stacked branch I'm preparing, removing it from this PR and it won't reappear.

ns-rse, line 966:
```suggestion
```

Clocked I'd left this lingering in the stacked branch I'm preparing, removing it from this PR and it won't reappear.

topostats/io.py
```diff
@@ -929,23 +931,54 @@ def dict_to_hdf5(open_hdf5_file: h5py.File, group_path: str, dictionary: dict) -
...
+                LOGGER.debug(f"[dict_to_hdf5] {key} is of type : {type(item)}")
                 open_hdf5_file[group_path + key] = str(item).encode("utf8")
+            # Extract ImageGrainCrops
+            elif isinstance(item, grains.ImageGrainCrops):
+                LOGGER.debug(f"[dict_to_hdf5] {key} is of type : {type(item)}")
+                dict_to_hdf5(open_hdf5_file, group_path + key + "/", item.above)
+                dict_to_hdf5(open_hdf5_file, group_path + key + "/", item.below)
+            elif isinstance(item, grains.GrainCropsDirection):
+                LOGGER.debug(f"[dict_to_hdf5] {key} is of type : {type(item)}")
+                dict_to_hdf5(open_hdf5_file, group_path + key + "/", item.crops)
```
ns-rse, line 974: **NB** - These are being worked on already!

tests/test_grains.py
```diff
@@ -68,6 +68,19 @@
 )
 
 
+def test_grain_crop_to_dict(dummy_graincrop: GrainCrop):
```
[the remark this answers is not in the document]
ns-rse: It was only because I clocked the use of `np.testing.assert_array_equal()` that you'd used above. :slightly_smiling_face:

ns-rse:
Thanks for the review @SylviaWhittle 

It was so easy because of the ground work you've laid with a nice `dict_to_hdf5()` function that we can call recursively and the excellent use of classes in your `Grains` refactor. I actually started writing out the `grain_crop_to_dict()` function to do it manually before realising there must be a dunder method to do this automatically and so switched to that.

Means when we come round to adding all the other properties to the `GrainCrop` things should "Just Work(tm)".

Just finishing off the last method of `ImageGrainCrops()` (might not complete it tonight though due to other responsibilities).

PR #1106, opened 2025-03-12
feature(grains|io): ImageGrainCrop to dictionary

ns-rse:
Final stacked PR should be merged after #1105

This adds a method and tests to convert `ImageGrainCrop` to dictionary and write to HDF5. The pattern is very similar to
previous. I'm not sure at the moment how this will slot in with the larger refactoring of passing these around between
the different classes but its there and ready to go if/when we need it.

- [X] Existing tests pass.
- [X] Pre-commit checks pass.
- [X] New functions/methods have typehints and docstrings.
- [X] New functions/methods have tests which check the intended behaviour is correct.


SylviaWhittle (approved):
Thank you so much for this, excellent

PR #1122, opened 2025-04-02
Adds a class name option to the config file and grainstats

MaxGamill-Sheffield:
# TopoStats Pull Requests

Enables the user to provide strings which represent the grain classes in TopoStats. These strings are then mapped to the `class_number` produced from the grainstats section to give physical meaning to the classes seen in the `grainstats.csv` output with a new `class_number` column. In the future, this parameter works when/if the user want's to merge multiple classes by simply adding those strings onto the end i.e: for a 3-class output and merging classes 1 and 2.
```
merge_classes: [[1,2]]
class_names: ['1_name', '2_name', '3_name', '1_name+2_name']
```
Summary:
- Adds a new config file option `class_names` which takes a list of strings.
- This is included in grainstats, but not used within the `GrainStats` module itself, only the processing of grainstats.
- Includes the missing `subgrain_number` and `class_number` as well as the new `class_number` to the empty grainstats dataframe.
- Fixes failing tests due to the extra column.
- Is extendable for the future.
- Added into entry point code.
- Added into config and docs.

---

- [ x] Existing tests pass.
- [ x] Documentation has been updated and builds. Remember to update as required...
  - [ x] `docs/configuration.md`
  - [ x] `docs/data_dictionary.md`
- [ x] Pre-commit checks pass.

- [ x] There is a comment adjacent to the option explaining what it is and the valid values.
- [ x] A check is made in `topostats/validation.py` to ensure entries are valid.
- [ x] Add the option to the relevant sub-parser in `topostats/entry_point.py`.


topostats/validation.py
```diff
@@ -262,6 +262,7 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 error="Invalid value in config for 'grainstats.extract_height_profile',"
                 "valid values are 'True' or 'False'",
             ),
+            "class_names": [str],
```
ns-rse, line 265: Could we be more explicit here and state what we expect these to be with...

```suggestion
            "class_names": Or(
                         "DNA",
                         "Protein",
                         error="Invalid value in config for 'class_names'"
                 ),
```

...although that probably isn't the way to validate a list but I've more to suggest on this which I'll write in the broader comment.

ns-rse (changes requested):
Very quick implementation of this and it looks ok but I wonder if this is making additional work and possibly creating scope for problems in the future.


At the moment classes are handled internally as `0`, `1` and possibly `2` (I haven't checked!) and they are used when building some aspects of the `GrainCrop` dictionaries where these values (numbers) are used as keys in dictionaries. But dictionaries can also take strings as keys and so we can be more explicit and avoid the need to map `class_names` with `enumerate()` by just using them directly. Even if they are used in `if class_index == 0:` or other places then we can also use `if class_name == "DNA":` (or better still `class_type`, should avoid use of `class` because of its similarity to `Class`!).

This would also avoid a problem with the use of `enumerate()` on the configuration options because at the moment it is conditional on the order of `class_names` remaining fixed in the `default_config.yaml` or any other configuration file that someone might create, use and possibly extend.

If someone created their own and for whatever reason changed the config file from...

```patch
  class_names: ["DNA", "Protein"]
```
...to...

```
  class_names: ["Protein", "DNA"]
```

...then the labeling in the `class_names` column of the output files would be flipped (I think if I've followed this all through correctly?).

Having had a very quick scan through there are a couple of places where `for class_index in range(1, mask.shape[2]):` is used which might cause some problems working out how many masks there are to loop over as this is probably where there is a separate mask for `DNA` and `Protein`.

A completely different approach might be to not make this an option in the ever expanding configuration file and instead define internally [enum](https://docs.python.org/3/library/enum.html) that map the name to the number. These aren't something I've really used but I think they may fit the use case here.

tests/test_utils.py
```diff
@@ -188,7 +188,7 @@ def test_get_thresholds_value_error(image_random: np.ndarray) -> None:
 @pytest.mark.parametrize(
     ("column_set", "n_columns", "columns_check"),
     [
-        pytest.param("grainstats", 27, {"image", "grain_number", "basename", "area"}, id="Empty grainstats dataframe"),
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Yep that's it!

topostats/entry_point.py
```diff
@@ -373,6 +373,13 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Extract height profiles along maximum feret of molecules",
     )
+    process_parser.add_argument(
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Ah good spot! I briefly forgot that there was the additional grainstats parser too!

I've just added that but found the parser doesn't work because `io.py` looks for "grain_masks" not "grain_tensors" but I'll make an issue and sort this today!

MaxGamill-Sheffield:
> Very quick implementation of this and it looks ok but I wonder if this is making additional work and possibly creating scope for problems in the future.

I agree that this could really mess with the internals of TopoStats, and so I opted for the simple solution of just dumping an extra `class_name` column into the grainstats which should only affect the output grainstats.

> I know that personally I would find using strings as dictionary keys much more intuitive than having to mentally interpolate these values when reading code.

I very much agree in that the dictionaries saved to the `.topostats` files would be more intuitive as strings. However, I've just had a discussion with Sylvia who believes that until we have the multi-class stuff into TopoStats and we know how we want these objects to look, and where is best to edit them, this superficial layer should be good enough for now to help the experimentalists process their data.

Overhauling the entire pipeline to ensure that each processing stage uses this part of the config would be required and we're not even sure how the tracing pipeline works with class / subgrain objects just yet. But I do agree and think this might be a road to cross in the future.

> If someone created their own and for whatever reason changed the config file from...
> 
>   class_names: ["DNA", "Protein"]
> ...to...
>   class_names: ["Protein", "DNA"]
> ...then the labeling in the class_names column of the output files would be flipped (I think if I've followed this all through correctly?).

This is wanted behaviour as the segmentation models that are used may have been trained differently / multi-class traditional thresholding may have a "reversed" order. These mappings should match those in the order specified in the config file (when the multi-class traditional thresholding feature is developed) but for now should only relate to the classes of any U-Net model.

> A completely different approach might be to not make this an option in the ever expanding configuration file and instead define internally [enum](https://docs.python.org/3/library/enum.html) that map the name to the number and use that at the very last step in the creation/modification of the dataframe to add the class_name column.

I believe this requires hardcoding the class names which isn't how this addition intends to work, but that's a cool function I didn't realise you could make your own enumeration, so thank you!

topostats/entry_point.py
```diff
@@ -373,6 +373,13 @@ def create_parser() -> arg.ArgumentParser:
...
+        "--grainstats-class-names",
+        dest="grainstats_class_names",
+        type=list,
```
ns-rse, line 379: I forgot to look this up the other day as I've never done it but `add_argument()` needs more than just the `type=list` option it needs the `nargs=` option, probably with the value `"+"` to indicate that there has to be one or more argument (i.e. `nargs="+"`) and these are then stored in a `list`.

ns-rse:
Thanks @MaxGamill-Sheffield clearly some lack of understanding on my behalf.

I like to try and avoid quick wins if we know that there will be a better way to do something further down the line.

To which end I've given this some thought...

By having the mapping defined in the configuration file it introduces a disconnect and relies on users getting things right. We've tried to make it so that these sort of user interaction/interventions are minimised or not required at all.

If the mapping of classes to object names is a feature of the particular Unet that is used to do the feature selection would it make sense to tie the mapping to the Unet model that is loaded and used? (That these aren't included in TopoStats is also a challenge and we need to think about how we are going to provision these).

That way when a particular Unet is loaded the class definitions are loaded with it and there is no need for users to either create a configuration file and edit the mapping nor remember to specify the correct mapping on the command line.

## `merge_classes`

I'm curious how the user specifies which classes to merge? There is the method `Grains.merge_classes()` which takes the argument `classes_to_merge` but as far as I can see this is only ever set to `None` as calls to `Grains` made from `processing.run_grains()` only takes `grains_config` for the options and there isn't any definition of
`classes_to_merge` (rather than the mentioned `merge_classes`) being added to `default_config.yaml` that would allow users to combine these so I don't understand how specifying `class_names: ['1_name', '2_name', '3_name', '1_name+2_name']` would actually result in combining classes.

I had a look at #1123 thinking it might be something @SylviaWhittle was adding but couldn't see that there either.

MaxGamill-Sheffield:
> I like to try and avoid quick wins if we know that there will be a better way to do something further down the line.

Usually I'd agree, but with the overhaul Sylvia is doing to enable mutli-class traditional thresholding I'd like to avoid 1. Conflicts, and 2. Modifying the internals before we understand how they work as @SylviaWhittle is still figuring this out and testing various modifications of the pipeline.

> By having the mapping defined in the configuration file it introduces a disconnect and relies on users getting things right. We've tried to make it so that these sort of user interaction/interventions are minimised or not required at all.

Yep! This is why I've tried to avoid integrating this into the internals of TopoStats so it's just a "visual" thing in the stats file to help the experimentalists. Maybe a compromise would be to have the default set as ["undefined", "undefined"].

> If the mapping of classes to object names is a feature of the particular Unet that is used to do the feature selection would it make sense to tie the mapping to the Unet model that is loaded and used? (That these aren't included in TopoStats is also a challenge and we need to think about how we are going to provision these).
> That way when a particular Unet is loaded the class definitions are loaded with it and there is no need for users to either create a configuration file and edit the mapping nor remember to specify the correct mapping on the command line.

So I have recommended that the coding team serialise their mappings in future so they are saved with the models. However, this would still then not be applied to the mutli-class traditional thresholding, hence the config options.

> ## `merge_classes`

So it turns out Sylvia has already added this but forgot to include it into the config and documentation etc so I'm just going to tidy it up. I undertstand that it should take a list of lists i.e: [[1,2],[1,3]] to make new tensor slices merging classes 1 & 2 and 1 & 3. Then in the class names, someone can just add `['1_name', '2_name', '3_name', '1_name+2_name', '1_name+3_name']` avoiding overhead of making a function to do this internally while things are still moving about.

> I had no plans to modify the tracing pipeline at all for writing `.topostats` files and adding all features to `GrainCrop` but you may want to hold back and see how that is panning out in the work I'm doing, otherwise we're going to have divergence.

Ace I'll tackle some of my other jobs in the meantime then 😊

topostats/entry_point.py
ns-rse: Sorry for not being explicit, the same option needs adding to the `grainstats_parser()` (but see comment below at that point).

topostats/entry_point.py
```diff
@@ -854,6 +861,14 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Extract height profiles along maximum feret of molecules",
     )
+    process_parser.add_argument(
```
ns-rse, line 864: This is wrong the argument should be added to the `grainstats_parser` and not `process_parser` (the hazards of copy and pasting!).

```suggestion
    grainstats_parser.add_argument(
```

MaxGamill-Sheffield:
Had confirmation from the gang to merge <3

PR #1124, opened 2025-04-02
tests(io): Fix test_dict_to_hdf5_graincrop()

ns-rse:
Thanks to @SylviaWhittle :eagle: :eye: in spotting an error in a fixture I added to `tests/conftest.py` I took a closer look at where it was used and spotted some problems with a previous PR of mine was a little hasty/not as thorough as it should have been and most of the `assert` statements were testing equivalence at all.

This corrects that problem.

- [X] Existing tests pass.
- [X] Pre-commit checks pass.


SylviaWhittle (approved):
Ah I see, it took me a while to notice what was wrong, a very understandable error 😛

PR #1127, opened 2025-04-04
Selective plotting of TopoStats images

### LINKED ISSUE #958: Ability to set image_set for filters, grains and tracing separately to reduce output
Users flagged that the new updates to DNA tracing significantly increase the amount of outputs produced when `image_set: all` is set in the config file. It was agreed that as all output images have potential to be useful for parameter refining, rather than reduce the output it would be better to have the option to set whether `core` or `all` should be output for filters, grains and tracing separately.

MaxGamill-Sheffield:
# TopoStats Pull Requests

Closes #958. This turns the config option `image_set` to use a list of strings. The strings relate to each module and then enable plotting of a few or many specified module images in order to speed-up processing on large datasets where some but not all diagnostic images are required.

This retains the single use all / core functionality but extends this to use combinations of:
- filters
- grain
- grain_crops
- disordered_tracing
- nodestats
- ordered_tracing
- splining

This includes updating the docs, adapting old tests, and creating a new test for this behaviour.

---

- [x] Existing tests pass.
  - [x] `docs/configuration.md`
  - [x] `docs/usage.md`
- [x] Pre-commit checks pass.


topostats/processing.py
```diff
@@ -200,74 +197,71 @@ def run_grains(  # noqa: C901
...
+                            LOGGER.debug(f"[{filename}] : Plotting {plot_name} image")
+                            plotting_config["plot_dict"][plot_name]["output_dir"] = grain_out_path_direction
+                            Images(
+                                data=np.zeros_like(array),
+                                masked_array=array,
+                                **plotting_config["plot_dict"][plot_name],
+                            ).plot_and_save()
+                    # Plot individual grain masks
+                    if direction == "above":
+                        direction_grain_crops: GrainCropsDirection = grains.image_grain_crops.above
```
ns-rse, line 227: Personally I find it easier and more logically consistent when instantiated objects follow the same order as the class they represent so pedantically suggest...

```suggestion
                        grain_crops_direction: GrainCropsDirection = grains.image_grain_crops.above
```


...which obviously needs cascading through (the language server in your IDE should make such a change trivial and is more reliable than a find/replace).

topostats/validation.py
```diff
@@ -563,6 +628,20 @@ def validate_config(config: dict, schema: Schema, config_type: str) -> None:
                 "figure",
                 error="Invalid value in config for 'dpi', valid values are 'figure' or > 0.",
             ),
+            "module": Or(
```
ns-rse, line 631: I'm not sure this is correct but may easily have misunderstood.

Doesn't the `module` field within the `plotting_dictionary.yaml` indicate what module the image belongs to?

If so the `inital_quadratic_removal` is an image from the `Filters()` stage of processing so should this not just be checking that the value is `filters` and nothing else, as we wouldn't want the image to be plotted as part of the `nodestats` set of images.

Same applies across all other additions to `topostats/validation.py`.

ns-rse (changes requested):
Great work sorting all this out @MaxGamill-Sheffield 

A bunch of comments are inline and for a couple of places I've gone ahead and made the changes on a branch and made PRs targeting your feature branch (i.e. `maxgamill-sheffield/958-modularise-image_set`) as it seemed easier/helpful to do so than making the changes in suggestions here on the PR.

One thing I noticed is that the `test_image_set()` takes a while to run as it is repeatedly processing `minicircle_small` which seems a bit wasteful (the test suite already takes a number of minutes to run in CI already and keep the feedback loop as short as possible is preferable). I was wondering if its possible to use an already processed `minicircle_small.topostats` object and just pass that into the plotting functions with the different parameterised options?

tests/test_processing.py
```diff
@@ -213,8 +213,12 @@ def test_process_scan_both(regtest, tmp_path, process_scan_config: dict, load_sc
...
-        ("all", True),
+        (["core"], False),
+        (["all"], True),
+        (["filters"], False),
+        (["grain_crops"], True),
+        (["filters", "grain_crops", "disordered_tracing"], True),
+        (["filters", "grains", "disordered_tracing"], False),
     ],
 )
 def test_save_cropped_grains(
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: merged PR :)

tests/test_processing.py
```diff
@@ -259,6 +263,89 @@ def test_save_cropped_grains(
...
+        (
+            pytest.param(
+                ["filters", "disordered_tracing", "splining"],
+                [True, True, False, False, True, False, False, True],
+                id="filters, disordered_tracing, splining",
+            )
+        ),
+    ],
+)
+def test_image_set(tmp_path: Path, process_scan_config: dict, load_scan_data: LoadScans, image_set, expected) -> None:
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Yep done! Also added typehints to the test above :)

topostats/validation.py
MaxGamill-Sheffield: Makes sense to me. I previously thought it was to check across the possible values, but this makes more sense. I wonder if this might be useful if someone want's to turn on/off plots this way - but this could just be done by switching the core_set field to True.

I've set it to just the module now :)

ns-rse (approved):
Thanks for addressing all the suggestions @MaxGamill-Sheffield 
 and accepting the small PR suggestions I made.
 
 I think this is ready to merge.

PR #1132, opened 2025-04-09
Extends height-based pruning beyond endpoints

MaxGamill-Sheffield:
# TopoStats Pull Requests

Sorry I missed it before but the height-based pruning only worked on skeleton segments with endpoints. This meant that false bridges, caused by holes in the object mask, could not be removed see below:
Before:
Screenshot 2025-04-09 at 15 08 39
Now:
Screenshot 2025-04-09 at 15 08 56

This was super helpful in the Over or Under? paper and so thought we would want to keep this ability.

This PR:
- Wraps the section of the code responsible for filtering out the non-endpoint branches into an `if` statement in `processing.py`.
- Adds a new config flag `disordered_tracing.pruning_params.only_height_prune_endpoints` decides whether to restrict height-based pruning to skeleton segments containing an endpoint or not via the above `if` statement.
- Adds associated validation, configuration.md docs, etc
- Adds an additional test similar to the already present figure 8 test but with a low middle region to be pruned.

---

- [x] Existing tests pass.
  - [x] `docs/configuration.md`
- [x] Pre-commit checks pass.


ns-rse (changes requested):
> Wraps the section of the code responsible for filtering out the non-endpoint branches into an if statement in processing.py.

Probably just an oversight (you're not the only one I forgot to include new `.npy` arrays in my current PR!) but current change set doesn't include any modifications to `processing.py`.

docs/configuration.md
```diff
@@ -137,14 +137,15 @@ Aside from the comments in YAML file itself the fields are described below.
...
+|                                                      | `dilation_iterations`                   | int                                                | `2`                         | The the number of dilations to perform to smooth. This will compete with `gaussian_sigma` to see which changes the grain mask least, ensuring quality over different scan sizes.                                                                                                                                                                                                                                                          |
```
ns-rse, line 140: Good spot, this wouldn't have been caught by the `codespell` hook by virtue of the `_`.

MaxGamill-Sheffield: Thank you!

MaxGamill-Sheffield:
> Probably just an oversight (you're not the only one I forgot to include new `.npy` arrays in my current PR!) but current change set doesn't include any modifications to `processing.py`.

@ns-rse This is my bad! Changes are in lines ~714-723 of `pruning.py` not processing as previously stated in the PR description (will amend this too).

But this should still work as expected and there shouldn't be any changes in `processing.py` due to the config dict being passed through and only unpacked in pruning

tests/tracing/test_pruning.py
```diff
@@ -2037,6 +2043,109 @@ def test_prune_skeleton(
...
+                    [0, 1, 0, 0, 0, 1, 0],
+                    [0, 0, 1, 0, 1, 0, 0],
+                    [0, 1, 0, 0, 0, 1, 0],
+                    [0, 1, 0, 0, 0, 1, 0],
+                    [0, 0, 1, 1, 1, 0, 0],
+                    [0, 0, 0, 0, 0, 0, 0],
+                ]
+            ),
+            id="bridge (low figure 8)",
+            # marks=pytest.mark.xfail(reason="Not sure middles are correct, arbitrarily takes point left or gith of even
```
ns-rse, line 2146: This looks like a hangover from copy and pasting so regardless of whether the above is correct can likely be removed.

tests/tracing/test_pruning.py
```diff
@@ -2037,6 +2043,109 @@ def test_prune_skeleton(
             # marks=pytest.mark.xfail(reason="Not sure middles are correct, arbitrarily takes point left or gith of even
```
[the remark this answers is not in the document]
MaxGamill-Sheffield: Yep, I've removed

tests/tracing/test_pruning.py
MaxGamill-Sheffield: have also removed from here :)

ns-rse (approved):
Cool, looks ready to merge :+1:

PR #1141, opened 2025-04-24
Improve curvature reliability

SylviaWhittle:
This PR improves the reliability of the curvature measurement of molecules.

Due to how the positions of the trace coordinates after `smoothing` can be quite irregularly spaced, this messes up with the curvature calculation often causing some spikes in the curvature readings and irregularities despite the smoothing.

I have been able to mitigate this pretty well by resampling the points (using a weird form of linear interpolation - NOT what you might think, using `scipy`'s built in interpolation is not the same as what I'm doing here unfortunately).

Here is an example of a trace where the coords are rather erratically spaced

image

And this is what it looks like when resampled to be evenly spaced at 1 pixel

image

When applied to actual images displaying the curvature values, the improvement is pretty good IMO: (here the high curvature is marked in deeper beige-red colour scale)

image

image

image

image

image

image

---

- [x] Existing tests pass.
  - [x] `docs/configuration.md`
  - [x] `docs/usage.md`
- [x] Pre-commit checks pass.


SylviaWhittle:
The difference in the regtests are the number of points in the trace (expected since it's been resampled) and a small (okay) difference in contour length:

image

tests/test_entry_point.py
```diff
@@ -552,6 +558,10 @@ def test_entry_point_subprocess_help(capsys, argument: str, option: str) -> None
                 "spline",
                 "--splining-window-size",
                 "1",
+                "--splining-rolling-window-resampling",
+                "True",
+                "--splining-rolling-window-resample-regular-spatial-interval",
```
ns-rse, line 563:
```suggestion
                "--rolling-window-resample-regular-spatial-interval",
```

SylviaWhittle:
@ns-rse

**No rush**, copy of question from Slack

After a lot of trial and error I've fixed the tests, but the config CLI params are not named the same as in `default_config.yaml` (same with the rest of the `splining` args) no idea why this is the case or if it's working so thought I'd ask you.

ns-rse:
I think its because...

## `splining.py`

```python
        rolling_window_resampling: bool = False,
        rolling_window_resample_regular_spatial_interval_nm: float = 0.5,
```

## `default_config.yaml`

These match the parameters in `splining.py` and so will be passed through as `**kwargs` :thumbsup:

```yaml
  rolling_window_resampling: true # Whether to resample the trace or not.
  rolling_window_resample_regular_spatial_interval: 0.5e-9 # The spatial interval to resample the trace to in nm.
```

## `entry_point.py`

Here we add the new options that we introduce to `default_config.yaml` and the `dest` for these should match the `default_config.yaml` but don't `splining_resampling != rolling_window_resampling` (typically the argument itself I will simply replace `_` with `-`.

```python
    process_parser.add_argument(
        "--splining-resampling",
        dest="splining_resampling",
        type=bool,
        required=False,
        help="Whether to resample the trace or not.",
    )
    process_parser.add_argument(
        "--splining-resample-regular-spatial-interval",
        dest="splining_resample_regular_spatial_interval",
        type=float,
        required=False,
        help="The spatial interval to resample the trace to in nm.",
    )
...
    splining_parser.add_argument(
        "--resampling",
        dest="resampling",
        type=bool,
        required=False,
        help="Whether to resample the trace or not.",
    )
    splining_parser.add_argument(
        "--resample-regular-spatial-interval",
        dest="resample_regular_spatial_interval",
        type=float,
        required=False,
        help="The spatial interval to resample the trace to in nm.",
    )
```



## `test_entry_point.py`

These match the values in `entry_point.py` but we've recognised already that these don't map to the `**kwargs` that they are meant to and will also need correcting.

```python
                "--resampling",
                "True",
                "--resample-regular-spatial-interval",
                "1.0",
...
                "resampling": True,
                "resample_regular_spatial_interval": 1.0,
...
                "--splining-resampling",
                "True",
                "--splining-resample-regular-spatial-interval",
                "1.0",
                "--splining-step-size",
                "0.2",
...
                "splining_resampling": True,
                "splining_resample_regular_spatial_interval": 1.0,
```

## Checking

We can also check by using `grep` (although I've started using [`rg` ripgrep](https://github.com/BurntSushi/ripgrep)) to search for the strings...

```bash
❱ rg rolling_window_resampling topostats tests
topostats/validation.py
360:            "rolling_window_resampling": bool,

topostats/default_config.yaml
100:  rolling_window_resampling: true # Whether to resample the trace or not.

topostats/tracing/splining.py
265:    rolling_window_resampling : bool, optional
276:        rolling_window_resampling: bool = False,
290:        rolling_window_resampling : bool, optional
299:        self.rolling_window_resampling = rolling_window_resampling
438:            if self.rolling_window_resampling:
448:            if self.rolling_window_resampling:
561:    rolling_window_resampling: bool = False,
592:    rolling_window_resampling : bool, optional
626:                        rolling_window_resampling=rolling_window_resampling,

❱ rg rolling_window_resample_regular_spatial_interval topostats tests
topostats/validation.py
361:            "rolling_window_resample_regular_spatial_interval": lambda n: n > 0.0,

topostats/default_config.yaml
101:  rolling_window_resample_regular_spatial_interval: 0.5e-9 # The spatial interval to resample the trace to in nm.

topostats/tracing/splining.py
267:    rolling_window_resample_regular_spatial_interval_nm : float, optional
277:        rolling_window_resample_regular_spatial_interval_nm: float = 0.5,
292:        rolling_window_resample_regular_spatial_interval_nm : float, optional
300:        self.rolling_window_resample_regular_spatial_interval_nm = (
301:            rolling_window_resample_regular_spatial_interval_nm / 1e-9
441:                    interval=self.rolling_window_resample_regular_spatial_interval_nm / self.pixel_to_nm_scaling,
451:                    interval=self.rolling_window_resample_regular_spatial_interval_nm / self.pixel_to_nm_scaling,
562:    rolling_window_resample_regular_spatial_interval: float = 0.5,
594:    rolling_window_resample_regular_spatial_interval : float, optional
627:                        rolling_window_resample_regular_spatial_interval_nm=rolling_window_resample_regular_spatial_interval,
```

Not matches in `topostats/entry_point.py` nor (unsurprisingly) in `tests/test_entry_point.py`

So the values in `entry_point.py` (and the associated tests) don't match those in `default_config.yaml`. As we update the configuration based on keywords the value defined in `test_entry_point.py` as command line arguments don't update those that have been loaded from `default_config.yaml`. Perhaps we should have a check that these match to double check when we are writing these?

If we want to test the effect of different parameters from the command line for these they need to match. It's an easy way of testing the effect, although we should perhaps test the effect directly by parametrizing `test_resample_points_regular_interval()` by varying `interval` in its tests.

Sorry its not a full review, got bedtime stories to do and snooker to watch.

ns-rse:
Looks like I messed up resolving the merge conflict, sorry. Will correct that now.

ns-rse (approved):

PR #1145, opened 2025-04-30
feature: Introduce TopoStats class and saving to HDF5 .topostats

ns-rse:
As we look to move towards restructuring the `.topostats` HDF5 file I've taken the logical extension to the introduction of `ImageGrainCrops` dataclass and the nested dataclass/class by introducing a `TopoStats` dataclass which defines the top-level attributes and structure that we wish to have in `.topostats` files.

One of the disconnects at the moment is that when we run the whole processing pipeline we load whatever image using AFMReader which returns `image` and `pixel_to_nm_scale`. This is ok if we're running the whole pipeline but when it comes to later stages and attempting to load `.topostats` files which can be passed into `Grains`/`GrainStats`/`DisorderedTracing`/etc. then we will be loading a `.topostats` file which is stored as HDF5, initially loaded as a dictionary but we will want to convert all previous steps of processing to `ImageGrainCrops` and its nested structure.

That means that `LoadScans` would return different things which isn't a good idea, it should consistently return the _same_ thing each time it loads files with `LoadScans.get_data()`, to which end I think it should create a dictionary of `TopoStats` objects. If data is being loaded from anything other than `.topostats` files then each value of the
dictionary, which is a `TopoStats` object will have the following attributes...

- `image_original`
- `pixel_to_nm_scale`
- `filename`
- `topostats_version` (populated on the fly from the current version)
- `img_path`

...but the `image_grain_crops` attribute will be empty.

If loading a `.topostats` file then `image_grain_crops` will hold the results of all previously processed data.

Nothing has _yet_ changed in how files are saved, but I'm (hopefully) setting things up for these changes to be made but breaking it down into a smaller commits/pull requests to ease the review process.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1262,6 +1262,7 @@ def test_hdf5_to_dict_nested_dict_group_path(tmp_path: Path) -> None:
 
 @pytest.mark.parametrize(
     (
+        "topostats_version",
         "image",
         "pixel_to_nm_scaling",
         "filename",
@@ -1273,6 +1274,7 @@ def test_hdf5_to_dict_nested_dict_group_path(tmp_path: Path) -> None:
     ),
     [
         pytest.param(
+            "0.2",
             np.arange(0, 100).reshape(10, 10),
             3.14159265,
             "below_grain_mask_with_grain_trace_data",
@@ -1386,6 +1388,7 @@ def test_hdf5_to_dict_nested_dict_group_path(tmp_path: Path) -> None:
             id="below_grain_mask_with_grain_trace_data",
         ),
         pytest.param(
+            "0.2",
             np.arange(0, 100).reshape(10, 10),
             3.14159265,
             "above_grain_mask_without_grain_trace_data",
@@ -1405,6 +1408,7 @@ def test_hdf5_to_dict_nested_dict_group_path(tmp_path: Path) -> None:
             id="above_grain_mask_without_grain_trace_data",
         ),
         pytest.param(
+            "0.2",
             np.arange(0, 100).reshape(10, 10),
             3.14159265,
             "above_and_below_grain_masks_without_grain_trace_data",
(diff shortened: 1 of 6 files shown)


SylviaWhittle:
This is an excellent plan, and should make things a lot simpler (and easier to understand for new devs of the project!)

tests/test_io.py
```diff
@@ -1436,39 +1467,52 @@ def test_save_and_load_topostats(
...
+            "image_original": image,
+            "image": image,
+            "grain_masks": {"above": grain_mask_above, "below": grain_mask_below},
+            "grain_trace_data": grain_trace_data,
+        }
+    else:
+        # If we are testing >= 2.4.0 then topostats_object should be ImageGrainCrops object which we load from fixture
+        topostats_object = request.getfixturevalue(image)
+    # Save the file
     save_topostats_file(
```
SylviaWhittle, line 1488: a very well-put-together test

tests/test_topostats.py
```diff
@@ -0,0 +1,81 @@
...
+            0.488,
+            "2.4.0",
+            str(GRAINCROP_DIR),
+            None,
+            None,
+            id="catenanes v2.4.0",
+        ),
+    ],
+)
+def test_topostats_to_dict(
```
SylviaWhittle, line 58: Again, I really like this test, things are plugging into each other nicely!

topostats/__init__.py
```diff
@@ -22,3 +29,244 @@
...
+        if not isinstance(other, TopoStats):
+            return False
+        return (
+            self.image_grain_crops == other.image_grain_crops
+            and self.filename == other.filename
+            and self.pixel_to_nm_scaling == other.pixel_to_nm_scaling
+            and self.topostats_version == other.topostats_version
+            and self.img_path == other.image_path
+            and self.image == other.image
+            and self.image_original == other.image_original
```
SylviaWhittle, line 92: Hm.. This seems like it shouldn't work..? 

Testing equality between two `npt.NDArray` objects is not allowed via `==` I thought
 
image

SylviaWhittle (changes requested):
Another PR bringing our repo into a more rust-like, type-safe setup, thank you

This is really neat and my brain is very pleased by the amount this codifies and tidies things up.

There is the potential issue with the `__eq__` method for the TopoStats class instances though (which I could be wrong about)

And as much as I hate to say it, perhaps we could do with a test for `TopoStats.__eq__()`..? 

But for the rest of the class methods, I don't think tests for Getters / Setters is necessary unless we add custom logic to them.

topostats/__init__.py
ns-rse: Very good point, tests duly written and we're now using `np.all(self.image == other.image)` to check for equality.

SylviaWhittle (approved):

PR #1149, opened 2025-05-07
refactor(TopoStats): class > instance attributes

ns-rse:
I realised whilst working on switching to using `TopoStats` classes to pass everything around that the initial
implementation which uses class level attributes causes problems as class level attributes are shared across all
instances of the class rather than attribute level attributes which are unique to each instance of a class.

Also realised that `img_path` needs to be non-`None` otherwise `pathlib` raises a `TypeError`. Solution is to set to
`Path.cwd()` if `img_path = None`.

### DIFF
--- a/topostats/__init__.py
+++ b/topostats/__init__.py
@@ -41,7 +41,7 @@ class TopoStats:
     """
     Class for storing TopoStats objects.
 
-    Attributes
+    Parameters
     ----------
     image_grain_crops : ImageGrainCrops | None
         ImageGrainCrops of processed image.
@@ -50,7 +50,7 @@ class TopoStats:
     pixel_to_nm_scaling : str | None
         Pixel to nanometre scaling.
     img_path : str | None
-        Original path to image.
+        Original path to image, if not specified defaults to current working directory.
     image : npt.NDArray | None
         Flattened image (post ''Filter()'').
     image_original : npt.NDArray | None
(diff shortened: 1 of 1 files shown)


SylviaWhittle:
Ah I'm sorry for missing this in my review.

Just checking so I ensure my own understanding of this:

I now remember that I clocked this when writing the `GrainCrop` code, where IIRC with a `dataclass`, variables can be declared in the class and it won't share these values across instances, but in a normal Python `class`, it will.

Am I right in that understanding from what you've read?

So as I understand it, this is okay because `@dataclass`:
image

but in this, we put it in the `__init__()` as it's a `class` not a `dataclass`
image

ns-rse:
> I now remember that I clocked this when writing the GrainCrop code, where IIRC with a dataclass, variables can be declared in the class and it won't share these values across instances, but in a normal Python class, it will.

Hmm, I just thought that `dataclass` were the same as plain classes so I've checked...

```
❱ git checkout 8660ab254
❱ ipython
~~~~~~~~~~~~ AUTORELOAD ENABALED ~~~~~~~~~~~~
Python 3.11.11 (main, Jan  6 2025, 08:45:23) [GCC 14.2.1 20240910]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.34.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from topostats import TopoStats
   ...: test1 = TopoStats(img_path="./")
   ...: test2 = TopoStats(img_path="~/tmp")
   ...: test1.img_path
Out[1]: PosixPath('.')

In [2]: test2.img_path
Out[2]: PosixPath('~/tmp')

In [3]: test1.img_path = "/mnt/elsewhere"

In [4]: test1.img_path
Out[4]: PosixPath('/mnt/elsewhere')

In [5]: test2.img_path
Out[5]: PosixPath('~/tmp')
```

Looks like they're _not_ class attributes and _aren't_ updated globally. Checking the docs, to mark them as class attributes we would have to have them typed as [`typing.ClassVar`](https://docs.python.org/3/library/dataclasses.html#class-variables).

I'll revert this and just include the check and fix for ensuring `img_path` is handled correctly.

Thanks for catching that.

SylviaWhittle:
Very good investigating. I thought, like you, that the instances would share the class attribute when not a `dataclass`. I should read up on this

PR #1151, opened 2025-05-07
TopoStats classes internally and for writing/reading HDF5

### LINKED ISSUE #1102: [feature] : Save ImageGrainCrops in .topostats files

We need a way to be able to save and load `ImageGrainCrops` objects into and out of `.topostats` files for reproducible work.

@ns-rse and I have been pondering it and think a simple solution of nesting keys and data in a structure similar to that of how the classes are laid out, would be sensible. It would require boilerplate code but probably the simplest solution than trying to make in Class mechanics into the file directly.

Max is also in need of this when we can do it.

Add simple packer - unpacker functions to allow `ImageGrainCrop` class objects to be serialized into `.topostats` HDF5 format.

Trying to be clever and listing the class behavior in the file structure - I believe Tensorflow might do this with their models? Ie listing what class an object should be, as metadata alongside the data as opposed to hardcoding it in a SCHEMA in TopoStats. But this would be more planning & work.

### Additional context

N/A

### LINKED ISSUE #1143: Move some regression tests to use syrupy
There are a number of places where we  have `.pkl` or `.npy` files which store the state we wish to test against. These
work but require manually updating by uncommenting lines which over-write them, running the tests to replace these then
commenting them out again.

[`pytest-regtest`]() isn't ideal for this situation since it relies on `print(..., file="regtest")` to write the objects
to file, some of which can be large and aren't always (by default) printed in full.

However we may be able to adopt using [syrupy](https://syrupy-project.github.io/syrupy/) which can be used to
compare [dictionaries](https://til.simonwillison.net/pytest/syrupy) and in-turn the arrays that are stored within them
to replace the current method.

ns-rse:
Closes #1102
Closes #1143

This PR (draft for now) is the logical extension of the `GrainCrops` `GrainCropDirections` and `ImageGrainCrops` introduced by @SylviaWhittle  in #1022 and switches to using the `TopoStats` class @ns-rse introduced in #1145 for handling images and the derived datasets (arrays) such that the unit of interest is individual grains.

It is at the moment far from complete as the checklist shows below but because of the large amount of changes and reorganisation I was keen to share it in stages. The tests for each commit pass (thanks `pytest-testmon` :grinning: ) but until all steps are complete the integration tests (`tests_processing.py` and `tests_run_modules.py` won't pass, I'm working on them as I go through each class).

Its perhaps worth reading the commit messages for the individual commits for a little more information on the re-organisation that has been done so far.

- [X] `LoadScans`
- [X] `Filters`
- [X] `Grains`
- [X] `GrainStats`
- [X] `DisorderedTracing`
- [x] `NodeStats`
- [x] `OrderedTracing`
- [x] `Splining`
- [x] `Curvature`

Of note...

## Shared Methods

Some methods from `Grains` were used `@staticmethod` from `GrainCrops` and have been moved out to `utils`. I've setup some skeleton tests for these but they fail (get 2x5 arrays back when I would have expected 5x5 arrays from flattening 5x5x3 arrays). Couldn't see any existing tests for these.

## Documentation

I intend to document the class structures and in turn the HDF5 format these are written to.

## Syrupy

I've closed #1143 as the key test which used `.pkl`'s and required manually updating has been addressed. I think we could still switch _all_ tests to use `syrupy` (see #1152).

## AFMReader

These changes will also require modifications to AFMReader and perhaps moving `LoadScans` over but I'm wary of introducing a circular dependency and have already discussed with @SylviaWhittle this. We felt that  perhaps `AFMReader` should only load files and return dictionaries. Re-constructing these to `TopoStats` / `ImageGrainCrops` / `GrainCropsDirection` / `GrainCrop` should be the domain of TopoStats. I don't think this should be a problem for the Napari plug-in as it could import and use whatever it needs from either.

---

- [ ] Existing tests pass.
- [X] Pre-commit checks pass.
- [X] New functions/methods have typehints and docstrings.


SylviaWhittle (approved):
This is looking very good!

I've checked this out locally and just followed the threads of where data goes, and it looks very nice.

One thing, is to maybe move things out of / don't add more to `utils.py` given that IIRC we are wanting to eventually eliminate it? Perhaps a `grain_handling.py` or something?

I'll have a further look next week with Laura but looking great so far 👍

SylviaWhittle:
Oops meant to merely comment, not approve, sorry!

ns-rse:
Now have `run_nodestats` working with `TopoStats` objects, haven't yet fixed all the unit tests against this but pushing on with `ordered_tracing`, `splining` and `curvature` and will address all failing unit tests once these are done.

ns-rse:
Answers after chatting with @SylivaWhittle

1. Keep `MatchedBranches` and have `UnmatchedBranches` as a separate class.

2. Everything nested within `fwhm` is required (i.e. `fwhm` / `half_maxs` / `peaks`) and it is fine to bump these up to
be attributes of `MatchedBranches` as long as there is clear nomenclature that all three pertain to "full-width
half-maximum".


Also discussed...

Having classes with attributes isn't too dissimilar to dictionaries with key/value pairs, but with the added benefit
that the types are clearly defined (no more :duck: typing in the code base). Because the code base is growing it is
becoming harder for any one to keep it all in their head and reason through all steps and so people will be working on
one small feature that has to fit into a bigger picture. Ensuring datatypes are consistent as they are passed
around/through the analysis pipeline makes development considerably easier as errors manifest themselves earlier in the
development process and can be corrected (I've found this to be the case so far with some types being wrong in the tests
as I've eased Pydantic dataclass validation into the current branch I'm working on).

ns-rse:
Progress, same issue has affected Ordered tracing.

## `main`

The two classes that perform ordered tracing [`OrderedTraceNodestats`](https://github.com/AFM-SPM/TopoStats/blob/1c47648eac15ba1115a81179841fe73014986fa7/topostats/tracing/ordered_tracing.py#L21) and [`OrderedTraceTopostats`](https://github.com/AFM-SPM/TopoStats/blob/1c47648eac15ba1115a81179841fe73014986fa7/topostats/tracing/ordered_tracing.py#L695) both only have parameters `skeleton` although unlike `NodeStats` the doc-strings here do actually state that these should be pruned which I missed and have mistakenly been passing the initial skeleton in.

Correcting this and we are almost there, grains 11, 12, 13 and 17 have been recovered and there are still some problems. Grain 5 now isn't traced/splined correctly, and there are small gaps which either shouldn't be there or are in the wrong place in grains 3 and 9.

| `main` | `ns-rse/1102-remove-unnecessary-classes` |
|--------|------------------------------------------|
| image | image |

## Grain 5

From ordered tracing we can see that the pruned skeleton that has been passed in has had the wrong section pruned, its the major loop that is removed rather than the smaller branch.

image

## Grain 3

The ordered trace doesn't have a break in it.

image

I wonder if this is introduced during splining perhaps and in the old its on the right of the junction and the new the left.

## Grain 9

Again no break in the ordered trace.

image


I'll keep digging as I'm slowly getting closer :crossed_fingers:

ns-rse:
For grain 5 the arrays going into ordered tracing are different between the current work and main. Its possibly something to do with removal of junction points.

This particular grain/molecule though has a bigger problem associated with it though in so much as the longer arm at the top is pruned off instead of a small side loop being pruned off. This leaves a trace with two junctions which I think is also causing some problems, but addressing those is beyond the scope of my current efforts.

Will return to this next week.

PR #1156, opened 2025-05-13
Add unet confidence as a configurable parameter

SylviaWhittle:
This PR simply adds confidence as a configurable unet parameter as I have omitted in doing it before.

Also added `--unet-remove-disconnected-grains` option to the tests in `test_entry_point.py` as I'd forgotten in previous PRs.

---

- [x] Existing tests pass.
  - [x] `docs/configuration.md`
  - [na] `docs/usage.md`
  - [na] `docs/data_dictionary.md`
  - [na] `docs/advanced.md` and new pages it should link to.
- [x] Pre-commit checks pass.
- [na] New functions/methods have typehints and docstrings.
- [na] New functions/methods have tests which check the intended behaviour is correct.


docs/configuration.md
```diff
@@ -122,6 +122,7 @@ Aside from the comments in YAML file itself the fields are described below.
...
+|                                                      | `confidence`                                       | float                                              | `0.5`                       | Confidence threshold for the UNet model. Smaller is more generous, larger is more strict.                                                                                                                                                                                                                                                                                                                                                 |
```
ns-rse, line 125:
```suggestion
|                                                      | `confidence`                                       | float                                              | `0.5`                       | Confidence threshold for the UNet model. Smaller is more generous, large is stricter.                                                                                                                                                                                                                                                                                                                                                 |
```

tests/test_entry_point.py
```diff
@@ -279,6 +279,10 @@ def test_entry_point_subprocess_help(capsys, argument: str, option: str) -> None
                 "14",
                 "--unet-lower-norm-bound",
                 "12",
+                "--unet-remove-disconnected-grains",
```
ns-rse, line 282: Good work testing the effect of varying this on the entry point :+1:

topostats/entry_point.py
```diff
@@ -345,6 +345,13 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Whether or not to remove grains that are disconnected from the original mask",
     )
+    process_parser.add_argument(
+        "--unet-confidence",
+        dest="unet_confidence",
+        type=float,
+        required=False,
+        help="Confidence threshold for the UNet model. Smaller is more generous, larger is more strict.",
```
ns-rse, line 353: Ditto on wording as suggested in `configuration.md` and the line below.

(Very minor and pedantic though so feel free to ignore).

topostats/grains.py
```diff
@@ -1248,7 +1250,7 @@ def improve_grain_segmentation_unet(
             predicted_mask = predict_unet(
                 image=graincrop.image,
                 model=unet_model,
-                confidence=0.1,
```
ns-rse, line 1251: Ah, so it was hard coded in the first instance, good work making this configurable.

ns-rse (approved):
Wholly approve of removing hard-coded values and making them configurable.

Minor grammar pedantry, feel free to ignore as it doesn't make much difference.

I'm surprised changing the value from `0.1` to `0.5` hasn't impacted the tests, but then perhaps we don't have tests which cover the usage of Unet analysis, which is something we should look to remedy in the future.

topostats/grains.py
llwiggins: If 0.1 is the most sensible value for our U-Nets, would it make sense for us to set that as default rather than 0.5?

topostats/entry_point.py
ns-rse:
```suggestion
        help="Confidence threshold for the UNet model. Smaller is more generous, large is stricter.",
```

topostats/entry_point.py
```diff
@@ -845,6 +852,13 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Whether or not to remove grains that are disconnected from the original mask",
     )
+    grains_parser.add_argument(
+        "--unet-confidence",
+        dest="unet_confidence",
+        type=float,
+        required=False,
+        help="Confidence threshold for the UNet model. Smaller is more generous, larger is more strict.",
```
ns-rse, line 860:
```suggestion
        help="Confidence threshold for the UNet model. Smaller is more generous, large is stricter.",
```

ns-rse (approved):
Same tweak in two other places. I'll commit them now and this is then ready to merge :+1:

ns-rse (approved):

SylviaWhittle:
Ah the failing test is cause of the `3.10` requirements for `AFMReader`

topostats/grains.py
SylviaWhittle: Nah, it was the most sensible for a specific model, but not generally IMO. 0.5 is where the model is more sure than not, so a good default I think?

SylviaWhittle:
Forcing it through since it's just python versions that's causing the issue

PR #1157, opened 2025-05-15
Minimum Python >= 3.10

### LINKED ISSUE #1146: Bump minimum Python version to 3.10
Officially Python 3.9 will reach its end-of-life in [2025-10](https://devguide.python.org/versions/).

It causes pain with type-hints as the `|` operator isn't supported in Python < 3.10 and so we have multiple `from __future__ import annotation` littered around.

We should bump the minimum supported version to `3.10`.

I suspect most users are at our upper limit anyway of `3.11` anyway (we can't support `3.12` yet because of the dependency on Topoly and a regression bug in that which hasn't yet been resolved, see #927).

- [X] `pyproject.toml` set `requites-python = ">= 3.10, < 3.12"` plus other minimum requirements in this file for linters etc.
- [X] Remove `from __future__ import annotations` from all files under `topostats/` and `tests/` (see output of `rg "__future__"` for a list of files).
- [X] Clear statement in docs and `README.md` as to minimum version would also be useful (don't think we currently have this in place)
- [X] Remove `3.9` from the matrices of workflows under `.github/workflows/*.yaml`

## Removing `from __future__ import annotations`

The following was used from the root of the `TopoStats` repository on the `main` branch to find the files that need modifying and formed the basis of the subsequent checklist (NB - `rg` is [ripgrep](https://github.com/BurntSushi/ripgrep)).

``` shell
(topostats-afmreader) ❱ rg "from __future__" topostats/ tests/
__init__.py
3:from __future__ import annotations

unet_masking.py
3:from __future__ import annotations

plottingfuncs.py
3:from __future__ import annotations

io.py
3:from __future__ import annotations

plotting.py
3:from __future__ import annotations

measure/curvature.py
3:from __future__ import annotations

grainstats.py
3:from __future__ import annotations

utils.py
3:from __future__ import annotations

filters.py
3:from __future__ import annotations

run_modules.py
8:from __future__ import annotations

measure/geometry.py
3:from __future__ import annotations

measure/height_profiles.py
3:from __future__ import annotations

measure/feret.py
12:from __future__ import annotations

grains.py
4:from __future__ import annotations

processing.py
3:from __future__ import annotations

tracing/splining.py
3:from __future__ import annotations

tracing/disordered_tracing.py
3:from __future__ import annotations

tracing/tracingfuncs.py
3:from __future__ import annotations

tracing/pruning.py
3:from __future__ import annotations

tracing/ordered_tracing.py
3:from __future__ import annotations

tracing/nodestats.py
3:from __future__ import annotations

../tests/test_plotting.py
3:from __future__ import annotations

../tests/conftest.py
5:from __future__ import annotations

../tests/test_classes.py
3:from __future__ import annotations

../tests/test_topostats.py
3:from __future__ import annotations

../tests/tracing/test_pruning.py
3:from __future__ import annotations

../tests/test_unet_masking.py
3:from __future__ import annotations

../tests/tracing/test_disordered_tracing.py
5:from __future__ import annotations

../tests/measure/test_height_profiles.py
3:from __future__ import annotations

../tests/test_grains.py
3:from __future__ import annotations

../tests/measure/conftest.py
3:from __future__ import annotations

../tests/measure/test_feret.py
3:from __future__ import annotations
```

Anyone who is working on new features/tests should make sure they remove these lines from any files they have
introduced.

- [X] `filters.py`
- [X] `grains.py`
- [X] `grainstats.py`
- [X] `__init__.py`
- [X] `io.py`
- [X] `measure/curvature.py`
- [X] `measure/feret.py`
- [X] `measure/geometry.py`
- [X] `measure/height_profiles.py`
- [X] `plottingfuncs.py`
- [X] `plotting.py`
- [X] `processing.py`
- [X] `run_modules.py`
- [X] `tracing/disordered_tracing.py`
- [X] `tracing/nodestats.py`
- [X] `tracing/ordered_tracing.py`
- [X] `tracing/pruning.py`
- [X] `tracing/splining.py`
- [X] `tracing/tracingfuncs.py`
- [X] `unet_masking.py`
- [X] `utils.py`
- [X] `tests/conftest.py`
- [X] `tests/measure/conftest.py`
- [X] `tests/measure/test_feret.py`
- [X] `tests/measure/test_height_profiles.py`
- [X] `tests/test_classes.py`
- [X] `tests/test_grains.py`
- [X] `tests/test_plotting.py`
- [X] `tests/test_topostats.py`
- [X] `tests/test_unet_masking.py`
- [X] `tests/tracing/test_disordered_tracing.py`
- [X] `tests/tracing/test_pruning.py`

### LINKED ISSUE #1155: Remove reference to topotracing
I noticed we have `topostats/__main__.py` which imports `topostats.topotracing()` and makes it available.

Looking at the [history](https://github.com/AFM-SPM/TopoStats/commits/main/topostats/__main__.py) reveals this is very
old code which is redundant and `__main__.py` should be removed.

Further, the example in `filters.Filter.filter_image()` also references this on line
[518](https://github.com/AFM-SPM/TopoStats/blob/0034b26454878974dc39b28a15a9d6146d2d4a93/topostats/filters.py#L518) and
should be changed to read...

``` python
        from topostats.filters import Filter
        from topostats.processing import process_scan
```

ns-rse:
Closes #1146
Closes #1155

- #1146 Drop support for Python 3.9 and set minimum version to Python 3.10.
- #1155 Remove reference to topotracing
- Upgrades minimum version of Numpy to >=2.0.0 in light of [topoly-1.1.0](https://github.com/ilbsm/topoly_tutorial/issues/4#issuecomment-2863318755) which works with Numpy >=2.0.0 and new versions of Matplotlib (removed restriction on versio of later previously).
- Switched to using a forked version of `numpyencoder` which I forked to the `AFM-SPM` organisation in light of a couple of minor bugs in their `0.3.0` version that occur when using Numpy >= 2.0.0 because of deprecated types.
- Addressed a number of linting and typehint issues that were raised as a consequence of these changes/the files being touched again (they perhaps crept through as when the changes were made `pre-commit` hooks weren't being used as `pylint` hook _doesn't_ run in CI, although some were throwing errors when running the test suite so unsure how they persisted!)

See comments in individual commits for greater details. Attempted to group changes into atomic commits and have added the linting commit to `.git-blame-ignore-revs` so that the stylistic changes aren't attributed to me and we can see who wrote the original code.

---

- [X] Existing tests pass.
- [X] Documentation has been updated and builds. Remember to update as required...
  - [X] `docs/configuration.md`
  - [X] `docs/usage.md`
  - [ ] ~`docs/data_dictionary.md`~
  - [ ] ~`docs/advanced.md` and new pages it should link to.~
- [X] Pre-commit checks pass.
- [ ] ~New functions/methods have typehints and docstrings.~
- [ ] ~New functions/methods have tests which check the intended behaviour is correct.~

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,3 +1,6 @@
+# ruff/pylint 2025-05-15 post removal of support for Python 3.9
+69ae81af3318d50cc89a610461a5e5ee997a6e7a
+
 # nodestats codespell corrections
 f13f8d40d6768f14a46315d105b3d80beaa33aaf
 284f113cf8416eed43b88b4c438d7552456616e3
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: 3.11
       - name: Installing the package
         run: |
           pip3 install .
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -55,4 +55,4 @@ jobs:
       - name: Setup tmate session
         if: ${{ failure() }}
         uses: mxschmitt/action-tmate@v3
-        timeout-minutes: 5
+        timeout-minutes: 25
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.10", "3.11"]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -55,4 +55,4 @@ jobs:
       - name: Setup tmate session
         if: ${{ failure() }}
         uses: mxschmitt/action-tmate@v3
-        timeout-minutes: 25
+        timeout-minutes: 90
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -55,4 +55,4 @@ jobs:
       - name: Setup tmate session
         if: ${{ failure() }}
         uses: mxschmitt/action-tmate@v3
-        timeout-minutes: 90
+        timeout-minutes: 5
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -15,5 +15,6 @@ globs:
 
 ignores:
   - "tmp/"
+  - "paper.md"
 
 fix: true
--- a/.pylintrc
+++ b/.pylintrc
@@ -542,10 +542,10 @@ exclude-too-few-public-methods=
 ignored-parents=
 
 # Maximum number of arguments for function / method.
-max-args=7  # Raised this from 5
+max-args=10  # Raised this from 5
 
 # Maximum number of attributes for a class (see R0902).
-max-attributes=7
+max-attributes=10
 
 # Maximum number of boolean expressions in an if statement (see R0916).
 max-bool-expr=5
--- a/.pylintrc
+++ b/.pylintrc
@@ -564,7 +564,7 @@ max-parents=7
 max-public-methods=20
 
 # Maximum number of return / yield for function / method body.
-max-returns=6
+max-returns=8
 
 # Maximum number of statements in function / method body.
 max-statements=50
--- a/README.md
+++ b/README.md
@@ -47,6 +47,9 @@ pip install topostats
 For more on installation and how to upgrade please see the [installation
 instructions](https://afm-spm.github.io/TopoStats/main/installation.html).
 
+**NB** The minimum supported version of Python is >=3.10 and because of a constraint in a dependency the maximum
+supported version is <= 3.11 (for now, we hope to support newer versions in the near future).
+
 ## How to Use
 
 ### Tutorials and Examples
@@ -87,7 +90,7 @@ to run the processing again. See the documentation on
 
 ## Contributing
 
-See [contributing guidelines](https://afm-spm.github.io/TopoStats/main/contributing.html).
+Please refer to our [contributing guidelines](https://afm-spm.github.io/TopoStats/main/contributing.html) documentation.
 
 ## Licence
 
@@ -95,7 +98,8 @@ See [contributing guidelines](https://afm-spm.github.io/TopoStats/main/contribut
 
 ## Citation
 
-Please use the [Citation File Format](https://citation-file-format.github.io/) which is available in this repository.
+If you use TopoStats in your work or research please cite us. There is a [Citation File
+Format](https://citation-file-format.github.io/) in this repository to aid citation.
 
 ### Publications
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -13,12 +13,9 @@ issue](https://github.com/AFM-SPM/TopoStats/issues/new/choose) using one of the 
 
 ### Cloning the repository
 
-If you wish to make changes yourself you will have to
-[fork](https://docs.github.com/en/get-started/quickstart/fork-a-repo) the repository to your own account and then [clone
-that](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) if you are not
-a member of AFM-SPM Organisation. If you are a member then you can [clone the
-repository](https://docs.github.com/en/repositories/creating-and-managing-repositories/cloning-a-repository) and make
-contributions directly.
+If you wish to make changes yourself you will have to [fork][git_fork] the repository to your own account and then
+[clone that][git_clone] if you are not a member of AFM-SPM Organisation. If you are a member then you can [clone the
+repository][git_clone] and make contributions directly.
 
 ```bash
 # Member of AFM-SPM Organisation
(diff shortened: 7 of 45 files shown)


SylviaWhittle (approved):
Looks good to me, appreciate the tidy up too. Type accuracy is very pleasing. Also the `from __future__ import annotations` is great to get rid of

I assume the issue regarding pySPM is still ongoing, so approving so that once it's resolved, it can be merged.

ns-rse:
> I assume the issue regarding pySPM is still ongoing, so approving so that once it's resolved, it can be merged.

I think I've solved that by forking to [AFM-SPM/pySPM](https://github.com/AFM-SPm/pySPM), updating the Numpy version and having `AFMReader` install from there (see [#147 on AFMReader](https://github.com/AFM-SPM/AFMReader/pull/147)).

But the tests fail on OSX/Windows for some other reason that I've not yet understood/got to the bottom of. :sob:

ns-rse:
I've tracked one of the issues down and we have a precision error...

```
FAILED tests/test_filters_minicircle.py::test_get_threshold_stddev - AssertionError: assert defaultdict(None, {'below': [np.float64(-7.484708050736528)], 'above': [np.float64(0.4990958836019106)]}) == approx({'below': [-7.484708050736529],'above': [0.4990958836019106]})
```

I've therefore tried switching to `topostats.io.dict_almost_equal()` but we still get the same error as the `abs_tol` isn't yet applied to elements of a list...

```
FAILED tests/test_filters_minicircle.py::test_get_threshold_stddev - AssertionError: assert False
 +  where False = dict_almost_equal({'below': [np.float64(-7.484708050736528)], 'above': [np.float64(0.4990958836019106)]}, {'below': [np.float64(-7.484708050736529)], 'above': [np.float64(0.4990958836019106)]}, abs_tol=1e-11)
 +    where {'below': [np.float64(-7.484708050736528)], 'above': [np.float64(0.4990958836019106)]} = <topostats.filters.Filters object at 0x139f93d60>.thresholds

```

Switching the `LOGGER.debug` statements to `LOGGER.info` as I can't at the moment remember how to turn on debugging when running `pytest` and I wanted to check before the `tmate` session timed out...

```
INFO     topostats:io.py:95 Comparing key below
INFO     topostats:io.py:112 Key below not equal: [np.float64(-7.484708050736528)] != [np.float64(-7.484708050736529)]
```
...shows that the `abs_tol` and `np.close()` to which the parameter is passed is not used here.

I've extended `topostats.io.dict_almost_equal()` to perform pairwise element comparison when the values of a dictionary are a list.

That will hopefully remove one point of failure, next to work out why the dataframes are missing columns! :thinking:

SylviaWhittle:
(lukewarm defense of the deprecation decision)
Okay so to be fair to the numpy devs on this:

It kind of doesn't make sense to be able to cross product a 2D vector since a cross produce naturally occurs in 3D space. Rather we are trying to cross produce two 3D vectors that happen to lie in a 2D plane.

> "In [mathematics](https://en.wikipedia.org/wiki/Mathematics), the cross product or vector product (occasionally directed area product, to emphasize its geometric significance) is a [binary operation](https://en.wikipedia.org/wiki/Binary_operation) on two [vectors](https://en.wikipedia.org/wiki/Euclidean_vector) in a three-dimensional [oriented](https://en.wikipedia.org/wiki/Orientation_(vector_space)) [Euclidean vector space](https://en.wikipedia.org/wiki/Euclidean_vector_space)"

**However** it is **very** irksome that `numpy` removes convenience and forces us to implement solutions ourselves it seems.

I believe this thread discusses the deprecation. https://github.com/numpy/numpy/issues/26620

Potential solutions:

We could manually calculate the cross product of the 2D vectors:
```
def cross_product_2d(v1: list | npt.NDArray, v2: list | npt.NDArray) -> float:
    """Calculate the 2D cross product of two vectors."""
    if isinstance(v1, list):
        v1 = np.array(v1)
    if isinstance(v2, list):
        v2 = np.array(v2)
    return v1[0] * v2[1] - v1[1] * v2[0]

cross_product_2d_result = cross_product_2d(v1, v2)
print("2D cross product of v1 and v2:", cross_product_2d_result)
```

```
>>> 2D cross product of v1 and v2: -2
```

Or we could convert the 2D vectors to 3D:
```
v1 = [1, 2]
v2 = np.array([3, 4])

def vector_2d_to_3d(vector: list | npt.NDArray) -> npt.NDArray:
    """Convert a 2D vector to a 3D vector by adding a z-component of 0."""
    if isinstance(vector, list):
        vector = np.array(vector)
    return np.append(vector, 0)

cross_product = np.cross(vector_2d_to_3d(v1), vector_2d_to_3d(v2))
print("Cross product of v1 and v2:", cross_product)
```

```
Cross product of v1 and v2: [ 0  0 -2]
```

(nerdy maths stuff)
Interestingly, the ability to compute the cross product in 2D to produce a scalar like this, is sort of a neat mathematical coincidence rather than intended. As formally, the algorithm to compute is is:
image

But in 2D vectors, any vector component with the subscript `3` here, is zero because they are at 0 z. If you look at the components, you'll see that the top two contain multiplications with a subscript `3` term, hence multiplication with zero, and the end result is a column vector:
```
0
0
s
```
where s is the scalar result you typically see in a 2D cross product.

When the result of a cross product is used as a scalar, the mathematical intent is to use the magnitude of the resulting vector orthogonal to the two input vectors, which in the planar (2D) case, happens to be simply equal to the 3rd component, hence the reduction to the commonly used formula: `(x1,y1)×(x2,y2) := x1y2−x2y1` where the other vector components can be dropped because they are multiplied by zero (the z component) which is negated due to being planar.

When vectors in 3D are used, the signed magnitude of the resulting vector is no longer simply the scalar in the z direction, but the normalized (`np.linalg.norm`) of the vector. Hence the use of it.

I believe the previous implementation in `numpy` returned a 3D vector from a `np.cross` of 2D vectors, **assuming** that the 3rd, not provided component to the 2D vectors was 0. This change in `numpy` now means we have to be explicit that our third dimension is zero.

SylviaWhittle (approved):

PR #1158, opened 2025-05-20
Updated documentation for grainstats and grain_finding

llwiggins:
Closes #1121 
This PR updates documentation for grainstats and grain_finding to align with recent multi-class functionality within the codebase. I've added information on the new format of the `all_statistics.csv`, defining what we mean by `grain_number`, `class_number` and `subgrain_number`. I've also added a summary and diagram of the vetting steps within grain_finding.md. Please let me know if these changes make sense or if any further clarification is needed, the vetting steps in particular can be quite tricky to get your head around so I'm happy to add more examples here if you think more explanation is needed!

Before submitting a Pull Request please check the following.

- [x] Existing tests pass.
- [x] Documentation has been updated and builds. Remember to update as required...
  - [ ] `docs/configuration.md`
  - [ ] `docs/usage.md`
  - [ ] `docs/data_dictionary.md`
  - [x] `docs/advanced.md` and new pages it should link to.
- [x] Pre-commit checks pass.
- [x] New functions/methods have typehints and docstrings.
- [x] New functions/methods have tests which check the intended behaviour is correct.

### `topostats/default_config.yaml`

If adding options to `topostats/default_config.yaml` please ensure.

- [ ] There is a comment adjacent to the option explaining what it is and the valid values.
- [ ] A check is made in `topostats/validation.py` to ensure entries are valid.
- [ ] Add the option to the relevant sub-parser in `topostats/entry_point.py`.

PR #1160, opened 2025-05-20
Remove Python 3.9 from workflows

ns-rse:
As there appear to be some issues with the full purging of Python 3.9 from TopoStats support (see #1157 where these are currently failing/waiting to be addressed) the interim solution here is to simply remove Python 3.9 from the workflows defined under `.github/workflows`.

Once merged @llwiggins should be able to rebase onto this branch to update `llwiggins/multiclass_documentation` and push the changes. PR #1158 should then pass the test suite. :fingers_crossed_tone1:

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -21,7 +21,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: 3.11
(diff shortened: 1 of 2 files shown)


ns-rse:
Thanks @llwiggins 

Unfortunatelythis has the same errors as #1157 despite only removing from Python 3.9 from the test matrix.

I'm going to close this for now and focus on solving the issues on #1157.

Issue #1152, opened 2025-05-09
Switch all regression tests to syrupy


ns-rse:
Further to #1143 which has been (or will be) closed by changes in #1151 I think we should consider looking at switching all of our tests from [`pytest-regtest`]() to [`syrupy`](https://syrupy-project.github.io/syrupy/).

Bit of a chore but will hopefully give finer grained insight onto difference in data frames and dictionaries (you can see an example of comparing dictionaries in the [syrupy video](https://asciinema.org/a/369462).

The code for doing so is I feel easier to understand...

```
    def test_syrupy(snapshot):
    assert "something" == snapshot
```

On first running this will fail because no snapshot exists so you have to run...

```
pytest --snapshot-update
```

This creates a directory `__snapshot__` which contains a file for each test that uses snapshots (remember to include these in the git commit and remove the related file under `tests/_regtest_outputs` when switching).

We can see which files need updating by looking in `tests/_regtest_outputs`...

```
❱ l tests/_regtest_outputs 
drwxr-xr-x neil neil 4.0 KB Tue Apr 29 15:26:46 2025  .
drwxr-xr-x neil neil 4.0 KB Fri May  9 10:35:03 2025  ..
.rw-r--r-- neil neil 1.4 KB Wed Apr  9 10:40:12 2025  test_grainstats_minicircle.test_grainstats_regression.out
.rw-r--r-- neil neil 560 B  Wed Oct  4 14:26:12 2023  test_plotting.test_computeStats.out
.rw-r--r-- neil neil 2.9 KB Tue Apr 29 15:26:46 2025  test_processing.test_process_scan_above.out
.rw-r--r-- neil neil 1.4 KB Mon Dec 16 15:53:46 2024  test_processing.test_process_scan_above_height_profiles.out
.rw-r--r-- neil neil 1.3 KB Wed Apr 23 09:41:52 2025  test_processing.test_process_scan_below.out
.rw-r--r-- neil neil 254 B  Mon Dec 16 15:53:46 2024  test_processing.test_process_scan_below_height_profiles.out
.rw-r--r-- neil neil 3.0 KB Tue Apr 29 15:26:46 2025  test_processing.test_process_scan_both.out
```

Thus the files/tests that need updating are...

- [ ] `test_grainstats_minicircle.py::test_grainstats_regression`
- [ ] `test_plotting.py::test_computeStats`
- [ ] `test_processing.py::test_process_scan_above`
[...]

PR #438, opened 2023-01-18
Adding pre-commit.ci badge

ns-rse:
Closes #357 

Added [pre-commit.ci](https://pre-commit.ci) and badge to show if its passing.

This should run `pre-commit` on pull requests here on GitHub.

PR #1165, opened 2025-05-30
Fix pytest approx breaking when using lists in dicts

SylviaWhittle:
This PR fixes an issue in `test_filters_minicircle.py::test_get_threshold_stddev` where the result was being compared using `pytest.approx`, however `pytest.approx` does not support comparing dictionaries with lists inside them.

See here:
image

I went to use `dict_almost_equal` but I remembered that I didn't add support for `lists`. So I have edited `io.py::dict_almost_equal` to accept `list`s and to be able to compare them assuming they are flat, and contain only numbers. (Couldn't be bothered to write it robust for nested lists or complex data types). 

I also added a lil check in `dict_almost_equal` to ensure the values of the dictionaries are the same type, as I didn't consider this when writing it.

---

- [x] Existing tests pass.
  - [x] `docs/configuration.md`
  - [x] `docs/usage.md`
- [x] Pre-commit checks pass.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -161,7 +161,7 @@ def lists_almost_equal(list1: list, list2: list, abs_tol: float = 1e-9) -> bool:
         LOGGER.debug(f"Lists not same length: {len(list1)} != {len(list2)}")
         return False
     for i, (item1, item2) in enumerate(zip(list1, list2)):
-        if isinstance(item1, int | float | np.int64)) and isinstance(item2, int | float | np.int64):
+        if isinstance(item1, int | float | np.int64) and isinstance(item2, int | float | np.int64):
             if not np.isclose(item1, item2, atol=abs_tol):
                 LOGGER.debug(f"List item {i} not equal: {item1} != {item2}")
                 return False


tests/test_filters_minicircle.py
```diff
@@ -30,21 +31,21 @@ def test_remove_quadratic_unmasked(minicircle_initial_quadratic_removal: Filters
...
 
-def test_get_threshold_otsu(minicircle_threshold_otsu: np.array) -> None:
+def test_get_threshold_otsu(minicircle_threshold_otsu: Filters) -> None:
     """Test calculation of threshold."""
     assert isinstance(minicircle_threshold_otsu.thresholds, dict)
     assert minicircle_threshold_otsu.thresholds["above"] == pytest.approx(expected=[0.40108209618256363])
 
 
-def test_get_threshold_stddev(minicircle_threshold_stddev: np.array) -> None:
+def test_get_threshold_stddev(minicircle_threshold_stddev: Filters) -> None:
```
ns-rse, line 40:
```suggestion
def test_get_threshold_stddev(minicircle_threshold_stddev: Filters, snapshot) -> None:
```

tests/test_filters_minicircle.py
```diff
@@ -30,21 +31,21 @@ def test_remove_quadratic_unmasked(minicircle_initial_quadratic_removal: Filters
...
     """Test calculation of threshold."""
     assert isinstance(minicircle_threshold_stddev.thresholds, dict)
-    assert minicircle_threshold_stddev.thresholds == pytest.approx(
-        {"below": [-7.484708050736529], "above": [0.4990958836019106]}
+    assert dict_almost_equal(
```
ns-rse, line 43: This is the sort of situation where I think [syrupy](https://github.com/syrupy-project/syrupy) can be really useful.

```suggestion
    assert minicircle_threshold_stddev.thresholds == snapshot
```

ns-rse (commented):
I think this is where adopting [syrupy](https://syrupy-project.github.io/syrupy/) can be useful (#1152).

It works perfectly for `dict` / `list` / almost any native Python object, it has some issues with Numpy arrays (and in turn Pandas) but they can be handled by writing to csv first (with desire precision).

Need to include the cached directory that `syrupy` creates in the Git repo so the CI tests will pass (I found that out the hard way!).

SylviaWhittle:
This seems ready. 

I would be in favor of switching to `syrupy` if they have an option for tolerance, since these values are floats there is going to be floating point error. 

@ns-rse can you see a way around the floating point error issue with syrupy? Apologies for pushing back a little, I'm not opposed to using it, just given time pressure, I tend to stick with what I know at the time.

ns-rse:
> I would be in favor of switching to syrupy if they have an option for tolerance, since these values are floats there is going to be floating point error.
>
> @ns-rse can you see a way around the floating point error issue with syrupy? Apologies for pushing back a little, I'm not opposed to using it, just given time pressure, I tend to stick with what I know at the time.

Searching and reading through the Syrupy issues (as this is unlikely to have been something that no one wanted to do before) led me to this thread [How to use pytest.approx() with syrupy · Issue #438 · syrupy-project/syrupy](https://github.com/syrupy-project/syrupy/issues/438) which proposes rounding prior to comparing the sanpshot.

How easy that would be for nested dictionaries and lists? Well it would require a recursive function to round all numbers and we already have the recursive function that does the comparison so there seems little point in replacing that.

Perhaps as we move away from using dictionaries to dataclasses we can gradually switch to `syrupy` ?

ns-rse (commented):
Back on TopoStats today and I've realised I encountered this and came up with an alternative solution in https://github.com/AFM-SPM/TopoStats/pull/1157/commits/ddf604f4f6f579cb9f5838661d05e233fa30852c on my open PR.

I have no preference as to which solution to take, breaking the functionality out into a separate, and tested, function is probably more robust.

I'm still amazed/baffled that we are having to implement this method of comparing dictionaries and lists with specified precision as I can't believe that no one in the vast world of Python has not had to deal with the issues that arise before and come up with a usable solution. It feels like the sort of thing that should be handled in core, perhaps with the `__eq__` method.

Equality of dictionary values and how it works isn't uncommon though [Equivalence of Python dict values - Python Help - Discussions on Python.org](https://discuss.python.org/t/equivalence-of-python-dict-values/14834/7) (probably the most useful post is from [Raymond Hettinger](https://discuss.python.org/t/equivalence-of-python-dict-values/14834/5)).

ns-rse:
Happy to go with your tested solution (its better than mine anyway!).

I needed my solution for the tests to pass on #1157 though and we've now solved the cross product deprecation of 2D vectors.

I suggest that #1157 is merged, this is merged and then we can `git revert` https://github.com/AFM-SPM/TopoStats/commit/0b7743c13545d511e7860bfea04556b5c6816453 (as I've managed to keep the commits fairly atomic although it does include bumping `max-returns = 8` in `.pylintrc` so I'll make sure after the reversion to add that back in. Written up in #1170

ns-rse:
There is a merge conflict which once resolved might avoid the need for a revert.

I can have a look at resolving the conflict tomorrow morning if you want (but Tue/Wed are my current TopoStats days so won't be able to spend too long on it).

ns-rse:
Ah neat, I always forget that merge conflicts are kind of like a rebase so tests all pass \o/

One small complaint from `ruff` about using `X | Y` within `isinstance()` rather than using tuples and this will be ready to merge :+1:

topostats/io.py
```diff
@@ -88,42 +89,89 @@ def dict_almost_equal(dict1: dict, dict2: dict, abs_tol: float = 1e-9):  # noqa:
...
+    ------
+    NotImplementedError
+        If the items in the lists are not of type int or float.
+    """
+    # Check if both lists are the same length and only contain numbers
+    if len(list1) != len(list2):
+        LOGGER.debug(f"Lists not same length: {len(list1)} != {len(list2)}")
+        return False
+    for i, (item1, item2) in enumerate(zip(list1, list2)):
+        if isinstance(item1, int | float | np.int64)) and isinstance(item2, int | float | np.int64):
```
ns-rse, line 164:
```suggestion
        if isinstance(item1, int | float | np.int64) and isinstance(item2, int | float | np.int64):
```

SylviaWhittle:
> Sorry for the errant `)` test pass good to go.
> 
> 
> 
> Thanks for your patience @SylviaWhittle :+1: 

Not at all, thank **you** for your help

PR #1166, opened 2025-05-30
Handling empty grainstats df

llwiggins:
This PR addresses a bug found earlier this week where an empty grainstats_df data frame is passed to the remaining processing stages, resulting in the following error: 

`KeyError: "The following id_vars or value_vars are not present in the DataFrame: ['basename']"`

The suggested solution was to check if grainstats_df is empty and, if it is, then skip `disordered_tracing`, `node_stats`, `ordered_tracing`, `splining`, `curvature_stats` and `image_statistics`. 

The code in this PR offers one approach to this but I'm sure there are many others so would appreciate any suggestions on better ways to approach this - keeping this as a draft PR for now. It's also unclear what caused grainstats to produce an empty data frame in the first place as we have struggled to reproduce this error on different computers/environments. 

---

- [x] Existing tests pass.
- [] Documentation has been updated and builds. Remember to update as required...
  - [ ] `docs/configuration.md`
  - [ ] `docs/usage.md`
- [x] Pre-commit checks pass.


ns-rse:
This is related to the issue I'm seeing in #1165 which in that case appears to be a problem with calculating ferret statistics because of a deprecation in Numpy 2.0 (although its weird the tests pass under Linux but not on OSX/Windows).

I think we should investigate where `basename` was introduced as I thought we had empty `grainstats_df` correctly handled a long time ago.

topostats/processing.py
```diff
@@ -1150,6 +1150,18 @@ def process_scan(
...
 
+        if grainstats_df is None:
+            LOGGER.warning(f"[{topostats_object['filename']}] : Grainstats dataframe is empty, stopping processing.")
+            # Early return to stop further processing
+            return (
+                topostats_object["img_path"],
+                grainstats_df,
+                height_profiles,
+                {},  # empty image_stats dictionary
+                pd.DataFrame(),  # empty disordered_tracing_stats DataFrame
```
ns-rse, line 1161: We have the `topostats.utils.create_empty_dataframe()` function which creates empty dataframes with the expected columns (pulled from the dictionary `COLUMN_SETS` which is defined at the top of the module (it includes sets for both `disordered_tracing_statistics` and `mol_statistics`).

```suggestion
                create_empty_dataframe(column_set="disordered_tracing_statistics"),  # empty disordered_tracing_stats DataFrame
```

Can use similar for the empty `molstats_df`

(Function is already imported)

PR #1167, opened 2025-05-30
Attempt to fix grain crop plotting size feature

SylviaWhittle:
This is a poor attempt at trying to fix the ability to plot grain crops at a specific size (eg 40nm x 40nm)

I've added lots of stuff

A new module for putting all our array manipulation scripts (`array_manipulation.py`) containing new functions for dynamic re-cropping of arrays and padding bounding boxes (properly this time).

Things are tested. Though several tests are failing on my local machine due to strange issues with numpy versions and the #1165 issue.

Upon trying on a local dataset, it works and is robust to grain crops being near the edge of the image.

If a requested grain crop, eg, 150x150nm is too big for the wider image, we skip plotting it, upon request by the experimentalists (thank you for your input, @TFirth2 & @tcatley).

!! issue !!
There is a major issue with this PR that I do not know how to solve:
- Notice that in `processing.py::run_grains` I need to pass the `grain_crop_plot_size_nm` argument in to be able to re-crop the grain crops for plotting.
- But the only ways to get arguments into `run_grains` from the config file are in `grain_config` and `plotting_config`. 
- Both of these dictionaries get unpacked into the `Grains` and `Images` classes and so need entries in the `__init__` methods respectively. 
- The issue with this is that the argument is not used in the class, resulting in a useless parameter being passed to the class. 
- I initially put it in `grains_config` but since decided to put it in `plotting_config` since it's more plotting relevant. **I have no clue what we should do here. I don't like having this hanging unused parameter in the `Images` class but I don't see a way around it...

---

Here's an example of the resulting 150x150 crop just for demonstration purposes:

image

---

- [x] Existing tests pass.
  - [ ] `docs/configuration.md`
  - [ ] `docs/usage.md`
- [ ] Pre-commit checks pass.


SylviaWhittle:
I'll add the configuration options for `grain_crop_plot_size_nm` if / when we decide what the best solution is to the issue of passing the config argument to `run_grains` without having to pass it to the `Images.__init__` method. (and using `.pop` is a bit tricky because of the amount of methods that use `**` unpacking into `Images.__init__()`..

ns-rse (approved):
Don't think there is anything preventing this from being merged now and looks ready to me.

PR #1176, opened 2025-06-17
ci(pypi): Improve PyPI publishing workflow

### LINKED ISSUE #1173: Improve pypi.yaml build
One of the developers of the `pypa/gh-action-pypi-publish` has highlighted a couple of
[issues](https://github.com/pypa/gh-action-pypi-publish/issues/363#issuecomment-2953804310) with
`.github/workflows/pypi.yaml` that we should address.

Noting here so I don't forget to do so, but the same changes should be applied to
[AFMReader](https://github.com/AFM-SPM/AFMReader) and [napari-AFMReader](https://github.com/AFM-SPM/napari-AFMReader)
where the same `pypi.yaml` has been used.

ns-rse:
Closes #1173

Following the [Publishing package distribution releases using GitHub Actions CI/CD workflows - Python Packaging User
Guide](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)
I've hopefully improved out workflow for building and releasing TopoStats to [PyPI](https://pypi.org/p/topostats).

- Removes duplicated building of the dist files.
- Isolates building from uploading of artifact into separate jobs.
- Runs on every push, but...
- Publishing to PyPI job only runs on tagged commits.
- Adds a job for publishing to TestPyPI which will run on every `push` (see note below).

Having the package build and publish to [TestPyPI](https://testpypi.org/p/topostats) might come in handy if we want an
easy way to guide people to installing the development version version rather than advising them to `git clone`.

```shell
pip install --index-url https://testpypi.org/simple topostats --extra-index-url https://pypi.org/simple
```

I'm deliberately _not_ updating this documentation just yet as I would like to test this out and work out if its
possible so that I get the documentation correct. I'll note this as a separate issue.


SylviaWhittle:
Nice! 



One thing, you say it runs on every push, from the code and context I assume you mean runs on push to a PR against `main` and not on every push in every branch?



Also just checking, do we anticipate any issues with GitHub getting annoyed at us for calling too much?

ns-rse:
You were quick off the mark today @SyllviaWhittle!

Yes, I realised after lazily copying & pastaing that we wouldn't want to be running the job on every `push` and
corrected it to run on `pull-request` that target `main` only.

I'm not sure this is going to work well on TestPyPI though because of the automated versioning we get from
`setuptools_scm` as I also realised after actually engaging my brain, but I've noted that in #1177.

PR #1182, opened 2025-06-24
Support old Bruker file extensions

### LINKED ISSUE #1181: Add support for old-style Bruker file formats
Over on [AFMReader](https://githbu.com/AFM-SPM/AFMReader) we have had a   [request](https://github.com/AFM-SPM/AFMReader/issues/157) from @tcately who is working with collaborators to support the old-style file-names for Bruker machines.

These carry the extension `.###` where `#` are numbers e.g. `.001`, `.002`, see [examples](https://drive.google.com/drive/folders/1CvNDgcTpzYNCEAVT0u9aNWNBsPgpL3jH) where the filenames are...

``` shell
tc_ls6_3007.002
tc_ls6_3007.004
```

As this small sample shows the file "stem" (the bit before the period) is the same for two files so we can not expect users to change the file extension without considerable hassle.

There are two solutions...

## Search for files with `\.\d+$`

The regex [`\.\d+$`](https://regex101.com/r/A24rgJ/1) should match and file that ends with a `.` period followed by one or more numbers.

We can modify the `io.find_files()` function to accept this as a valid regular expression and ensure that matches the `LoadScans.load_spm()` method is called with such matches.

## Convenience function

We could also add a convenience function that uses `io.find_files()` to find such files matching the above regular expression and then appends to them the suffix `.spm`.

Probably redundant really but kind of makes for a cleaner solution for users as files than have a recognised [filename extension](https://en.wikipedia.org/wiki/Filename_extension) that can be associated with a given application on an operating system.

Given that we need to add regex handling to `io.find_files()` anyway it would make sense to add a function to do this renaming and provide an entry point for users to batch rename their files.

ns-rse:
Closes #1181

- Includes old format bruker file extensions (`.001`, `.002`, `.###`) when `file_ext: spm` in configuration file (or
  explicitly on the command line with `topostats --file-ext spm`).
- Parameterizes `tests.test_io.test_find_files()` to check all supported file formats are found.
- Adds entry point and `run_modules.bruker_rename()` to rename old Bruker files with a suffix `.spm`.
- Documents usage.
- Corrects old lines in example output from TopoStats in `docs/usage.md`.
- Remove configuration to always run tests in parallel

## Removing parallel tests

I found this to be a hinderance during development as setting up workers can take longer than running small batches of tests.

Therefore removing this option the CI already uses `--numprocesses=logical` and I think we as developers can now to run `pytest -n #` when we want to run things in parallel.
---

- [x] Existing tests pass.
  - [ ] ~~`docs/configuration.md`~~
  - [x] `docs/usage.md`
  - [ ] ~~`docs/data_dictionary.md`~~
  - [ ] ~~`docs/advanced.md` and new pages it should link to.~~
- [x] Pre-commit checks pass.


SylviaWhittle (approved):
I haven't ran the code locally but I have read through all of it, it makes total sense to me.

Unsure what the parallel test stuff is but happy to defer to your judgement on that ofc.

ns-rse:
Good job I went back and looked at this as the testing of found files wasn't being done correctly (well it sort of worked but in a very convoluted way testing each element of the expected list was in the found list when the whole point of sorting them was so that we could compare the two lists directly!)

PR #1183, opened 2025-06-30
Hotfix to get old bruker files loading (file.0.002)

SylviaWhittle:
A very quick and jank hotfix to skip past the guard clause in `get_data()` that maps file extensions to the appropriate load function, that didn't get added to during #1181 

This PR is janky and needs redoing when Neil is back, but this gets it working in the meantime

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -889,6 +889,8 @@ class LoadScans:
             LOGGER.info(f"Extracting image from {self.img_path}")
             LOGGER.debug(f"File extension : {suffix}")
 
+            OLD_BRUKER_RE = re.compile(r"\.\d+$")
+
             # Check that the file extension is supported
             if suffix in suffix_to_loader:
                 data = None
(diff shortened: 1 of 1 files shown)


llwiggins:
This looks great and works well on Tom's images, happy to merge!

ns-rse:
I thought I'd fixed this with #1182 and am curious what didn't work about that?

In #1182 I modified the `io.find_files()` such that if the `file_ext: spm` (or modified on command line to be `spm`)
then both `.spm` and old files with the same regex (`re.compile(r"\.\d+$")`) should have been concatenated..

``` python
    if file_ext == ".spm":
        return list(base_dir.glob("**/*" + file_ext)) + _find_old_bruker_files(base_dir)
```

...and `io._find_old_bruker_files()`

``` python
def _find_old_bruker_files(base_dir: Path) -> list[Path]:
    r"""
    Find old Bruker files that have old extensions.

    Older Bruker files have extensions such as ``.001``, ``.002`` rather than ``.spm``. AFMReader can handle these fine
    but TopoStats needs to include them when finding files.

    Parameters
    ----------
    base_dir : Path
        Directory to search for files.

    Returns
    -------
    list[Path]
        List of files that match the regex ``\.\d+$``.
    """
    # Compile regex to match files ending with a period and one or more digits. List comprehension of matches is
    # returned.
    OLD_BRUKER_RE = re.compile(r"\.\d+$")
    return [p for p in base_dir.glob("**/*") if OLD_BRUKER_RE.match(p.suffix)]
```

It finds both `.spm` and `.002` / `.004` files that are within the test directory as the following shows..

``` python
In [1] from topostats.io import find_files

In [2]: find_files("./tests", ".spm")
Out[4]:
[PosixPath('tests/resources/minicircle.spm'),
 PosixPath('tests/resources/old_bruker.004'),
 PosixPath('tests/resources/old_bruker.002')]
```

What was the problem that needed fixing with this PR?

Issue #1171, opened 2025-06-03
Improve reporting of errors from GrainStats


ns-rse:
We are finding the broad `try:...except:` in `processing.run_grainstats()` to be frustrating when we should improve the
reporting of errors from this (and nested) `try: ... except:` statements so we can see what errors are being raised,
likely with debug.

tobyallwood:
I've added the specific exception and line number it occurs on into the logging error message, so it now reads like this:

```
[Wed, 06 Aug 2025 13:44:12] [ERROR   ] [topostats] [processing.py] [391] [minicircle] : An error occurred whilst calculating grain statistics on line 383: 'NoneType' object does not support item assignment
Returning empty dataframe.
```

The error was just an example error made by me which is removed now, but hopefully having more info including line number will help in future.

PR #1192, opened 2025-08-06
Include error message and line numbers in GrainStats error reporting

tobyallwood:
An update to issue #1171 so now errors are logged along with the line number that caused the exception. This should make it easier to locate issues within the try: section.

---

- [ ] Existing tests pass.
  - [ ] `docs/configuration.md`
  - [ ] `docs/usage.md`
- [ ] Pre-commit checks pass.

### DIFF
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -1,6 +1,6 @@
 """Functions for processing data."""
 
-import logging
+import logging, sys
 from collections import defaultdict
 from pathlib import Path
 
@@ -384,9 +384,11 @@ def run_grainstats(
             LOGGER.info(f"[{filename}] : Calculated grainstats for {len(grainstats_df)} grains.")
             LOGGER.info(f"[{filename}] : Grainstats stage completed successfully.")
             return grainstats_df, height_profiles_dict, grainstats_calculator.grain_crops
-        except Exception:
-            LOGGER.info(
-                f"[{filename}] : Errors occurred whilst calculating grain statistics. Returning empty dataframe."
+        except Exception as e:
+            _, _, e_traceback = sys.exc_info()
+            e_line_no = e_traceback.tb_lineno
+            LOGGER.error(
+                f"[{filename}] : An error occurred whilst calculating grain statistics on line {e_line_no}: {e}\nReturning empty dataframe."
             )
             return create_empty_dataframe(column_set="grainstats"), height_profiles_dict, {}
     else:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ repos:
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.12.2
+    rev: v0.12.7
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix, --show-fixes]


topostats/processing.py
```diff
@@ -384,9 +385,11 @@ def run_grainstats(
...
             LOGGER.info(f"[{filename}] : Grainstats stage completed successfully.")
             return grainstats_df, height_profiles_dict, grainstats_calculator.grain_crops
-        except Exception:
-            LOGGER.info(
-                f"[{filename}] : Errors occurred whilst calculating grain statistics. Returning empty dataframe."
+        except Exception as e:
+            _, _, e_traceback = sys.exc_info()
+            e_line_no = e_traceback.tb_lineno
+            LOGGER.error(
+                f"[{filename}] : An error occurred whilst calculating grain statistics on line {e_line_no}: {e}\nReturning empty dataframe."
```
ns-rse, line 392: The logger is setup in `topostats/logs/logs/py` and we do set `LOG_ERROR_FORMATTER` to include `%(lineno)s` so I'm not sure whether you need to import `sys` and get a traceback to include the line-number. It might just be sufficient to `LOGGER.info` > `LOGGER.error` and the line number would be reported.

ns-rse (changes requested):
Hi @tobyallwood 

Thanks for taking on this issue.

I think a more generic solution would be to set the formatting for the `LOGGER` to include the line number.

As noted we should have this already as we setup the formatting in `topostats/logs/logs.py` as noted inline.

It would be worth investigating whether this works as it should do and if not solving why as it's a more generic solution that would work wherever `LOGGER.error()` is called and avoids having to add the same functionality at all instances where we might want to have the line number repeated.

Long term I thought we had an outstanding issue to migrate to using [loguru]() but I can't seem to find it (I think @SylviaWhittle and I discussed this as its used in AFMReader) so I've just created #1194 so we don't miss it in the future. Loguru too includes the ability to include line numbers in output (by including `logger.add(..., backtrace = True)` I think) .

ns-rse:
Thanks for your on-going work on this @tobyallwood getting finer grained logging statements in `topostats/processing.py` is a good start but there is unfortunately more places where failure can occur that we would want to pin-point.

The main bulk of the processing is done in `topostats/grainstats.py` and it is the various methods/functions within here that the actual errors can occur that we want to pin point.

If you're able to pepper this with `try: ... except: ...` statements that would be wonderful. From memory one of the challenges is knowing what exceptions can occur though otherwise we fall foul of having a [broad except](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/broad-except.html). My approach here would be to see if people have files that failed Grainstats and try and work out where these are happening and what exceptions are raised, otherwise disable this linting check (for both [Pylint](https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html#message-control) and [Ruff](https://github.com/astral-sh/ruff/discussions/3442#discussioncomment-5269347)).

ns-rse:
Forgot to ask earlier, how are things going with peppering `topostats/grainstats.py` with lots of `try: ... except: ...` statements?

ns-rse:
:cool: thanks for the update @tobyallwood

ns-rse:
Configuration will be `topostats/default_config.yaml` as I don't tinker.

Log files are long gone and I've moved on from refactoring the Grains section. I will have removed the `try: ... except:` to work out where the problem was arising and fixed it as it would have been an aspect of my refactoring, but its a pain to have to remove that `try: ... except:` to dig in and is why we need finer grained error reporting.

Sorry that's not very helpful.

PR #1227, opened 2025-10-23
feature(classes): Pydantic classes and pickling

ns-rse:
Moves to [Pydantic  dataclasses](https://docs.pydantic.dev/latest/concepts/dataclasses/) for stricter data validation.

This means we can pickle `TopoStats` objects which is useful because in the test suite we don't want to run the whole
pipeline when we want to test e.g. `Nodestats`. As a consequence we now have pickles which are loaded as
`pytest.fixtures` (from `tests/conftest.py`) rather than lines of code within tests themselves that save and modify
`.npy`/`.pkl` files.

There are therefore three sets of pickles...

- `minicircle_small`
- `catenanes`
- `rep_int`

...at different stages...

- `_post_grainstats`
- `_post_disordered_tracing`
- `_post_nodestats`

...and we will develop additional fixtures for...

- `_post_ordered_tracing`
- `_post_curvature`
- `_post_splining` (optional, not required at the moment as no subsequent processing is done after this)

A slight disconnect might arise from how these pickles were created, at the moment it is code in a `.py` file on @ns-rse
computer. @ns-rse will look at adding this as an additional script in the repository, but as more work is required its
not included at the moment.

This now allows me to finish of re-factoring and writing the integration test for `ordered-tracing`.

---

Tests fail as this is part of a larger piece of work to switch to classes providing a consistent way of passing data
around, saving and loading files. See #1102 for details on progress


tests/test_processing.py
```diff
@@ -1106,671 +1158,600 @@ def test_run_disordered_tracing(
...
                         "error": False,
                         "pixel_to_nm_scaling": 1,
                         "unmatched_branch_stats": {
                             0: {"angles": np.float64(0.0)},
-                            1: {"angles": np.float64(63.62466925374735)},
-                            2: {"angles": np.float64(29.510196308405636)},
-                            3: {"angles": np.float64(106.0045892179681)},
+                            1: {"angles": np.float64(139.2976869956517)},
+                            2: {"angles": np.float64(40.75146951858621)},
+                            3: {"angles": np.float64(116.78086102349846)},
```
SylviaWhittle: I assume that the apparent change in values of this result is actually okay and not due to a change in behavior?

SylviaWhittle (approved):
This looks great, thank you.

I've just read through it, no verification on my end.

So it seems that you have simplified the testing of expected results for `test_run_nodestats` in that we now only test a few of the results rather than all? This seems fine since this isn't a unit test of features but more of a test to see if it runs relatively sensibly? 

Re: `classes.py` The sight of that many lines of boilerplate code being removed is joyous.

PR #1230, opened 2025-10-27
Add option to number grain masks in plots

tobyallwood:
This PR adds a bool in default_config.yaml to let users decide whether each grain mask should be numbered. Numbers are 0 indexxed and appear in the top left of each grain's bounding box. 

---

- [x] Existing tests pass.
  - [x] `docs/configuration.md`
- [x] Pre-commit checks pass.

### DIFF
--- a/topostats/default_config.yaml
+++ b/topostats/default_config.yaml
@@ -1,4 +1,4 @@
-base_dir: ./tests/resources # Directory in which to search for data files
+base_dir: ./ # Directory in which to search for data files
 output_dir: ./output # Directory to output results to
 log_level: info # Verbosity of output. Options: warning, error, info, debug
 cores: 2 # Number of CPU cores to utilise for processing multiple files simultaneously.


topostats/entry_point.py
```diff
@@ -625,6 +625,13 @@ def create_parser() -> arg.ArgumentParser:
         required=False,
         help="Whether to ignore warnings.",
     )
+    process_parser.add_argument(
+        "--number-masks",
```
ns-rse: This should be consistent with the parameter it modifies so either...

```suggestions
        "--number-masks",
```

...or if we got with `number_grain_plots` then...

```suggestions
        "--number-grain-plots",
        dest="number_grain_plots",
```

topostats/plottingfuncs.py
```diff
@@ -646,6 +662,45 @@ def add_bounding_boxes_to_plot(fig, ax, shape: tuple, region_properties: list, p
...
+def number_grain_plots(
+    fig, ax, shape: tuple, region_properties: list, pixel_to_nm_scaling: float, offset: tuple
+) -> tuple:
+    """
+    Add the grain numbers to the plot.
+
+    Parameters
+    ----------
+    fig : plt.figure.Figure
+        Matplotlib.pyplot figure object.
```
ns-rse: These could have markup so they are rendered differently.

```suggestion
        ``Matplotlib.pyplot`` figure object of labelled grains to be numbered.
```

Same applies for other references.

ns-rse (commented):
Sorry for not looking through this Monday I was working on another project in the morning and teaching in the afternoon but looking through it I would have suggested a number of small changes that keep things consistent and slightly more descriptive. Perhaps these tweaks could be rolled into #1234 ?

ns-rse:
Positioning things like this gets really complicated very quickly, I like the KISS principle and am happy that users re-run if something is necessary.

I do recall seeing an R package that did this sort of thing very "intelligently" (i.e. smart coder) but can't recall the name of it otherwise I would link to it should anyone want to look at methods for implementing.

PR #1233, opened 2025-10-28
fix(create_config): Corrects generalised configuration file writing

### LINKED ISSUE #1229: topostats generates empty config files
In the most recent pull from main, the commands:

```
topostats create-config

topostats create-config --filename
```

now generate empty config files that cannot be used. The most recent config file prior to my latest pull still works.

ns-rse:
Closes  #1229

- Correctly decode bytes to string when writing configuration file.
- Update parameter names for `test_reconcile_config_args_partial_config_with_overrides()`.
- Removes extra '# ' from header of some configuration files.
- Sets a the `default="default"` for the `--config` option to `create-config` sub-parser.
- Removes option `--simple` from the `create-config` sub-parser in favour of `--config simple`
- Removes the `create-matplotlibrc` sub-parser completely now handled by in favour of `--config mplstyle` which writes
  `topostats.mplstyle`.
- Removes the option `--create-label-file` from the `summary` sub-parser and instead we now create this as `topostats
  create-config --config var_to_label` as its less code and more consistent.
- Updates `docs/usage.md` to reflect these changes (it was outdated anyway).
- Removes redundant tests from `tests/test_plotting.py`
- Updates tests of `tests/test_config.py` with all combinatinos of parameters for the four supported configuration
  files. 
- Updates all necessary tests of `test/test_entry_points.py` removing one and instead parameterising another as well as
  updating `tests_config.py`.
- Adds `create-config` to the list of modules in `update_module()` function which sets `args.module` to `topostats` so
  that the correct configuration file is loaded for the given package and calls `update_module()` priort to attempting
  to load the file for writing.

The last step will need revising in the future if/when we introduce sub-sub-parsers (as proposed in the docstring for
`update_module()`) to allow the flexibility of working with different pipelines.

---

- [X] Existing tests pass.
- [ ] ~Documentation has been updated and builds. Remember to update as required...~
- [X] Pre-commit checks pass.
- [ ] ~New functions/methods have typehints and docstrings.~
- [ ] ~New functions/methods have tests which check the intended behaviour is correct.~

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -235,15 +235,33 @@ want to make to the default configuration and how to make them.
 TopoStats will use some reasonable default parameters by default, but typically you will want to customise the
 parameters that are used. This is achieved using a [configuration](configuration) file. This is a
 [YAML](https://yaml.org) file that contains parameters for different settings. For convenience you can generate
-a sample configuration file in your current working directory using the `topostats create-config` sub-command. It
-takes a single argument, the name of the file to save the configuration to (e.g. `config.yaml` or `settings.yaml`), and
-it will write the current default configuration to that file.
+a sample configuration file in your current working directory using the `topostats create-config` sub-command. There are
+a number of arguments available. They key one is `--config` which defines what configuration file is written and
+currently has three possible options `default`, `simple`, `mplstyle` or `var_to_label`. See `topostats create-config ---help`
+for descriptions of other available options. **NB** if `--filename` is omitted then the value of `--config` determines
+the filename the configuration is written to as shown below.
 
 ```bash
-topostats create-config --filename my_config.yaml
-ls -l
-my_config.yaml
-sample_image_scan_2022-12-08-1204.spm
+topostats create-config --config default --filename my_default_config.yaml
+[Tue, 28 Oct 2025 16:12:16] [INFO    ] [topostats] TopoStats version : 2.3.2.dev600
+[Tue, 28 Oct 2025 16:12:16] [INFO    ] [topostats] Commit            : 54e4b940a
+[Tue, 28 Oct 2025 16:12:16] [INFO    ] [topostats] A sample configuration has been written to : my_default_config.yaml
+topostats create-config --config default
+[Tue, 28 Oct 2025 16:12:18] [INFO    ] [topostats] TopoStats version : 2.3.2.dev600
+[Tue, 28 Oct 2025 16:12:18] [INFO    ] [topostats] Commit            : 54e4b940a
+[Tue, 28 Oct 2025 16:12:18] [INFO    ] [topostats] A sample configuration has been written to : default_config.yaml
+❱ topostats create-config --config simple
+[Tue, 28 Oct 2025 16:12:22] [INFO    ] [topostats] TopoStats version : 2.3.2.dev600
+[Tue, 28 Oct 2025 16:12:22] [INFO    ] [topostats] Commit            : 54e4b940a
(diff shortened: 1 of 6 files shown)


ns-rse:
Thanks @SylviaWhittle this PR set `args.config="default"` so it should be fixed.

Personally I'd rather not have a `default` / `simple` option for configurations as it adds another layer of complexity and increases maintenance overhead when things need changing (as I've discovered!) but you wrote code that works so I should maintain that state.

Whilst investigating why tests were failing I've noticed also that when we write `var_to_label.yaml` it is no longer
valid YAML (its a dictionary). That appears to have been the case for a while so the tests were poor and it needs fixing
too.

ns-rse:
> So args.simple is never used and args.config is always "default"

I remember now/found what it is and the reminder was in the updated help for `--config`...

In #1221 I was trying to move to a more flexible way of specifying what configuration is required. We already have the
`--config` option which had...

``` python
        help="Configuration to use, currently only one is supported, the 'default'.",
```

...and we had the `--simple` option which was boolean. As I want to be able to add other types of configuration from
other packages I opted to condense this so that `--config` holds the configuration that is required, which is kind of as
intended (based on above original `help`) and we should have had it take `default` or `simple` as an argument in the
first place rather than adding _another_ option...

``` python
        help="Configuration to use, currently 'default' and 'simple' are supported.",
```

Thus I've removed `--simple` in my current work (soon to be added with `git commit --amend` to keep history clean).

I can now run the following and we get a simple and full (aka default) configuration file...

``` shell
topostats create-config --config simple --filename simple.yaml
topostats create-config --config default --filename default.yaml
wc -l simple.yaml default.yaml
   57 simple.yaml
  140 default.yaml
  197 total
```

What I've also discovered though is that we can reduce the amount of code further!

``` shell
topostats create-config --config topostats.mplstyle
[Tue, 28 Oct 2025 15:55:36] [INFO    ] [topostats] TopoStats version : 2.3.2.dev600
[Tue, 28 Oct 2025 15:55:36] [INFO    ] [topostats] Commit            : 54e4b940a
[Tue, 28 Oct 2025 15:55:36] [INFO    ] [topostats] A sample configuration has been written to : topostats.mplstyle
```

and this is what we want (i.e. a copy of `topostats/topostats.mplstyle`). We can therefore remove from `entry_point.py`
the `create_matplotlibrc_parser` sub-parser! Things should be setup to use the `write_config_with_comments()` function
in other packages since it uses `args.module` to determine what package to use (via `get_data()` function, note that for
now `args.module` is updated to `topostats` if running `process` / `filter` / `grains` but I've left an indication of
intent in the docstring of `update_module` for introducing sub-sub parsers as we look to expand `topostats`
functionality with other pipelines/modules.

Just got to work out why `var_to_label.yaml` comes out as a dictionary instead of valid YAML!

TFirth2:
I just used

`git checkout ns-rse/1229-fix-create-config`

to switch to the branch where the fix is and then

`pip install -e .`

to install the branch.

I ran both

```
topostats create-config

topostats create-config --filename test_config.yaml
```

and both made empty config files with 3 commented rows at the top and 'none' on the fourth row. 

Did I make a mistake somewhere?

ns-rse:
@TFirth2 and I had a quick video chat and resolved the issue. `git pull` was required to update the
`ns-rse/1229-fix-create-config` locally. Once that was done all was good in the world and configurations could be
created again.

This PR is ready for review.

SylviaWhittle:
Confirmed, features seem to work locally

Issue #1089, opened 2025-02-13
[Bug]: disordered tracing : mask smoothing sometimes lets mask touch edge of mask if padding is small


SylviaWhittle:
### Checklist

- [x] Find the offending file in the output. If processing halts, re-run analysis with `topostats --core 1 process`.
- [x] Describe the bug.
- [x] Include the configuration file.
- [x] Copy of the log-file from running with `topostats --log-level debug <command>`.
- [x] The exact command that failed. This is what you typed at the command line, including any options.
- [x] TopoStats version, this is reported by `topostats --version`
- [x] Operating System and Python Version

### Describe the bug

When padding is too small, the mask smoothing at the start of disordered tracing `trace_dna()` causes the mask to touch the border of the image.

Original mask:

![Image](https://github.com/user-attachments/assets/53650344-e9a5-4d1f-8bfe-08c0bf36a92d)

Smoothed mask:

![Image](https://github.com/user-attachments/assets/ba062f31-2514-4d08-ace3-e5a19d3e5197)

This issue is solvable by simply increasing the `grain_crop_padding` to, say, 10.

I don't think we should spend time trying to prevent this occurring, rather I think we should make it clear in the logs when this error occurs that it can be solved by increasing the padding.

Here is the output currently when the error is thrown:

```
[Thu, 13 Feb 2025 09:58:54] [ERROR   ] [topostats] [20221019_Ni_Topo_20ng_Shelterin_40nM_5min.0_00060] : Disordered tracing of grain 1 failed. Consider raising an issue on GitHub. Error: 
Traceback (most recent call last):
  File "/Users/laura/TopoStats/topostats/tracing/disordered_tracing.py", line 367, in trace_image_disordered
    disordered_trace_images: dict | None = disordered_trace_grain(
                                           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/laura/TopoStats/topostats/tracing/disordered_tracing.py", line 654, in disordered_trace_grain
    disorderedtrace.trace_dna()
  File "/Users/laura/TopoStats/topostats/tracing/disordered_tracing.py", line 127, in trace_dna
    ).get_skeleton()
      ^^^^^^^^^^^^^^
  File "/Users/laura/TopoStats/topostats/tracing/skeletonize.py", line 70, in get_skeleton
    return self._get_skeletonize()
           ^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1242, opened 2025-11-05
Warn user of missing padding after smoothing masks

tobyallwood:
Partially addresses #1089 - waiting for confirmation this is the route we want to take to solve it.

After smoothing GrainCrop masks, a check will be performed to ensure all masks still have padding of at least 1 on all edges of the image as occasionally smoothing will remove this.
If padding has been accidentally removed, the grain is skipped and a warning is logged informing the user of the issue and prompting them to re-run the process with a higher padding value in the config file (`grains.grain_crop_padding`).
If all grains have the padding issue (therefore no grains are successfully processed) an empty dataframe will be returned to `process_scan()` and subsequent analysis steps will be skipped.

### Please someone check:
- The masks that have been added to tests/resources (as .npz) are between 100 and 200 pixels in each direction. I feel like smaller files would be best but I could not find any examples other than these as it's quite a rare bug. Would be nice to have confirmation that this is still ok though.

---

- [x] Existing tests pass.
- [x] Pre-commit checks pass.


tests/tracing/test_disordered_tracing.py
```diff
@@ -1287,6 +1287,24 @@ def test_disordered_trace_grain(
...
+        pytest.param("test_check_missing_padding_missing.npz", True, id="padding missing"),
+        pytest.param("test_check_missing_padding_not_missing.npz", False, id="padding exists"),
+    ],
+)
+def test_check_missing_padding(
+    mask_filename: npt.NDArray,
+    expected_return: bool,
+) -> None:
+    """Test the check_missing_padding() method."""
+    data = np.load(GENERAL_RESOURCES / mask_filename)
```
ns-rse: Some more comments on reflection.

Its generally better to define these things as `pytest.fixture`'s within `conftest.py` as they can then be re-used across tests.  You would then need to use the `request` fixture as a parameter to the test and load it with `request.getfixturevalue()`.

But this is slightly redundant in light of my suggestion for simplifying the test case.

topostats/tracing/disordered_tracing.py
```diff
@@ -296,6 +303,32 @@ def calculate_dna_width(
...
+        smoothed_mask : npt.NDArray
+            Numpy array of mask of graincrop after smoothing.
+
+        Returns
+        -------
+        bool
+            True or False if padding is missing.
+        """
+        rows, cols = smoothed_mask.shape
+        for r in range(rows):
```
ns-rse: This loops over all rows and all columns, we don't need to do that though and on large arrays results in a lot of unnecessary calculations.

The arrays are boolean here and so we can instead take a slice which is the first row and sum it. If the sum of that row is > 0 then there is a pixel touching the edge. Same is true of first column, last row and last column.

tobyallwood: Ok got it, that sounds good. Will get this changed now

ns-rse: And here are some timings on larger arrays...

```python
timer_missing_padding = timeit.Timer(lambda: check_missing_padding(test_arrays["big_top"]))
timer_pixels_touching = timeit.Timer(lambda: check_pixels_touching_border(test_arrays["big_top"]))

def test_methods(timer: timeit.Timer, n_timeit = 100000, n_trials=30, type="Test"):
    results = [timer.timeit(number=n_timeit) for _ in range(n_trials)]
    print(f"Results from {n_trials} trials using {type}:")
    print(f"Mean (std) execution time : {statistics.mean(results)} ({statistics.stdev(results)})")
    print(f"Min execution time : {min(results)}")
    print(f"Max execution time : {min(results)}")
```

The results of these are...

## Original Method running across all rows and columns

```
test_methods(timer_missing_padding, type="All rows and cols")
Results from 30 trials using All rows and cols:
Mean (std) execution time : 6.222153178867302 (0.14937840294788263)
Min execution time : 5.982440954016056
Max execution time : 6.693786427000305
```

## Alternative method just checking borders

```
test_methods(timer_pixels_touching, type="Just borders")
Results from 30 trials using Just borders:
Mean (std) execution time : 3.7323081669339446 (0.19466266479402933)
Min execution time : 3.544059545994969
Max execution time : 4.371990004001418
```

Just checking the borders is, perhaps obviously, faster by about 40%.

## Final thought

Also by checking the sum of the rows and columns is zero, which means the image is correctly padded and there are no pixels touching the edge, of the image it means the function is generalisable and works with arrays of heights as well as with boolean masks. 

We can write tests to check that too (not included!). :magic_wand:

tests/tracing/test_disordered_tracing.py
```diff
@@ -1287,6 +1287,24 @@ def test_disordered_trace_grain(
...
+    ("mask_filename", "expected_return"),
+    [
+        pytest.param("test_check_missing_padding_missing.npz", True, id="padding missing"),
+        pytest.param("test_check_missing_padding_not_missing.npz", False, id="padding exists"),
+    ],
+)
+def test_check_missing_padding(
+    mask_filename: npt.NDArray,
+    expected_return: bool,
+) -> None:
```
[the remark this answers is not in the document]
ns-rse: That means the test of `trace_dna()` needs testing to see if the warning is raised, more an integration test than a unit test.

topostats/tracing/disordered_tracing.py
tobyallwood: That makes sense. If we're having this function that generalisable does it make sense to move the function out of the disorderedTracing class and into something more general too? Like utils.py maybe?

topostats/tracing/disordered_tracing.py
```diff
@@ -304,30 +304,21 @@ def calculate_dna_width(
...
-        for r in range(rows):
-            for c in range(cols):
-                if smoothed_mask[r, c]:
-                    neighbors = smoothed_mask[r - 1 : r + 2, c - 1 : c + 2]
-                    if neighbors.shape != (3, 3):
-                        # Grain pixel found on edge of image
-                        return True
-        # No grain pixels are touching the edge
-        return False
+        return mask[:, 0].any() or mask[:, -1].any() or mask[0, :].any() or mask[-1, :].any()
```
ns-rse: Checked the timings on this too...

```python
def check_any(mask: npt.NDArray) -> bool:
    return mask[:, 0].any() or mask[:, -1].any() or mask[0, :].any() or mask[-1, :].any()

timer_any = timeit.Timer(lambda: check_any(test_arrays["big_top"]))

test_methods(timer_any, type="Just borders")

Results from 30 trials using Just borders:
Mean (std) execution time : 8.439903999267456 (0.23607712830966773)
Min execution time : 8.10486877997755
Max execution time : 8.971229262009729
```

This turns out to be slower than your original function and my suggestion!

I suspect as it has to do four calculations every single time.

Your original breaks out early if it finds a pixel on the border.

My suggestion is slightly quicker as its summing a single array and breaks out early the minute one of the edges match.

I could do further checks on the speed of each border but have to get on with refactoring.

ns-rse (approved):

PR #1255, opened 2025-11-13
Added __str__() methods to classes in classes.py

tobyallwood:
A `__str__()` method has been added to all classes in `classes.py`, along with test functions for each. This will allow data and attributes from classes to be displayed in a user-friendly way when the class name is typed into the terminal. For more complex data types general information about the attribute is shown rather than all the data (e.g. arrays printing their shape)

`spline_coords` has also been renamed to `splined_coords` in a few places to ensure consistency 

---

- [x] Pre-commit checks pass.


ns-rse (changes requested):
Thanks @tobyallwood one minor thought on using `str()` v `print()` that we could expand the tests to cover.

PR #1265, opened 2025-12-02
Created functions to convert classes to dictionaries

tobyallwood:
This PR appears very large at first, but > 3000 of the added lines are from the new snapshot file used in testing.

In this PR the main changes are two new functions in `classes.py`:

- `convert_to_dict()` - takes a class instance and converts all attributes to a dictionary
- `prepare_data_for_df()` - using a `TopoStats` class instance and a string as input parameters, this function creates a dictionary of the attributes selected in the mapping (chosen by another mapping of the input string and a dictionary) from multiple classes which consists of the data required for a CSV output. There are four possible configurations that will be used for each of the four CSVs that TopoStats produces (image_statistics.csv, branch_statistics.csv, molecule_statistics.csv, grain_statistics.csv).

The classes have also been expanded where required to hold the necessary data, this includes their dummy fixtures in `conftest.py`. Setter and getter functions have not yet been set up for these new attributes.

Due to my own time constraints more work is required on this for it to be adequately pythonic and functional, for now it's at a point where the function successfully creates the required dictionaries given the correct mapping, but is not called by anywhere in the TopoStats process (except for some example code I have written in `run_modules.py` so you can see my vision on this function's implementation).
On top of this, the old method of passing `DataFrame` objects around still exists and is still used in the process, and new code populating the classes with required values (which `prepare_data_for_df()` accesses) are next to it. The idea is that eventually all code passing `DataFrames` can be removed and everything will just save the data to the classes as attributes.

`prepare_data_for_df()` will only be called in `run_modules.py process()` when creating CSVs, and will be used in place of the `DataFrames` that're given e.g:
```
results = pd.DataFrame(prepare_data_for_df(topostats_object, "grain_statistics"))
results.to_csv(config["output_dir"] / "grain_statistics.csv", index=True)
```
A rough example of how the function would be implemented is in `run_modules.py` in `process()`. After finishing them, checking that these examples properly work and removing the old method the function should get a decent amount smaller.

`prepare_data_for_df()` is an unfortunately large and 'clunky' function that will need simplifying at some point and causes a number of linting issues due to its size and complexity. I hope however that it is a suitable start point for the final implementation of this idea.

The tests for these two new functions pass without issue, however pre-commit checks fail due to the unfinished nature of both my changes and the branch I originally forked from.

---


topostats/classes.py
```diff
@@ -1125,30 +1235,214 @@ def __str__(self) -> str:
...
     Returns
     -------
-    dict[str, Any]
-        Dictionary representation of the object.
+    list[dict[str, Any]]
+        A list of dictionaries containing required data, ready to be converted into a pd.DataFrame.
     """
-    return dict(something.__dict__)
+    mappings = {
+        "image_statistics": {
```
ns-rse: Lets drop the `_statistics` from each of these, less to be mis-spelt.

```suggestion
        "image": {
```

topostats/classes.py
```diff
@@ -46,6 +48,10 @@ class GrainCrop:
         Thresholds used to find the grain.
     filename : str
         Filename of the image from which the crop was taken.
+    threshold : str
```
ns-rse: I realise this is required to fit into the structure of the current output but I think we need to think carefully about `threshold` since we have removed the notion of "direction" by removing the `ImageGrainCrop` and `GrainCropDirection`.

Configuration allows a number of thresholds to be set, both could be positive, so what does being "`above`" a threshold actually mean?

ns-rse (commented):
Thanks for your work on this @tobyallwood gives me a good starting point.

I've left notes in-line for my reference as I went through this but have merged this locally as I find it easier to work through the conflicts that arise.

One thing that I did clock was that the point at which data should be extracted was wrong. It looked as though you were expecting to have `topostats_object: TopoStats` returned by the `process_scan` function and _then_ extract the statistics. This isn't how the workflow is setup as `process_scan()` (and other `process_*` functions) are meant to return the dataframes because they are used in the `partial()` function so that images can be processed in parrallel. If we have them return the required dataframes then we don't need to modify anything after the parallel processing has completed.

PR #1278, opened 2026-01-05
Hotfix - Comments out `save_folder_grainstats` until resolved

Max-Gamill:
# TopoStats Pull Requests

Hey everyone! Hope you're all doing well!

While I was trying to get my persistence length code for Simons work into TopoStats as it's just kind of hanging around my laptop but, I was greeted with an error trying to run `main` which complained there was no "basename" column in the dataframe. I took a look at the tests and this column is dropped from the regtests with the comment "Remove the basename column as this differs on CI".

So I thought I'd make a quick hotfix which just comments this function out allowing TS to run to completion.  I'm not sure if this is going to be addressed later so I didn't want to delete anything and do extra cleaning as in my mind the folder stats could still be useful but I'm unsure of the plan.

Anyway, I hope this helps :))

P.S. If there's a linear DNA image someone could send me to verify the PL measurements I'd be super greatfull!

---

- [x] Existing tests pass.
  - [ ] `docs/configuration.md`
  - [ ] `docs/usage.md`
- [x] Pre-commit checks pass.


ns-rse:
Hi @Max-Gamill,

Hope you are getting on ok in London.

I'm not sure we should be commenting out the calls to `save_folder_grainstats()` for everyone as they may well be using the resulting file and I would rather solve the problem than comment something important out.

Some questions...

- What does the configuration file look like?
- How many files are being processed within the `base_dir`?
- Is there a log/traceback for the error please?

With this information we can document the problem in an issue and start investigating a solution rather than commenting out code which could then creep into a release which would be unwanted.

Max-Gamill:
Hey Neil, nice to hear from you!

Going well down here in the big smoke, thank you.

I agree a permanent solution is ideal but I thought a little temp fix would help as it looked like it was midway through removal with the tests and prevent new users from running the example files.

Anyway, a nice little update from this is that upon pulling the most recent `main` it seems fixed, not sure how though as not much has been merged since I cloned it only the other day.

Thanks for your rapid response, closing with comment :)

PR #1284, opened 2026-01-23
Introduce dataclasses and remodel workflows to use these

### LINKED ISSUE #1112: `ImageGrainCrops` - `GrainStats`
Refactor `Grains` class and processor functions to work with `ImageGrainCrops`.

Has been refactored to take these as input but the `calculate_stats()` method returns a dictionary.

I think we want to return this dictionary for ease of combining with other data and for now to save having to write additional code that pulls all the data out from the new classes but with an eye to that in the future we could add the statistics to each `GrainCrop` that is nested within the `ImageGrainCrops` (may not need to explicitly return the updated objects, I can't remember at the moment!)

### LINKED ISSUE #1113: `ImageGrainCrops` - `disorderedTrace`
Refactor `disorderedTrace` class and processor functions to work with `ImageGrainCrops`.

I think we want to add all the arrays that are returned as a dictionary to be components of the `GrainCrop` class for each grain and for now also return the dictionary to retain functionality with subsequent steps. Once subsequent steps have been adapted to work with `ImageGrainCrops` (which contains nested `GrainCrop` objects/classes) we can then remove returning the dictionary.

### LINKED ISSUE #1114: `ImageGrainCrops` - `nodestats
Refactor `disorderedTrace` class and processor functions to work with `ImageGrainCrops`.

Add the `NodeDict` / `MatchedBranch` / `ImageDict` classes that are currently returned by `nodeStats.get_node_stats()` to the `GrainCrop` class for each grain.

As with other steps of the refactoring for now also return the dictionary to retain functionality with subsequent steps. Once subsequent steps have been adapted to work with `ImageGrainCrops` (which contains nested `GrainCrop` objects/classes) we can then remove returning the dictionary.

### LINKED ISSUE #1115: `ImageGrainCrops` - `OrderedTraceNodestats`
Refactor `OrderedTraceNodestats` and `OrderedTraceNodestats`  class and processor functions to work with `ImageGrainCrops`.

There both of which...

- `ordered_trace_data` - a dictionary (indexed by `mol_{i}`) of `ordered_coords`, `heights`, `distances` and `mol_stats`
- `grain_mol_tracing_stats` - a dictionary (indexed by `mol_{i}`) of `circular`, `topology` and `topology_flip`
- `self.grain_tracing_stats` - a dictionary of `num_mols`, `writhe_string`
- `self.images` - a dictionary of `trace_segments`, `over_under`, `all_molecules` and `ordered_traces`.

Also updates the `self.nodestats_dict()` (see lines 674-675)

As with other steps of the refactoring for now also return the dictionary to retain functionality with subsequent steps. Once subsequent steps have been adapted to work with `ImageGrainCrops` (which contains nested `GrainCrop` objects/classes) we can then remove returning the dictionary.

### LINKED ISSUE #1116: `ImageGrainStats` - `splineTrace`
Refactor `splineTrace` and `windowTrace` classes and processor functions to work with `ImageGrainCrops`.

There bothreturn...

- `splined_trace` - a Numpy array.
- `self.tracing_stats` - a dictionary of `contour_length` and `end_to_end_distance`.

As with other steps of the refactoring for now also return the dictionary to retain functionality with subsequent
steps. Once subsequent steps have been adapted to work with `ImageGrainCrops` (which contains nested `GrainCrop`
objects/classes) we can then remove returning the dictionary.

### LINKED ISSUE #1117: `ImageGrainCrops` - `Curvature`
The `measure.curvature.calculate_curvature_stats_image()` _function_ is called from `processing.run_curvature_stats()` and takes in the `grain_trace_data[direction]` which is a dictionary with `molecule_data` being one element and is itself a dictionary of `spline_coords` `tracing_stats` (returned by the ordered tracing step).

The `grain_curvature_stats` are returned.

As with other steps of the refactoring for now also return the dictionary to retain functionality with subsequent steps. Once subsequent steps have been adapted to work with `ImageGrainCrops` (which contains nested `GrainCrop` objects/classes) we can then remove returning the dictionary.

### LINKED ISSUE #1134: Move Grain Classes to their own objects
We've found that the `topostats.grains` module and the associated tests are getting rather large.

To this end after #1123 has been merged we propose splitting code (and in turn the tests) into smaller modules.

| Current                                | Proposed                                    |
| `topostats.grains.GrainCrop`           | `topostats.dataclasses.GrainCrop`           |
| `topostats.grains.GrainCropsDirection` | `topostats.dataclasses.GrainCropsDirection` |
| `topostats.grains.ImageGrainCrops`     | `topostats.dataclasses.ImageGrainCrops`     |
| `topostats.grains.Grains`              | `topostats.grains.Grains`                   |

Further the `topostats.grains.Grains.vet_grains()` static method could also be moved out to its own sub-module rather
than being a method of the class.

### LINKED ISSUE #1140: Update HDF5 field `topostats_file_version` > `topostats_version`
As we move to recording the version of TopoStats that a `.topostats` file was created with rather than a
`topostats_file_version` we should record this as `topostats_version` in the HDF5 itself.

This will require changes in `TopoStats` and `AFMReader` but should be backwards compatible so we should use `try: ...`
to maintain functioning with existing files that use `0.2` as the `topostats_file_version` field.

### LINKED ISSUE #1152: Switch all regression tests to syrupy

### LINKED ISSUE #1256: [feature] : Consider using compression when saving to .topostats file

Currently we don't use compression when saving to .topostats files, resulting in huge file system weight. My own laptop is 90% full, mostly from `.spm` and `.topostats` files.

At a cost of I/O speed, we could enable compression of our HDF5 datasets when saving to `.topostats` files.

Here's some documentation on compression when saving to `.HDF5` files using `h5py`.
https://docs.h5py.org/en/latest/high/dataset.html

I haven't looked into this further but I did do a little investigation a while back and saw that compressed `.HDF5` files were impressively small.

Note that `array_size` here is the size of one side of the square, so `array_size = 100` means a `100 x 100` image.

Also note that the load time differences for anything below a `2000 x 2000` pixel image are negligible in my test case.

Image

Image

_No response_

### Additional context

_No response_

### LINKED ISSUE #1279: [feature] : Remove "grain junctions" from grain_statistics.csv

"grain_junctions" is calculated as the number of pixels that are designated as junctions (number of pixels that have > 2 neighbours).

Users are mistakenly thinking this means number of DNA crossings. This is bad and not correct. Users don't need to know the number of pixels that are designated as junctions, I don't see a use for it.

Remove it from the `grain_statistics.csv` file.

_No response_

### Additional context

_No response_

### LINKED ISSUE #1282: Whole image curvature plots are slightly off
In the final throes of #1102 I'm looking at the output for many things and noticed that the curvature plots for whole images are every so slightly mis-aligned by one pixel.

Taking `minicircle_small.topostats` from the test suite we have...

## Splining/Smoothing plot

Image

## Curvature plot

Image

The curvature traces are shifter to the left by one pixel, which happens to be the same value as `grain_crop.padding`

## Individual grains

| Grain 0 | Grain 1 | Grain 2 |
|---------|---------|---------|
| Image | Image | Image |

Its not an issue with the location of the curvature plots in and of themselves its more to do with how they are placed on the whole image plot.

##  `plottingfuncs.Images.plot_curvatures()`

The lines where the plots are overlaid to the whole image reside within this function on lines [387-404](https://github.com/AFM-SPM/TopoStats/blob/725cf662bcac9076b92d396b87ae303802438c90/topostats/plottingfuncs.py#L387)...

``` python
                    cmap = mpl.cm.coolwarm
                    for index, point in enumerate(molecule_trace_coords):
                        color = cmap(normalised_curvature[index])
                        if index > 0:
                            previous_point = molecule_trace_coords[index - 1]
                            ax.plot(
                                [
                                    (min_col - pad_width + previous_point[1]) * self.pixel_to_nm_scaling,
                                    (min_col - pad_width + point[1]) * self.pixel_to_nm_scaling,
                                ],
                                [
                                    (image.shape[0] - (min_row - pad_width + previous_point[0]))
                                    * self.pixel_to_nm_scaling,
                                    (image.shape[0] - (min_row - pad_width + point[0])) * self.pixel_to_nm_scaling,
                                ],
                                color=color,
                                linewidth=1,
                            )
```

**NB** `pad_width` comes from `grain_crop.padding` attribute (see previous lines)

If we set the `pad_width` to be `3` _just within this function_ then this shifts the curvature trace further left and upwards

Image

To get the trace in the correct place we therefore need to _not_ subtract the `pad_width` from `min_col`...

```python
                                [
                                    (min_col + previous_point[1]) * self.pixel_to_nm_scaling,
                                    (min_col + point[1]) * self.pixel_to_nm_scaling,
                                ],
                                [
                                    (image.shape[0] - (min_row + previous_point[0]))
                                    * self.pixel_to_nm_scaling,
                                    (image.shape[0] - (min_row + point[0])) * self.pixel_to_nm_scaling,
                                ],
```

...which aligns things, note that the `min

Image

## Other consideration(s)

I think the reason for this is the introduction of `array_manipulation.pad_bounding_box_dynamically_at_limits()` which is called from `array_manipulation.re_crop_grain_image_and_mask_to_set_size_nm()` which itself is called `processin.run_grains()` (i.e. very early on in the processing pipeline).

## Solution

I (@ns-rse) have already fixed this and will be in including this in my forthcoming Pull Request but wanted to explicitly document the problem and change.

ns-rse:
**NB** - I expect there to be questions and problems with this PR, it needs and would benefit from wide testing on lots
of different samples.

Building on the introduction of dataclasses that @SylviaWhittle begun for `GrainCrops` this pull request rounds that out
and implements dataclasses throughout. The top-level `TopoStats` dataclass holds meta-data and image information and
there are classes for...

- `GrainCrops`
- `DisorderedTrace`
- `Node`
- `OrderedTrace`
- `MatchedBranch`
- `Molecule`

We use [Pydantic](https://docs.pydantic.dev/latest/) dataclasses which offer data validation on instantiation of classes
and because of the options that have been set when attempting to modify/update values validation is also used. This
ensures the types of data are set/used in each of the classes is always correct and helps minimise edge case problems
when developing in the future.

## Direction

After discussion we have removed the concept of processing for `above` and `below` and in turn removed the
`ImageGrainCrops` and `GrainCropsDirection` classes. Both `above` and `below` can now have multiple thresholds (so that
different types of molecules can be detected) and so it was felt that the concept/distinction was arbitrary. Thresholds
are available in both of these traditional directions and can have one or more values but we just don't nest the results
by direction. Instead `GrainCrop.threshold` is an attribute which defines the threshold used to detect that particular
grain.

## Output

As a consequence some of the output files have changed (as of writing I may have one more of some output fields to
do). Some of the more detailed, grain level plots have moved location to all be in nested directories reflecting the
processing stage they have come from.

## Development Implications

As a consequence we no longer pass around dataframes that need things adding to them. Instead all data is stored as
attributes to the different levels of classes. At the end of processing we build dictionaries of the attributes. These
are aggregated into nested dictionaries and are converted to Pandas Dataframes for writing to CSV.

### Nomenclature

A big challenge in the refactoring stemmed from ambiguous nomenclature, "skeletons" are used in many stages of tracing
but there are _lots_ of different types of skeleton. This has hopefully been clarified so that it is clear which
skeleton is used at which stage. Further abbrv. where encoutered have been renamed to be more descriptive. They might
have made sense when writing but lack meaning to those new to the code (including future selves!).

## Configuration

Where configuration options have changed the corresponding validation has been updated.

## Tests

Many tests required updating as a consequence and I think (hope!) I have managed to preserve consistency.

## Regression tests

In doing this we have switched from [`pytest-regtest`](https://pytest-regtest.readthedocs.io/en/latest/) to
[`syrupy`](https://syrupy-project.github.io/syrupy/) for regression testing of output.

I have also attempted to remove a large number of `.pkl`'s (Python pickles which are seralised Python objects saved to
disk) which were used to compare test results against. Some still remain though. The solution to this is noted in
`tests/conftest.py` but broadly it is to keep fully processed `.topostats` files for test images, load these as fixtures
and then set the attributes that are to be tested to `None` for the tests that perform the calculations. I would have
liked to have done this myself but have run out of time.

## Closes

This pull request closes multiple issues. I think it covers the following at the very least but may also close others
not listed below

Closes #1112
Closes #1113
Closes #1114
Closes #1115
Closes #1116
Closes #1117
Closes #1134
Closes #1140
Closes #1152
Closes #1256
Closes #1279
Closes #1282

## ToDo

There are still entry points that are required to be written (see #517) but now that we have a consistency between the internal
representation of `TopoStats` objects and how these are stored in `.topostats` HDF5 files (and in turn loaded) these
should be relatively straight-forward. See the existing examples for `filters` / `grains` / `grainstats` for examples of
how to do this (adding sub-parsers with relevant options and a function in `topostats.run_modules.py` that uses
`partial()` to run a function defined in `topostats.processing.py` which loads `.topostats` files and runs the module
using the relevant `topostats.processing.run_<module>()`).

---

- [x] **Existing tests pass** - but have been dramatically overhauled to reflect the changes.
- [x] **Documentation has been updated and builds. Remember to update as required...**
  - [ ] ~`docs/configuration.md`~
  - [ ] ~`docs/usage.md`~
  - [ ] ~`docs/data_dictionary.md`~
  - [x] **`docs/advanced.md` and new pages it should link to.** Includes `docs/advanced/classes.py` which seeks to explain
        how thte dataclasses are structured.
- [x] Pre-commit checks pass.
- [x] **New functions/methods have typehints and docstrings.** - Have improved typehints in many places and ensured they
      are used in new functions.
- [x] **New functions/methods have tests which check the intended behaviour is correct.** - New functions have tests included.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,6 +1,6 @@
 name: Bug report
 description: Create a report to help us improve
-title: "[Bug]: "
+title: "[bug]: "
 labels: "bug"
 assignees:
   - ""


topostats/tracing/nodestats.py
```diff
@@ -189,53 +131,34 @@ def __init__(
...
-        tuple[dict, dict]
-            Dictionaries of the node_information and images.
+        dict[int, npt.NDArray]
+            Dictionary of images.
         """
         LOGGER.debug(f"Node Stats - Processing Grain: {self.n_grain}")
-        self.conv_skelly = convolve_skeleton(self.skeleton)
-        if len(self.conv_skelly[self.conv_skelly == 3]) != 0:  # check if any nodes
+        self.grain_crop.convolved_skeleton = convolve_skeleton(self.skeleton)
+        if self.grain_crop.convolved_skeleton.max() == 3:  # check if any nodes
```
tobyallwood:
```suggestion
        if self.grain_crop.convolved_skeleton.max() == 3:  # check if any nodes are crossings
```

topostats/classes.py
```diff
@@ -0,0 +1,1220 @@
...
+    branch_stats : dict[int, MatchedBranch]
+        Dictionary of ``MatchedBranch`` objects.
+    unmatched_branch_stats : dict[int, UnMatchedBranch]
+        Dictionary of ``UnmatchedBranch`` objects.
+    node_coords : npt.NDArray[np.int32]
+        Numpy array of node coordinates.
+    confidence : np.float64
+        Confidence in ???.
+    reduced_node_area : ???
+        Reduced node area.
```
SylviaWhittle:
```suggestion
        The molecule skeleton, with all branches removed that are not connected to the current node.
```

topostats/classes.py
```diff
@@ -0,0 +1,1220 @@
...
+    pixel_to_nm_scaling : np.float64
+        Pixel to nanometre scaling.
+    branch_stats : dict[int, MatchedBranch]
+        Dictionary of ``MatchedBranch`` objects.
+    unmatched_branch_stats : dict[int, UnMatchedBranch]
+        Dictionary of ``UnmatchedBranch`` objects.
+    node_coords : npt.NDArray[np.int32]
+        Numpy array of node coordinates.
+    confidence : np.float64
+        Confidence in ???.
```
SylviaWhittle:
```suggestion
        Normalized confidence of the crossing, 0-1.
```

topostats/classes.py
```diff
@@ -0,0 +1,1220 @@
...
+    grain_molstats : Any | None
+        Grain molecule statistics.
+    molecules : int
+        Number of molecules within the grain.
+    writhe : str
+        The writhe sign, can be either `+`, `-` or `0` for positive, negative or no writhe.
+    pixel_to_nm_scaling: np.float64 | None
+        Pixel to nm scaling.
+    images: dict[str, npt.NDArray] | None
+        Images of ???
```
SylviaWhittle:
```suggestion
        Diagnostic images produced during processing.
```

topostats/classes.py
```diff
@@ -0,0 +1,1220 @@
...
+    threshold : str
+        Direction from threshold of molecule (above / below)
+    molecule_number : int
+        Index of the molecule (per grain)
+    circular : str, bool, optional
+        Whether the molecule is circular or linear.
+    processing : str
+        Which processing type was used, topostats or nodestats.
+    topology : str, optional
+        Unknown?
```
SylviaWhittle:
```suggestion
        Topological classification of the molecule.
```

topostats/classes.py
```diff
@@ -0,0 +1,1220 @@
...
+    config : dict[str, Any] | None
+        Configuration used when processing the grain.
+    full_image_plots : dict[str, npt.NDArray[np.float64]]
+        Dictionary of numpy arrays where disordered tracing, nodestats and ordered tracing skeletons are mapped back to
+        the original image.
+    basename : str
+        Basename of image locations.
+    """
+
+    grain_crops: dict[int, GrainCrop] | None = None
```
SylviaWhittle: [LOW PRIORITY / FIX IN LATER PR]
TopoStats objects' properties are able to be None when, during processing, they shouldn't be / can't be. Would there be a way to ensure that their type is no longer split with None after initialization / during processing?

This is absolutely not a priority right now though!

topostats/processing.py
```diff
@@ -73,80 +64,98 @@ def run_filters(
...
+                        if plot_name not in ["scan_raw"]:
+                            if plot_name == "extracted_channel":
+                                array = np.flipud(array.pixels)
+                            plotting_config["plot_dict"][plot_name]["output_dir"] = (
+                                core_out_path
+                                if plotting_config["plot_dict"][plot_name]["core_set"]
+                                else filter_out_path
+                            )
+                            try:
+                                # ns-rse 2025-12-03 Could perhaps move logic for plotting here rather incurring cost of
```
SylviaWhittle: Yeah. Low priority but yes.

topostats/processing.py
```diff
@@ -1052,23 +970,22 @@ def get_out_paths(
     core_out_path.mkdir(parents=True, exist_ok=True)
     filter_out_path = core_out_path / filename / "filters"
     grain_out_path = core_out_path / filename / "grains"
+    # ns-rse 2025-12-25 Do we want to keep tracing nested or promote disordered/ordered/nodes/curvature/splining up a
```
SylviaWhittle: For now, keep nested. Can promote later if it becomes an issue?

topostats/tracing/disordered_tracing.py
```diff
@@ -662,6 +715,7 @@ def disordered_trace_grain(  # pylint: disable=too-many-arguments
         of coordinates.
     """
     disorderedtrace = disorderedTrace(
+        # grain_crop=grain_crop,
```
SylviaWhittle:
```suggestion

```

SylviaWhittle (commented):
A very rapid run-through of the actual code.

I'll start another pass through now.

Also found an issue in processing `minicircle.spm`. I'll document that in an issue.

topostats/default_config.yaml
tobyallwood: Small thing for future, can ignore for now:

Would it be helpful to list the options in the comment of this line?

ns-rse:
Thanks for finding this problem @tobyallwood I wouldn't have thought to check this.


> `run_modules.py process()` line 222: `grain_stats_all[str(filename)] = grain_stats_df.dropna(axis=1, how="all")`
> ```
>         # Grain Statistics
>         grain_stats = {
>             grain_number: grain_crop.stats for grain_number, grain_crop in topostats_object.grain_crops.items()
>         }
>         grain_stats_df = pd.DataFrame.from_dict(
>             {
>                 (grain_number, class_type, subgrain_number): grain_stats[grain_number][class_type][subgrain_number]
>                 for grain_number, _ in grain_stats.items()
>                 for class_type, _ in grain_stats[grain_number].items()
>                 for subgrain_number, _ in grain_stats[grain_number][class_type].items()
>             },
>             orient="index",
>         )
>         if grain_stats_df.shape != (0, 0):
>             grain_stats_df["image"] = topostats_object.filename
>             grain_stats_df["basename"] = topostats_object.img_path
>             grain_stats_df.index.set_names(["grain_number", "class", "subgrain"], inplace=True)
>         else:
>             grain_stats_df = None
> ```
> 
> `grain_stats_df` can return `None` for two reasons:
> 
> * `topostats_object.grain_crops` itself is either `None` or has a length of 0
> * `grain_stats_df.shape` is `(0, 0)`


Indeed, this happens when `TopoStats.grain_crops == None` we set `grain_stats_df = None` and return it. I cant see where we set `grain_stats_df = None` if the `grain_stats_df.shape == (0, 0)` though, there is a clause there to only add the columns `image` and `basename` and set the index only when there is data though.

 
> I can see this issue also appearing on lines 224 and 225 of `run_modules.py`:
> 
> ```
> image_stats_all[str(filename)] = image_stats_df.dropna(axis=1, how="all")
> disordered_tracing_all[str(filename)] = disordered_tracing_df.dropna(axis=1, how="all")
> ```
> 
> As they do not support cases for `image_stats_df` and `disordered_tracing_df` being `None` even though this is also a possible result from `process_scan()`

Yep, good spot too, this will also cause problems if `image_stats_df` and `disrodered_trace_df` respectively are `None`.


 
> Possible fix:
> 
> Due to `grain_stats_df` (and the existence of grains) being required for the rest of processing it may be best to exit the program when it sets all `*_df` values to `None`. It already skips grainstats and tracing stages (and lets the user know in the terminal output) but then goes on to try and process image statistics with this non-existent data.

Because we might be processing multiple images only one might fail we try to make sure things continue to run so that statistics are gathered on images that can be processed.

Looking at the code here I see that for `branch_stats_df` and `molecule_stats_df` I had recognised/anticipated the possibility of returning these values as `None` and have the clauses to only drop missing and add to the dictionaries if they are not `None`.

```
                if branch_stats_df is not None:
                    branch_stats_all[str(filename)] = branch_stats_df.dropna(axis=1, how="all")
                if molecule_stats_df is not None:
                    molecule_stats_all[str(filename)] = molecule_stats_df.dropna(axis=1, how="all")
```

I think therefore in this case that we should add similar clauses for `grain_stats_all`, `image_stats_df` and `disordered_tracing_df` so that we build these dictionaries out of dataframes that have been successfully returned.

Will also need some clauses further down though so that if _all_ images failed to produce these dataframes that we don't try to `pd.concat()` them as the dictionary would be empty in each case.

> Let me know if any more information is required for this

Excellent level of detail, thanks for digging in and doing such detailed testing and the comprehensive feedback. :+1:  :slight_smile:

It has however raised a couple more problems that I hadn't thought of as our tests have two files `old_bruker.002` and `old_bruker.004` and because we strip the file-extension the logs show these as being the same file and potentially the resulting in conflicts/ambiguity in the resulting CSV files. I'll write that up as an issue.

topostats/tracing/ordered_tracing.py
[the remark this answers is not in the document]
ns-rse: I think this is a little strange and indicative of a problem further up-stream in the processing because, possibly nodestats, if there are no segments that cross, which is the case here because `minus.max()` is `0` indicating that there are no points next to each other (`minus` is the result of applying [`skimage.measure.label()`](https://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.label) to the skeleton after the crossings have been removed which should leave distinct regions.

This particular grain is I think this one and processes and is ordered ok when the grain thresholding is `std_dev`...

image

Although note that it doesn't have individual molecules visualised/traced either but didn't throw this error when processing and working through the refactoring.

image

It does have a weird branch though with a loop in it, and this may be the unique feature that is causing the strange behaviour.

**NB** - We can't get the same images above out of TopoStats because they are produced _after_ Nodestats tracing has completed which is where the error arises.

As noted to reproduce this result...

> The only changes I made to the config to get this error was switching the grains thresholding method from `std_dev` to `absolute`, and I'm only running this on minicircle.spm for now to avoid the errors already found.

PR #1301, opened 2026-02-17
feature(config|run_modules): make default_config flexible

ns-rse:
Whilst developing entry points for [AFMSlicer](https://github.com/AFM-SPM/AFMSlicer) I found that the current
implementation of the `config.reconcile_config_args()` function was too restrictive. Try and use it in another package
and it will load the `default_config.yaml` from TopoStats.

This pull-request changes the functionality and adds the parameter `default_config` to the
`config.reconcile_config.args()`. This is loaded via `get_data(package=topostats.__package__,
resource="default_config.yaml")` prior to calling `reconcile_config_args()` (and so AFMSlicer will use
`get_data(package=afmslicer.__package__, resource="default_config.yaml")`).

All pre-commit and unit/regression tests pass.

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

### DIFF
--- a/topostats/io.py
+++ b/topostats/io.py
@@ -7,6 +7,7 @@ import os
 import pickle as pkl
 import re
 import struct
+from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
 from typing import Any
@@ -916,7 +917,7 @@ class LoadScans:
             img_path=self.img_path.with_name(filename),
             image=None,
             image_original=image,
-            config=self.config,
+            config=deepcopy(self.config),
         )
 
 


ns-rse:
Thanks @SylviaWhittle, yes works with AFMSlicer and I can now load the correct configuration.

Working out how to build on and extend functionality is interesting, there are a few other places (such as creating configuration files) where I've already made things more flexible (I broke that in doing so before getting it right!). 

It's well worth it though.

ns-rse:
Weird the test fail under Python 3.10 on M$-Win again, I'll re-run once but if it fails will merge.

PR #1311, opened 2026-03-05
Improve typing consistency in codebase

### LINKED ISSUE #1299: Improve typing consistency in Topostats
Based on the suggestion by @ns-rse in #1280 

I'm going to run type checking over the codebase (starting with `processing.py`) and fix issues as I find them.

I'm also happy to set up type checking as a CI action for future work.

ubdbra001:
Fix #1299 

This PR addresses the inconsistent typing in the codebase.
Given the size of this I'm going to do it incrementally in this PR. The initial stage of this is picking out the "easy wins" e.g. cases where the type hints are missing or incorrect and are easy/obvious to update.

At this stage all the tests still pass.

---

- [x] Existing tests pass.
  - [ ] `docs/configuration.md`
  - [ ] `docs/usage.md`
- [ ] Pre-commit checks pass.


topostats/processing.py
```diff
@@ -1059,7 +1059,7 @@ def process_scan(
     """
     # Setup configuration, we use that from the topostats_object.config if not explicitly given an option
     config = topostats_object.config.copy()
-    base_dir = config["base_dir"] if base_dir is None else base_dir
+    base_dir = Path(config["base_dir"] if base_dir is None else base_dir)
```
ns-rse: Neat trick! I'd never have thought to enclose the whole conditional within `Path()` and would have put it around both of the values, but it makes total sense as only one of the two is returned and is then converted to `Path()`.

ns-rse (commented):
Good work tackling this behemoth of a task @ubdbra001 and apologies for it being so messy (typing only came onto my radar way after I'd started).

Some comments in-line from a quick scan, don't have time to investigate more deeply I'm afraid.

SylviaWhittle (commented):
Just a thing I noticed skim-reading this.

topostats/tracing/ordered_tracing.py
```diff
@@ -229,7 +229,7 @@ def remove_common_values(
 
         Returns
         -------
-        np.array
+        npt.ArrayLike
```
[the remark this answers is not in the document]
ubdbra001: `npt.NDArray` sounds more sensible, and I suspect I've used it elsewhere, so I'll stick to that one. Thanks!

topostats/processing.py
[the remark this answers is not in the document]
ubdbra001: This was the only one I've spotted so far, happy to open an issue to be addressed later

ns-rse (commented):
Some responses in-line...

topostats/measure/feret.py
```diff
@@ -368,8 +367,8 @@ def min_max_feret(
 
     Returns
     -------
-    dictionary
-        Tuple of the minimum feret distance and its coordinates and the maximum feret distance and  its coordinates.
+    dict
+        Needs updating: Tuple of the minimum feret distance and its coordinates and the maximum feret distance and  its coordinates.
```
SylviaWhittle: Dictionary containing the minimum and maximum feret diameters of the grain and their corresponding coordinates of contact with the calipers.

topostats/measure/feret.py
```diff
@@ -409,8 +408,8 @@ def get_feret_from_mask(mask_im: npt.NDArray, axis: int = 0) -> tuple[float, tup
 
     Returns
     -------
-    Tuple[float, Tuple[int, int], float, Tuple[int, int]]
-        Returns a tuple of the minimum feret and its coordinates and the maximum feret and its coordinates.
+    dict
+        Needs updating: Returns a tuple of the minimum feret and its coordinates and the maximum feret and its coordinates.
```
SylviaWhittle: Ditto as above.

topostats/scars.py
```diff
@@ -348,7 +348,7 @@ def remove_scars(
     threshold_high: float = 0.666,
     max_scar_width: int = 4,
     min_scar_length: int = 16,
-):
+) -> tuple[npt.NDArray, npt.NDArray | None]:
```
SylviaWhittle: This should never return None as the second value, perhaps add an assert guard clause before the return? On the first iteration it's set, and it should not be able to be called with iterations == 0, so maybe another guard clause for that? Unsure.

topostats/scars.py
```diff
@@ -385,9 +385,11 @@ class was instantiated.
 
     Returns
     -------
-    self.img
+    img: npt.NDArray
         The original 2-D image with scars removed, unless the config has run set to False, in which case it
         will not remove the scars.
+    first_marked_mask: npt.NDArray | None
+        Needs updating
```
SylviaWhittle: The scars detected during the first iteration. (I forget why I added this, will look into, it's years old hah. Also very slow.)

topostats/tracing/ordered_tracing.py
```diff
@@ -214,7 +214,7 @@ def compile_images(
     @staticmethod
     def remove_common_values(
         ordered_array: npt.NDArray, common_value_check_array: npt.NDArray, retain: list = ()
```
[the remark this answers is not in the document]
SylviaWhittle: This is a good solution and yeah, default args should not be data structures as you point out.

SylviaWhittle (changes requested):

SylviaWhittle:
Looked at all changes and have provided answers to queries for docstrings.

This is good to go once the docstrings are updated and provided tests pass.

Thank you very much @ubdbra001.

topostats/scars.py
ubdbra001: So when using this function the `removal_iterations` argument should be >=2?

PR #1340, opened 2026-04-15
Automate mkdocs deployment

ubdbra001:
This PR updates the existing workflow for auto-deploying the Topostats docs. It:
- Updates the `docs` dependencies to use `mkdocs` and the required plugins
- Removes the old `sphinx` workflow
- Adds a new workflow using `mkdocs` and `mike` to produce versioned docs

The new workflow does the following:
- On a merged PR or a manual dispatch the `dev` version of the docs will be updated. Older snapshots (e.g. 2.4.0) will be left as they are.
- When a new release is created (not a pre-release) the workflow will take a snapshot of the docs, add it to the list of versions available on the site using the version ID used in the release, and move the '(latest)` suffix to the most recent release of the docs.

I have tested this on my own branch and it does appear to be working correctly. I think that does everything discussed previously, but let me know if there is anything else required.


SylviaWhittle (approved):
🤞 that this all works - it will be great. :)

Issue #1339, opened 2026-04-14
[feature] : Revert basename to only referencing the directory, not the file?


SylviaWhittle:
### Is your feature request related to a problem?

Post-refactor, the `basename` column now displays the filename as well as the directory structure, making it harder to filter for experimental conditions when plotting.

```
basename
─────────────────────────────────────────────────
all_data/sc/20251031_SCpicoz_8ng_nicl.0_00058.spm
all_data/sc/20251031_SCpicoz_8ng_nicl.0_00056.spm
all_data/sc/20251031_SCpicoz_8ng_nicl.0_00055.spm
```

### Describe the solution you would like.

Could / should we revert this? Was there a reason for it initially?

### Describe the alternatives you have considered.

Perhaps it was done for a reason? I fail to recall.

SylviaWhittle:
Done in #1368

PR #1342, opened 2026-04-16
Add close end joining and overhaul curvature statistics (sorry for multiple changes in one PR, wasn't my decision)

SylviaWhittle:
# TopoStats Pull Requests

Adds:
- End joining.
- Curvature overhaul.
- Preparatory work for Tobi's and my project.

---

Before submitting a Pull Request please check the following.

- [x] Existing tests pass.
- [x] Documentation has been updated and builds. Remember to update as required...
  - [x] `docs/usage/configuration.md`
  - [x] `docs/usage/data_dictionary.md`
  - [x] `docs/usage/advanced.md` and new pages it should link to.
- [x] Pre-commit checks pass.
- [x] New functions/methods have typehints and docstrings.
- [NA] New functions/methods have tests which check the intended behaviour is correct. - Nope but not planned / allowed / no time.

### `topostats/default_config.yaml`

If adding options to `topostats/default_config.yaml` please ensure.

- [x] There is a comment adjacent to the option explaining what it is and the valid values.
- [x] A check is made in `topostats/validation.py` to ensure entries are valid.
- [ ] Add the option to the relevant sub-parser in `topostats/entry_point.py`.

PR #1349, opened 2026-04-28
Move .topostats output files + remove filename from basename

### LINKED ISSUE #1338: [feature] : Move topostats output files back up one directory to lie beside the corresponding images

It's annoying to navigate deeper into a specific directory, remembering the filename and grabbing it - even with terminal commands / CLI file managers (before anyone assumes it's just a GUI nav issue).

Move the `.topostats` files back to where they were.

Not sure if there is a reason necessitating their presence where they are?

### Additional context

_No response_

tobyallwood:
# TopoStats Pull Requests

Closes #1338 and #1339

There are two changes in this PR, they are both very small though so have been put together

1) `.topostats` files are now saved to the parent directory of their previous location (`processed/output/file.topostats` rather than `/processed/output/file.spm/file.topostats`)

2) the basename field in output `.csv`s no longer has the filename included; only the directory the file is stored in is shown (`tests/resources/` rather than `tests/resources/minicircles.spm` for example). This should make it easier for users to sort + filter by basename after batch processing without having to filter one file at a time.

### Note:

During testing all `.csv`s containing a `basename` field have `tests/resources` (as expected) except for `grain_statistics.csv` which just has a '`.`'. Is this correct? I can't see any specific difference in the changes I made to `grain_statistics.csv` over other output sheets but I'm confused as to why there is a difference

---

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

(documentation/ new tests not required)

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -58,11 +58,11 @@ def test_process_scan(tmp_path, process_scan_config: dict, load_scan_data: LoadS
     assert grain_stats.to_string(float_format="{:.4e}".format) == snapshot
 
     # Regtest for the topostats file
-    assert Path.exists(tmp_path / "tests/resources/test_image/processed/topostats/minicircle_small.topostats")
+    assert Path.exists(tmp_path / "tests/resources/test_image/processed/minicircle_small.topostats")
     # Load the results, note that we use AFMReader.topostats.load_topostats() here which simply loads the data as
     # dictionaries and means it is easy to compare to syrupy snapshots
     saved_topostats = load_topostats(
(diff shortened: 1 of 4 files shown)


tobyallwood:
Only ubuntu failed? A single test failed where curvature values were a little different to the comparison snapshot. Is this a known issue? Can't see how it's related to the changes in this PR at any rate.

tobyallwood:
#1342 fixes the ubuntu issue, this PR should be left until that is merged (hopefully soon)

SylviaWhittle:
Unblocked by #1342 now.

SylviaWhittle (approved):
This is great, thank you.

PR #1362, opened 2026-06-04
Updated link to config docs

### LINKED ISSUE #1350: __init__ references old docs
__init__.py still references the old docs (with the `/main/` slug) which leads to a 404. This is a quick fix and I'll get on it later so I'm putting this here to remember

tobyallwood:
Closes #1350 

When updating docs a link in `__init__.py` was missed that points to the configuration documentation, this has been fixed.

### DIFF
--- a/topostats/__init__.py
+++ b/topostats/__init__.py
@@ -27,7 +27,7 @@ else:
     TOPOSTATS_BASE_VERSION = str(TOPOSTATS_VERSION)
     TOPOSTATS_COMMIT = ""
 CONFIG_DOCUMENTATION_REFERENCE = """# For more information on configuration and how to use it:
-# https://afm-spm.github.io/TopoStats/main/configuration.html\n"""
+# https://afm-spm.github.io/TopoStats/dev/usage/configuration/\n"""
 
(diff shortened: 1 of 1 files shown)


topostats/__init__.py
```diff
@@ -27,7 +27,7 @@
     TOPOSTATS_BASE_VERSION = str(TOPOSTATS_VERSION)
     TOPOSTATS_COMMIT = ""
 CONFIG_DOCUMENTATION_REFERENCE = """# For more information on configuration and how to use it:
-# https://afm-spm.github.io/TopoStats/main/configuration.html\n"""
+# https://afm-spm.github.io/TopoStats/dev/usage/configuration/\n"""
```
ubdbra001: Just double checking: Should this be `dev` or `latest`?
`latest` will be the docs for the latest release
`dev` will be for the most recent commit

ubdbra001 (commented):
One query

topostats/__init__.py
tobyallwood: Good point, yeah latest would be best

ubdbra001 (approved):
Good to merge!

PR #1363, opened 2026-06-04
Added optional pad_inches parameter to Images

tobyallwood:
When calling methods from the `Images` class in `plottingfuncs.py` for PerovStats I sometimes require custom `pad_inches` values to fit all titles, axes labels and hightscales into the image. The default value cuts some text from the image.

`pad_inches` is now an optional parameter of `Images`, if no value is given the matplotlib default value is used.

---

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

### DIFF
--- a/topostats/plottingfuncs.py
+++ b/topostats/plottingfuncs.py
@@ -155,6 +155,8 @@ class Images:
         Optionally number each grain in a plot.
     trace_linewidth : float
         Width of lines when plotting splines/curvature.
+    pad_inches : float
+        Inches to pad a figure by to allow for heightscales + names.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


tobyallwood:
Unsure why these tests are failing? All pass for me locally. @ubdbra001 @SylviaWhittle any ideas?

ubdbra001:
It looks like in the original run one of the snapshot test failed, but it look like a flaky test if re-running the job is enough to get it to pass. Depending on how flaky it is it may be a pain to track down why...

tobyallwood:
Yeah that was it, re-run the tests and they've all passed.

On a side note, this PR is mostly so I can create the figures I want to in my PerovStats project meaning to actually get use out of it this will possibly need adding to an actual release? How is best to go about this?

I figured updating TopoStats to suit it now would be best as they'll be getting merged eventually anyway.

Edit: For now I'll copy paste the `Images` class from TopoStats to PerovStats and when we get to the point of putting out a new topostats release I'll return to directly referencing TopoStats

tobyallwood:
Closed due to other options being tested

PR #1368, opened 2026-06-11
Correctly save basename to grain_stats.csv

tobyallwood:
Recently how `basename` is used was changed, and an error has been discovered that when creating the dataframe for `grain_statistics.csv` `basename` is always `.`.

This was because `.name` was being incorrectly used, `.name` strips all directory info from the variable and only keeps the filename, which then can't give a parent so the output is just `.`.

I think during the recent changes to basename functionality I must've missed changing the `grain_statistics.csv` basename to the final version as all other `.csv`s have working basename fields. Apologies!


ubdbra001 (approved):
Seems fine to me

PR #1374, opened 2026-06-24
Add time taken to the completion message

### LINKED ISSUE #1294: [feature] : Add time taken to process to output.

_No response_

It would be good to know how long processing took, as often when running large datasets, we leave our computers running and come back to it later.

If we knew how long the processing took, then we could more accurately plan our tasks to come back around the time it took last time we processed the dataset.

Plus it's just generally nice to know.

Eg:

```
  TopoStats Version           : 2.3.2
  TopoStats Commit            : 1e6f50fb7.d20260204
  Base Directory              : ../all_data
  File Extension              : .spm
  Files Found                 : 386
  Successfully Processed^1    : 383 (99.22279792746114%)
  Time taken                  : 01:05:22 (HH-MM-SS).
  All statistics              : output/grain_statistics.csv
  Distribution Plots          : Disabled. Enable in config 'summary_stats/run' if needed.
```

_No response_

### Additional context

_No response_

ubdbra001:
An attempt to resolve #1294 

This records the `datetime` for each function in `run_modules` that later calls `completion_message`, and then passes it when `completion_message` is called to calculate the duration elapsed.

It doesn't break any tests but it would be good if someone else could test it on a real run.

### DIFF
--- a/topostats/processing.py
+++ b/topostats/processing.py
@@ -1644,7 +1644,9 @@ def check_run_steps(  # noqa: C901
         LOGGER.info("Configuration run options are consistent, processing can proceed.")
 
 
-def completion_message(config: dict, img_files: list, summary_config: dict, images_processed: int, start_time: datetime) -> None:
+def completion_message(
+    config: dict, img_files: list, summary_config: dict, images_processed: int, start_time: datetime
+) -> None:
     """
     Print a completion message summarising images processed.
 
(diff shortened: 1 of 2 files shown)


SylviaWhittle:
@SylviaWhittle to check if this works in actual usage.

SylviaWhittle (approved):
Tested locally. Works. :) 

Thank you.

PR #1376, opened 2026-07-08
Added disordered trace grain data to grain_statistics.csv

tobyallwood:
# TopoStats Pull Requests

Closes #1371
Closes #1372 

After the refactor some disordered trace data was collected but never saved to `.csv`, this PR rectifies that.
Documentation reflecting this has also been updated.

Further changes to the documentation have also been made; these changes are primarily focussed on ensuring the correct `.csv` names are referenced.

---

Before submitting a Pull Request please check the following.

- [x] Existing tests pass.
- [x] Pre-commit checks pass.

PR #1380, opened 2026-07-10
Fix failing and slow testing suite

### LINKED ISSUE #1377: [bug]: TopoStats testing suite failing without apparent cause
### Checklist

- [x] Run using the `main` branch to check if the issue persists. Instructions on how to do this can be found [here](https://afm-spm.github.io/TopoStats/main/installation.html#installing-from-github).
- [x] Find the offending file in the output. If processing halts, re-run analysis with `topostats --core 1 process`.
- [x] Describe the bug.
- [x] Include the configuration file.
- [x] Copy of the log-file from running with `topostats --log-level debug <command>`.
- [x] The exact command that failed. This is what you typed at the command line, including any options.
- [x] TopoStats version, this is reported by `topostats --version`
- [x] Operating System and Python Version

### Describe the bug

No time to properly document. See #1376 and #1374 to see that the tests are failing at the same point on image comparisons in two PRs that do not mess with the underlying data.

Someone look into this please. Thanks.

Image

### Copy of the log-file from running with `topostats --log-level debug <command>`

NA

NA

### To Reproduce

NA

### TopoStats Version

Git main branch

### Python Version

3.11

### Operating System

MacOS M1/M2 (post-2021)

### Python Packages

NA

ubdbra001:
Pin dependencies to fix failing and slow test suite.

- `matplotlib` 3.11 produces different images to the baseline. Updating the baseline breaks tests for Python 3.10, so pinning this for now.
- `syrupy` >= 5.5 slows down running the tests so reverting to a previous version that deosn't cause this issue

Both of these are plaster fixes, longer term solutions are required:
- Baseline images should be updated, probably a job for when python 3.10 is dropped
- The underlying issue for `syrupy` is likely the large snapshot for the nodestats tests, this should be investigated.


ns-rse:
I've only had a cursory look at what tests are failing (spent the morning trying to recover what looks like a failed harddrive so on a temporary computer that isn't ideally configured).

You could try adding a `marks=pytest.mark.skipif()` clause to skip tests where images are changing so they don't run if `Python != 3.10` for the time being, then update the images with newere Python/matplotlib/numpy. Or update the images and skip if on 3.10. You should be able to get this from environment variables in some manner.

I'd probably want to check what is changing in the images though but I'm afraid I don't have time to investigate that myself right now (perhaps add them to #1377 if they are readily available they will be under somewhere like `/tmp/pytest/`. You can also run specific tests and use the flag `--mpl-generate-summary=html` which is a useful way of getting an insight to what is changing.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
   "AFMReader",
   "h5py",
   "keras",
-  "matplotlib",
+  "matplotlib<3.11",
```
ns-rse: Perhaps leave a note referencing why?

```suggestion
  "matplotlib<3.11", # See #1377
```

```shell
git blame -- pyproject.toml
```

...is useful in the short term but only shows the last commit and it can be cumbersome to dig deeper into the history. Linking to the issue makes it easier for people to understand why the version has been pinned (the issue in turn will of course link to this PR).

pyproject.toml
```diff
@@ -59,7 +59,7 @@ tests = [
   "pytest-profiling",
   "pytest-xdist",
   "filetype",
-  "syrupy",
+  "syrupy<5.5",
```
ns-rse:
```suggestion
  "syrupy<5.5", # See #1377
```

ns-rse (commented):
Perhaps make it easier for people to find why the versions have been pinned.

pyproject.toml
ubdbra001: That's a good shout, will do.

ubdbra001:
To do:
1. Skip testing for 3.10, and update figure snapshots
2. Pin `syrupy` and add issue outlining why

ubdbra001:
Test are failing due to a bug introduced in #1376, looks like it is caught and fixed in #1382

Edit: [Tests passing](https://github.com/ubdbra001/TopoStats/pull/4) on my fork where I haven't added #1376 yet

tests/resources/img/test_plot_and_save_no_axes_no_colorbar.png
SylviaWhittle, line 1: Interesting that this changed too since it is only supposed to be a font change from what we were talking about. But I am not concerned.

SylviaWhittle (approved):
All good!

tobyallwood (approved):
Everything looks good to me I think

PR #1381, opened 2026-07-14
Removal of above/ below in grain thresholds

### LINKED ISSUE #1332: [feature] : Tweak: Default config: Remove the below thresholds (since they now always run)

In the new refactor version of TopoStats, with the removal of the `grains.direction` parameter, both the `above` and `below` options for each threshold are applied (correctly). But the default still runs a below threshold since it's set.

Image

Remove the below values from the default config since it's so rare that it will be used.

Additionally, I had envisioned that the whole `above` and `below` keys be removed in #1235 but it never got merged due to conflicts with the refactor.

I only bring this up because I have lost some time to tracing a bug that was caused by inadvertently running the below as well as above accidentally.

_No response_

### Additional context

_No response_

### LINKED ISSUE #1333: [feature] : Tweak: Remove above / below

As per #1235 and sibling issue with #1332.

Remove the above / below from the config and use positive / negative signage in the lists of thresholds instead. It'll simplify things IIRC from what we were planning a while back. I could be wrong, head not in that space currently, not a hard push for this but a firm one.

_No response_

### Additional context

_No response_

### LINKED ISSUE #1334: [bug]: Removal of below threshold doesn't prevent thresholding below
### Checklist

- [x] Run using the `main` branch to check if the issue persists. Instructions on how to do this can be found [here](https://afm-spm.github.io/TopoStats/main/installation.html#installing-from-github).
- [x] Find the offending file in the output. If processing halts, re-run analysis with `topostats --core 1 process`.
- [x] Describe the bug.
- [x] Include the configuration file.
- [x] Copy of the log-file from running with `topostats --log-level debug <command>`.
- [x] The exact command that failed. This is what you typed at the command line, including any options.
- [x] TopoStats version, this is reported by `topostats --version`
- [x] Operating System and Python Version

### Describe the bug

I don't have time to write this up right now properly, pressure to get things done fast.

TDLR:

I removed the below threshold from the list in the config file as such:

Image

But TS still ran the thresholding with it, resulting in a full mask.

### Copy of the log-file from running with `topostats --log-level debug <command>`

[Fri, 10 Apr 2026 13:10:01] [INFO    ] [topostats] [minicircle.spm] : *** Grain Finding ***
[Fri, 10 Apr 2026 13:10:01] [INFO    ] [topostats] [minicircle.spm] : Finding below grains, threshold: ([])
traditional_full_mask_tensor.shape: (1024, 1024, 1) unique: [ True] sum: 1048576
class sums: [1048576]

Default config with the `grains.threshold_absolute.below` set to `[]`.

This would be easily fixed by #1332 I think.

### To Reproduce

Run topostats with described config edit.

### TopoStats Version

Git main branch

### Python Version

3.11

### Operating System

MacOS M1/M2 (post-2021)

### Python Packages

```
absl-py==2.4.0
AFMReader==0.0.1
annotated-types==0.7.0
art==6.5
asttokens==2.4.1
astunparse==1.6.3
biopython==1.86
certifi==2026.2.25
charset-normalizer==3.4.5
cheap_repr==0.5.2
contourpy==1.2.1
coverage==7.6.1
cycler==0.12.1
decorator==5.1.1
et_xmlfile==2.0.0
executing==2.0.1
filetype==1.2.0
flatbuffers==25.12.19
fonttools==4.53.1
gast==0.7.0
google-pasta==0.2.0
grpcio==1.78.0
h5py==3.11.0
idna==3.11
igor2==0.5.8
imageio==2.35.1
iniconfig==2.0.0
ipython==8.26.0
jedi==0.19.1
Jinja2==3.1.4
joblib==1.4.2
keras==3.13.2
kiwisolver==1.4.5
lazy_loader==0.4
libclang==18.1.1
llvmlite==0.46.0
loguru==0.7.2
markdown-it-py==4.0.0
MarkupSafe==2.1.5
matplotlib==3.9.2
matplotlib-inline==0.1.7
mdurl==0.1.2
ml_dtypes==0.5.4
namex==0.1.0
networkx==3.3
numba==0.64.0
numpy==2.0.2
numpyencoder==0.3.0
opencv-python==4.10.0.84
openpyxl==3.1.5
opt_einsum==3.4.0
optree==0.19.0
packaging==24.1
pandas==2.2.2
parso==0.8.4
pexpect==4.9.0
pillow==10.4.0
pluggy==1.5.0
prompt_toolkit==3.0.47
protobuf==7.34.0
psutil==5.9.8
ptyprocess==0.7.0
pure_eval==0.2.3
pydantic==2.12.5
pydantic_core==2.41.5
pyfiglet==1.0.2
Pygments==2.18.0
pyparsing==3.1.2
pyspm==0.6.3
pytest==7.4.4
pytest-cov==5.0.0
pytest-github-actions-annotate-failures==0.2.0
pytest-lazy-fixture==0.6.3
pytest-mpl==0.17.0
pytest-regtest==2.1.1
python-dateutil==2.9.0.post0
pytz==2024.1
PyYAML==6.0.2
requests==2.32.5
rich==14.3.3
ruamel.yaml==0.18.6
ruamel.yaml.clib==0.2.8
schema==0.7.7
scikit-image==0.24.0
scikit-learn==1.5.1
scipy==1.14.0
seaborn==0.13.2
setuptools==82.0.1
six==1.16.0
skan==0.13.1
snoop==0.4.3
stack-data==0.6.3
tensorflow==2.21.0
termcolor==3.3.0
threadpoolctl==3.5.0
tifffile==2024.8.10
toolz==1.1.0
topoly==1.1.0
-e git+ssh://git@github.com/AFM-SPM/TopoStats.git@80543d675064be9c6e86071a4e197bfc5d3c3040#egg=topostats
tqdm==4.66.5
traitlets==5.14.3
ty==0.0.22
typing-inspection==0.4.2
typing_extensions==4.15.0
tzdata==2024.1
urllib3==2.6.3
wcwidth==0.2.13
wheel==0.46.3
wrapt==2.1.2
```

tobyallwood:
# Remove above/ below from grain thresholds

Closes #1332 
Closes #1333 
Closes #1334

This PR simplifies the grain finding thresholds by removing a layer of indentation (`above`/ `below` keys). The direction of the threshold is now selected by the sign of the threshold value in the config. For example:

__old__:
```
threshold_std_dev:
    below: 10.0
    above: 1.0
```

__new__:
```
threshold_std_dev: [-10.0, 1.0]
```

---

This also means that `area_thresholds` has changed to lose the directional indentation layer:

__old__:
```
area_thresholds:
  above: [null, null]
  below: [300, 3000]
```

__new__:
```
area_thresholds: [[null, null], [300, 3000]]
```

The new `area_thresholds` argument is a 2d list, each sub-list is for the threshold it shares an index number with.

---

Any reference/ use of direction in loops through TopoStats have either been removed or changed into a single loop of all thresholds in the list regardless of their direction. They are referred to via their index in the thresholds list.

---

- [x] Existing tests pass.
  - [x] `docs/usage/configuration.md`
  - [x] `docs/usage/data_dictionary.md`
  - [x] `docs/usage/advanced.md` and new pages it should link to.
- [ ] Pre-commit checks pass.


topostats/utils.py
```diff
@@ -142,19 +191,76 @@ def get_mask(image: npt.NDArray, thresholds: dict, img_name: str = None) -> npt.
...
+    image : npt.NDArray
+        2D Numpy array of image to be masked.
+    threshold_method : str
+        Method for thresholding, 'otsu', 'std_dev' or 'absolute' are valid options.
+    otsu_threshold_multiplier : float
+        Scaling value for Otsu threshold.
+    threshold_std_dev : list[float]
+        List of thresholds for the standard deviation method.
+    absolute : list[float]
+        List of absolute thresholds.
```
ubdbra001: These are optional arguments. include this in the docstring

topostats/utils.py
```diff
@@ -118,10 +143,34 @@ def _get_mask(image: npt.NDArray, thresh: float, threshold_direction: str, img_n
         return image > thresh
     LOGGER.debug(f"[{img_name}] : Masking (below) Threshold: {thresh}")
     return image < thresh
-    # LOGGER.fatal(f"[{img_name}] : Threshold direction invalid: {threshold_direction}")
 
 
-def get_mask(image: npt.NDArray, thresholds: dict, img_name: str = None) -> npt.NDArray:
+def get_grain_mask(image: npt.NDArray, thresholds: list, img_name: str = None) -> npt.NDArray:
```
ubdbra001: If the argument could be `None` then make sure that is reflected in the type hint

```suggestion
def get_grain_mask(image: npt.NDArray, thresholds: list, img_name: str | None = None) -> npt.NDArray:
```

topostats/utils.py
```diff
@@ -118,10 +143,34 @@ def _get_mask(image: npt.NDArray, thresh: float, threshold_direction: str, img_n
...
+
+    Parameters
+    ----------
+    image : npt.NDArray
+        2D Numpy array of the image to have a mask derived for.
+    thresholds : dict
+        Dictionary of thresholds, at a bare minimum must have key 'below' with an associated value, second key is
+        to have an 'above' threshold.
+    img_name : str
+        Image name that is being masked.
```
ubdbra001: optional argument

topostats/utils.py
```diff
@@ -93,7 +93,32 @@ def convert_path(path: str | Path) -> Path:
     return Path().cwd() if path == "./" else Path(path).expanduser()
 
 
-def _get_mask(image: npt.NDArray, thresh: float, threshold_direction: str, img_name: str = None) -> npt.NDArray:
+def _get_grain_mask(image: npt.NDArray, thresh: float, img_name: str = None) -> npt.NDArray:
```
ubdbra001: If the argument could be `None` then make sure that is reflected in the type hint.

```suggestion
def _get_grain_mask(image: npt.NDArray, thresh: float, img_name: str | None = None) -> npt.NDArray:
```

topostats/utils.py
```diff
@@ -93,7 +93,32 @@ def convert_path(path: str | Path) -> Path:
...
+    Calculate a mask for pixels that exceed the threshold.
+
+    Parameters
+    ----------
+    image : np.array
+        Numpy array representing image.
+    thresh : float
+        A float representing the threshold.
+    img_name : str
+        Name of image being processed.
```
ubdbra001: optional argument

topostats/simple_config.yaml
```diff
@@ -24,14 +24,9 @@ grains:
...
   threshold_method: std_dev # Options : std_dev, otsu, absolute, unet
-  threshold_std_dev:
-    below: 10.0 # Threshold for grains below the image background
-    above: 1.0 # Threshold for grains above the image background
-  threshold_absolute:
-    below: -1.0 # Threshold for grains below the image background
-    above: 1.0 # Threshold for grains above the image background
-  area_thresholds:
-    above: [300, 3000] # above surface [Low, High] in nm^2 (also takes null)
+  threshold_std_dev: [-10, 1] # Thresholds for grains relative to the image background. Negative values are below and positive values are above. List[float].
```
ubdbra001, line 27: Just for my understanding:
- Is there a specific order for these (i.e. would `[-10, 1]` produce the same result as `[1, -10]`)?
- Are only 2 values allowed?
- Should one be negative and one positive?

topostats/grains.py
```diff
@@ -79,12 +79,12 @@ class Grains:
...
+    threshold_std_dev : list[float] | None
+        List of factors by which standard deviation is multiplied to derive the threshold.
         if threshold_method is 'std_dev'.
-    threshold_absolute : dict[str, float | list] | None
-        Dictionary of absolute 'below' and 'above' thresholds for grain finding.
-    area_thresholds : dict[str, list[float | None]]
+    threshold_absolute : list[float] | None
+        List of absolute thresholds for grain finding.
+    area_thresholds : list[list[float]]
         Dictionary of above and below grain's area thresholds.
```
ubdbra001: Description doesn't match the line above (list of lists not a dict)

ubdbra001 (commented):
I had a quick skim and made some comments on things that jump out at me.

The changes to the docs are minimal, and from looking at them I'm not clear about what values I can include for these config options (Does order matter, how many values can I include, etc.) I'm probably not remembering the context for these options, but I think it would be good to spell them out somewhere.

topostats/simple_config.yaml
tobyallwood: - There's no specific order but it's important to be aware of the order you chose as in the output `.csv`s the thresholds are referenced by index.

- No, it can be any amount (except none)

- No, you can have any configuration

I should definitely make this clearer, I will in the docs but would more information be good in the config file too?

ubdbra001: If you want to you can, but I don't think it's vital. It's worth bearing in mind that if you have the same info in multiple places and they both have to be manually updated for future changes there is a reasonable chance that one of those places will get overlooked and you will end up with differing information.

I saw you've updated the schema for validation. It would be good to update this to ensure an empty list can't be passed (if I'm interpreting your answer correctly)

tobyallwood:
Failing tests all seem to be plot image checks, the output image and the expected image have data such as axes ticks, plot titles, legends etc. translated a few pixels over

image


Edit: this now seems to have fixed itself??

tobyallwood:
One thing I'm noting here for documentation/ future reference purposes:

In `grain_statistics.csv` do we want to reference the threshold by its index in the threshold list or by the actual value of the threshold? Or is this field required at all?

Assuming it is required...

If the thresholds are `[-1, 1]` and a grain is found using -1 (aka below 1), would we want to use

```threshold: 0```

or

```threshold: -1```

---

My immediate thought is that the second option seems better but might make `std_dev` thresholds look confusing to a user as the number saved would not be the number they entered into the config.

On the other hand clear and obvious documentation will be needed if we reference the threshold by index so a `threshold: 1` is understood as index 1 rather than the actual threshold value being 1.

Previously the closest thing we had to this value was called `direction` and held either `"above"` or `"below"` but I think it would be more useful to point to a specific threshold seeing as multiple thresholds can be in the same direction

As far as I can see either option will require fairly minimal additions to implement, the threshold index is already stored and used in log messages and the actual threshold value can easily be found with that index value.

**Comment by SylviaWhittle** [comment]
> One thing I'm noting here for documentation/ future reference purposes:
> 
> In `grain_statistics.csv` do we want to reference the threshold by its index in the threshold list or by the actual value of the threshold? Or is this field required at all?
> 
> Assuming it is required...
> 
> If the thresholds are `[-1, 1]` and a grain is found using -1 (aka below 1), would we want to use
> 
> `threshold: 0`
> 
> or
> 
> `threshold: -1`
> 
> My immediate thought is that the second option seems better but might make `std_dev` thresholds look confusing to a user as the number saved would not be the number they entered into the config.
> 
> On the other hand clear and obvious documentation will be needed if we reference the threshold by index so a `threshold: 1` is understood as index 1 rather than the actual threshold value being 1.
> 
> Previously the closest thing we had to this value was called `direction` and held either `"above"` or `"below"` but I think it would be more useful to point to a specific threshold seeing as multiple thresholds can be in the same direction
> 
> As far as I can see either option will require fairly minimal additions to implement, the threshold index is already stored and used in log messages and the actual threshold value can easily be found with that index value.

(gently held opinion) I think we should use index for now.

Perhaps in the future we make the user name the thresholds, ie:

```
threshold_std_dev
 - "main" : 1.0
 - "protein" : 2.2
 - "holes" : -0.5
```

File: topostats/entry_point.py [hunk 8]
```diff
@@ -16,6 +16,31 @@
 # pylint: disable=too-many-statements
 
 
+def pair_float(list_str: str):
```
SylviaWhittle, line 19: ♡

SylviaWhittle (approved):
This is great, thank you for redoing this work. I appreciate it.

I also like the sweeping up you've done, keeping our kitchen tidy. :)

SylviaWhittle:
That's very odd - its failing on the test for unet preprocessing for making bounding boxes square. This code should be very isolated and resilient...

tobyallwood:
```
XFAIL tests/test_unet_masking.py::test_make_bounding_box_square[rectangular image with large rectangular box] - square bounding box not possible
FAILED tests/test_processing.py::test_process_scan - AssertionError: assert [+ received] == [- snapshot]
```

@SylviaWhittle Is it not the snapshot comparison in test_process_scan and test_unet_masking is just an xfail?

SylviaWhittle:
> ```
> XFAIL tests/test_unet_masking.py::test_make_bounding_box_square[rectangular image with large rectangular box] - square bounding box not possible
> FAILED tests/test_processing.py::test_process_scan - AssertionError: assert [+ received] == [- snapshot]
> ```
> 
> @SylviaWhittle Is it not the snapshot comparison in test_process_scan and test_unet_masking is just an xfail?

Ah yes. I skimmed too rapidly hah.

PR #1382, opened 2026-07-15
Include node count in `grain_statistics.csv`

### LINKED ISSUE #1379: [feature] : Re-add correct node count

_No response_

Would like an output of the node count per grain. @SylviaWhittle removed a broken previous version of this but it would be useful to re-implement so that it isn't just a count of the total node pixels, and is actually a count of the nodes once the pixels are combined. 

_No response_

### Additional context

_No response_

tobyallwood:
# Add node count to grain stats

Closes #1379 

After nodestats has run the total number of nodes can be collected and added to the grain's stats dictionary (called `num_nodes`)

The most appropriate `.csv` to save to is `grain_statistics.csv`, but there was an issue with how we'd do this as nodestats are collected on a graincrop level but `GrainCrop.stats` (therefore also `grain_statistics.csv`) is further divided into `class` and `subgrain`. For now I have just duplicated the `num_nodes` value across all rows related to a grain - not a perfect solution but better than having empty cells that could cause confusion.

A small bug has also been fixed related to the missing disordered tracing stats that were recently re-added where not running disordered tracing would cause a `KeyError`. This has been rectified with a quick check before attempting to access.

---

- [x] Existing tests pass.
  - [x] `docs/usage/data_dictionary.md`
- [x] Pre-commit checks pass.


ubdbra001 (commented):
Not super happy with the nested loops, but it looks like the usual way of doing things, so I can give it a pass.

I'd recommend adding test cases for `highlight_node_centres` at least. There is a stub already, but now would be a good time to complete it.

SylviaWhittle:
Calculated by count of labels, this doesn’t do what was done before where we counted simply the number of pixels in the node mask, which returned an erroneously high number of nodes for a given image.

tobyallwood:
I've added a test for `highlight_node_centres`, and from this some commented out fixtures in `tests/tracing/conftest.py` have been brought back.

I've also added `__init__.py`s for both `/tests/` and `/tests/tracing/` as not having these was causing pylint to throw an error and fail.

ubdbra001 (changes requested):
It would be good if you could include tests for no nodes and multiple nodes if possible, just so the behaviour of the function is explicit and tested for these cases too.
Though that's more of a should than a must, so if it's not something that could be added in (max) an afternoon then just add a note to the test that these cases need to be added later.

Once either of the above is done then this looks good to go.

ubdbra001 (approved):
Great, ready to merge!

Thanks!

PR #1388, opened 2026-08-04
Update docs deploy workflow

ubdbra001:
This fixes the issue with deploying a new release of the docs.
It also adds the option to manually deploy a new 'latest' version of the docs based on the most recent release tag (and will error if that version already exists or a tag can't be found).

Plus there are a few more messages to help with debugging, which can't hurt.

### DIFF
--- a/.github/workflows/deploy-mkdocs.yaml
+++ b/.github/workflows/deploy-mkdocs.yaml
@@ -10,6 +10,11 @@ on:
     types:
       - published
   workflow_dispatch:
+    inputs:
+      is_release:
+        description: 'Deploy as a Release version (using most recent release)?'
+        type: boolean
+        default: false
 
 permissions:
   contents: write
@@ -26,6 +31,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           fetch-depth: 0
+          fetch-tags: true
           ref: ${{ github.event_name == 'pull_request' && 'main' || github.ref }}
 
       - name: Set up Python from pyproject.toml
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-mkdocs.yaml
```diff
@@ -44,23 +50,46 @@ jobs:
...
+
+          # Check if docs for this version already exist
+          if mike list | grep -q "\"$TAG\""; then
+            echo "Docs for $TAG already exist. Skipping deploy."
+            exit 0
+          fi
+
           # Clean up the title of the previous 'latest' version
-          PREV_LATEST=$(mike list | grep "(latest)" | awk '{print $1}')
+          PREV_LATEST=$(mike list | grep "(latest)" | awk '{print $1}' | sed 's/"//g')
```
tobyallwood, line 82: This is more just for my understanding rather than a concern about the code, why is it important to remove quotes from the version? Aka. when do the quotes appear?

ubdbra001: Running `mike list` will produce an output where the latest version will be formatted like so: 
`"v2.4.0 (latest)" (v2.4.0) [latest]`

Then `awk '{print 1}'` will grab the first group (using spaces as a separator) which produces:
`"v2.4.0`

and then the code then code crashes because it can't match that version, due to the leading `"`

So `sed 's/"//g'` fixes this by finding any `"` and replacing it with nothing, which returns: `v2.4.0`

tobyallwood: Aah makes sense, thank you

PR #1395, opened 2026-08-18
Add patch to allow using laura's unet

SylviaWhittle:
Patch to allow an old unet model to run in our code.

TLDR:
- The model was trained weird, and so expects (B, RGB(R), RGB(G), RGB(B), W, H) RGB data as opposed to what we are invariably working with (B, C(arbitrary channels, not RGB), W, H) since we work with height map data.

This simply tries running the model and if it crashes because of input shape mismatch it tries again after repeating the input data channel across the RGB channels.

It's a frantic patch just to get the model working on main for a paper.

Probably wants removing at some point in the future, it's messy.

---

- [ ] Existing tests pass.
  - [ ] `docs/usage/configuration.md`
  - [ ] `docs/usage/data_dictionary.md`
  - [ ] `docs/usage/advanced.md` and new pages it should link to.
- [ ] Pre-commit checks pass.

### DIFF
--- a/topostats/unet_masking.py
+++ b/topostats/unet_masking.py
@@ -166,8 +166,16 @@ def predict_unet(
 
     # Predict the mask
     LOGGER.info("Running Unet & predicting mask")
-    prediction: npt.NDArray[np.float32] = model.predict(np.expand_dims(image_resized_np, axis=(0, 3)))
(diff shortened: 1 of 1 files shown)


SylviaWhittle:
Pushing this through given time pressure.

